PR #358, opened 2015-03-24
simplified ssl certificate creation and wiseService added to easybutton

WilliamSalusky:


awick:
Should we just have the plugins/Makefile install wiseService?

WilliamSalusky:
I don't believe wiseService should be installed to /plugins. As a standalone service, it belongs at the top level $TDIR. If you want to add it to a Makefile, I'm for whatever works but considering we can guide its deployment now for future consistency. I drop mine to /data/moloch/wiseService.  This is how I've already trained >60 people.

awick:
* moved wiseService install to "make install"
* took most of your cert creation stuff, awesome!

Issue #257, opened 2014-06-11
add by1, by2,  db1, db2, pa1,pa2 to sessions table


hillar:
Please add in sessions view into sessions table some more columns, or make it configurable :).

      "friendlyName": "Src Bytes",
      "help": "Total number of raw bytes sent by source in a session",
      "dbField": "by1"
      "friendlyName": "Dst Bytes",
      "help": "Total number of raw bytes sent by destination in a session",
      "dbField": "by2"

      "friendlyName": "Src data bytes",
      "help": "Total number of data bytes sent by source in a session",
      "dbField": "db1"
      "friendlyName": "Dst data bytes",
      "help": "Total number of data bytes sent by destination in a session",
      "dbField": "db2"

      "friendlyName": "Src Packets",
      "help": "Total number of packets sent by source in a session",
      "dbField": "pa1"
      "friendlyName": "Dst Packets",
      "help": "Total number of packets sent by destination in a session",
      "dbField": "pa2"

WilliamSalusky:
Just updated and I love it. However, I've just noticed a peculiarity where if you modify the displayed field set in any way from the default set, the dest port loses right click context. To reproduce, remove any field from the default display set in settings (i.e. remove src port), go to sessions and attempt to right click dst port. NOPE!

Additional enhancement: Any field added to the sessions display should also become right clickable. For example: If destination databytes is added the sessions display, it should be right click context aware to update the query bar with 'databytes.dst==${value}'

@jsarjas @awick

PR #381, opened 2015-05-10
New sessions table columns and configuration

jsarjas:
One column rename and three new columns.
Replacing listbox with checkboxes.
(Related to issue #257)

### DIFF
--- a/viewer/views/index.jade
+++ b/viewer/views/index.jade
@@ -33,11 +33,14 @@ block content
         th Packets
         th Src Packets
         th Dst Packets
-        th Bytes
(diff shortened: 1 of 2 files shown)


awick:
Thanks for all the work

* Can you squash commits for pull requests https://www.google.com/#q=github+squash+pull+request

* I'm not sure I want to make the default strings longer, even though it would describe what they are better.  I've tried really hard to make sure moloch will function on terrible 1024 or 1280 projectors, or  when using increased font size, which is what has led to the current names.  Maybe tool tips (just have title=fullname) would be a good solution.

PR #386, opened 2015-05-13
Fixed TCP flags offset

dis0ps:
Hello, 

I've been a long time Moloch user, and have recently experimented with integrating the netflow plugin with Logstash.  

I noticed the TCP flags seemed strange, and in doing some digging found that the offset within the TCP header was incorrect.  It's a minor fix, but one that I'm using to get accustomed to using GitHub.

I did a fork/pull-request on this, but if there are better ways to do minor fixes please let me know.  Thanks!

Dave

### DIFF
--- a/capture/nids.c
+++ b/capture/nids.c
@@ -382,7 +382,7 @@ void moloch_nids_cb_ip(struct ip *packet, int len)
                  session->addr2 == packet->ip_dst.s_addr &&
                  session->port1 == ntohs(tcphdr->th_sport) &&
                  session->port2 == ntohs(tcphdr->th_dport))?0:1;
-        session->tcp_flags |= *((char*)packet + 4 * packet->ip_hl+12);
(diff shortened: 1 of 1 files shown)


awick:
I'm not sure +13 is right, what strangeness are you seeing?  

Really the code should be changed to use th_flags, something like
session->tcp_flags |= tcphdr->th_flags;

dis0ps:
I was seeing consistent values way above the threshold of correct flags.  In most cases I was seeing values of 240, 80, and the like.  I'm testing with fprobe currently, and Moloch seems to be giving values that agree with it after the fix.

I agree, using the library's dissection of the packet is ideal.  This is my first foray into delving deep into C code, so I may use this to explore libnids a bit further.

awick:
yep I can repo, 13 is right, I think Ill probably still switch to th_flags since the rest of the code uses the th values, not sure why that didn't :)

PR #395, opened 2015-06-08
Enable batch build in scripts

kost:
The code still supports old interactive mode of scripts, but also this patch enables that moloch can be built automatically, something like:

USEPFRING=no ESMEM="512M" DONOTSTART=yes USERNAME=daemon GROUPNAME=daemon PASSWORD=0mgMolochRules1 INTERFACE=eth0 BATCHRUN=yes ./easybutton-singlehost.sh

Something like this helps in automatically building lxc/docker/vm images.

### DIFF
--- a/easybutton-config.sh
+++ b/easybutton-config.sh
@@ -9,11 +9,11 @@ fi
 
 clear
 
-if [ -z $USERNAME ]; then
+if [ -z $MOLOCHUSER ]; then
 	echo -n "Moloch service userid: [daemon] "
-	read USERNAME
+	read MOLOCHUSER
 fi
-if [ -z $USERNAME ]; then USERNAME="daemon"; fi
+if [ -z $MOLOCHUSER ]; then MOLOCHUSER="daemon"; fi
 
 if [ -z $GROUPNAME ]; then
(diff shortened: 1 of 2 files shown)


awick:
USERNAME is probably bad since thats set in linux.

A more meta question is why not just build it with your own scripts, the easy button stuff isn't required.

kost:
USERNAME is already used in the original script, but I can change that. 

Regarding meta question: i don't wanna build my own tree and scripts - would be good to stay as close as possible with original. I need that for docker image I've built (as I didn't see any):
https://github.com/kost/docker-moloch/

And currently I'm using my own branch:
https://github.com/kost/docker-moloch/blob/master/scripts/buildmoloch.sh

Would like to use original repository/branch (aol/moloch).

kost:
Renamed USERNAME to MOLOCHUSER in additional commit/push. So, I guess it is ready to go ;)

awick:
Not sure this is really the direction to go, but folks seem to want it.  Wasn't asking you to maintain a branch but all this script does is build it using normal configure and make sure dependancies are installed.  Must be a better way ...

PR #411, opened 2015-07-23
Wise 5tuple support

mak:
This introduces support for tagging exact connections 
(using standard tuple - proto:src_ip:src_port,:dst_ip:dst_port),
which can be used for example, to add references for ssl-stripped streams


awick:
sorry for delay was on vacation, looks good although why the underscore for _5tuple?

mak:
I think you cant have number at the begging of names in js or c

awick:
You have to use quotes, kind of a pain.

awick:
I reverted this for a couple reasons
* Since it isn't cacheable, except for probably udp, should be disabled by default
* Formatting should be fixed to match, and _5tuple changed to 5tuple or many folks use tuple5 to deal with leading number
* needs tests (which I can do if required)

mak:
I'll be gone for next few weeks, so if you can handle those changes it will be great
if not I'll get back to you in September ;]

PR #427, opened 2015-11-07
introduce quiet flag to disable regular logging

jbremer:
Disable the logging to reduce noise when used in an automated fashion.


awick:
Doesn't need to be a function like debug since it isn't incrementing.  You want to fix or should I?

jbremer:
Go ahead :-)

PR #428, opened 2015-11-13
Added node distance selection to the connections graph

DesertStorm:
In the connections tab, I added a drop down for selecting the distance of the d3 force variable, in order to make some dense graphs more readable.


awick:
Cool, do you think we could detect a better default then 120 based on number of nodes and/or links?

DesertStorm:
It's difficult because the number of nodes and the size of the graph varies quite drastically.

PR #431, opened 2015-11-25
Added connection highlighting on mouse hover/click

DesertStorm:
Added connection highlighting on mouse hover/click, like demonstrated here: http://bl.ocks.org/eyaler/10586116#index.html
This highlights only the first-level connection node.


awick:
very cool, thanks

PR #434, opened 2015-12-01
Fixed the node popup on mousehover

DesertStorm:
Adding the highlight broke node popup on the mousehover, fixed it by adding the highlighting to the original popup event listener.

### DIFF
--- a/viewer/views/connections.jade
+++ b/viewer/views/connections.jade
@@ -313,6 +313,7 @@ block content
                 window.clearTimeout(popupTimer);
               }
               popupTimer = window.setTimeout(showNodePopup, 800, this, d);
+              set_highlight(d);
             }).on("mouseout",function(d){
               window.clearTimeout(popupTimer);
             })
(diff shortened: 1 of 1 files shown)


viewer/views/connections.jade
```diff
@@ -352,10 +353,7 @@ block content
             focus_node = null,
             highlight_node = null;
 
-        node.on("mouseover", function(d) {
-          set_highlight(d);
-        })
-        .on("mousedown", function(d) {
+        node.on("mousedown", function(d) {
```
awick: Hmm, this shouldn't need to change

node.on().on()  should work vs
node.on(); node.on();

DesertStorm:
Well, we have "mouseover" event defined 2 times, although the expected result would be that there would be 2 mouseover events, there was only one which was later defined. Then I just added the set_highlight to the original mouseover event handler which made the node info popup appear again and work with the highlight.

PR #435, opened 2015-12-01
Connections graph node hover fix mouseout

DesertStorm:
Added a fix for the mouseout also to clear the popupTimer.


DesertStorm:
Sorry for double pull requests, wanted to merge it to my branch instead first.

PR #446, opened 2016-02-26
Adding report fragments

DesertStorm:
Added support for writing own custom report fragments (both query and session). Also wrote 2 simple examples.

### DIFF
--- a/viewer/public/common.js
+++ b/viewer/public/common.js
@@ -785,6 +785,42 @@ $(document).ready(function() {
     });
     return false;
   });
+  
+  //////////////////////////////////////////////////////////////////////////////////
+  // Reports
+  //////////////////////////////////////////////////////////////////////////////////
+  $("#queryReport").click(function (e) {
+    var that = this;
+    showActionsDialog({
+      submit: "Download report",
+      title: "Report - " + $(this).attr("label"),
+      input: "Filename",
+      query: "Export",
+      defaultInput: $(this).attr("label") + ".html"
+    }, function(qs, ids, filename) {
+      if (ids) {
+        qs.push({name: "ids", value: ids});
+      }
+      window.location = "queryReport/" + $(that).attr("data-filename") + "/" + filename + "?" + $.param(qs);
+    });
+    return false;
+  });
+  
+  $("#sessionReport").click(function (e) {
+    var that = this;
+    showActionsDialog({
+      submit: "Download report",
+      title: "Report - " + $(this).attr("label"),
+      input: "Filename",
+      query: "Export",
+      defaultInput: $(this).attr("label") + ".html"
+    }, function(qs, ids, filename) {
+      qs.push({name: "ids", value: $(".sessionDetail").attr("sessionid")});
+
+      window.location = "sessionReport/" + $(that).attr("data-filename") + "/" + filename + "?" + $.param(qs);
+    });
+    return false;
+  });
 
   //////////////////////////////////////////////////////////////////////////////////
   // Send Session Dialog
--- a/viewer/viewer.js
+++ b/viewer/viewer.js
@@ -755,7 +755,11 @@ app.get('/settings', checkWebEnabled, function(req, res) {
       }
       actions = actions.sort();
       if (typeof user.fragments !== 'object')
-        user.fragments = {};
+        user.fragments = {
+          query: {},
+          session: {}
+        };
+        
       res.render('settings', {
         user: req.user,
         suser: user,
@@ -3592,12 +3596,13 @@ app.post('/uploadFragment', checkWebEnabled, fragmentUpload, function(req, res) 
     }
     
     var fragmentName = req.body.fragmentName;
+    var identifier = req.body.identifier;
     if (fragmentName.length < 3 && fragmentName.length > 20) {
       return error("Invalid fragment name.");
     }
     
-    if (req.files.fragment[0].originalname.indexOf(' ') !== -1) {
-      return error("Fragment filename must not contain spaces.");
+    if (identifier.indexOf(' ') !== -1 && identifier.length > 12) {
+      return error("Fragment identifier must not contain spaces and be 12 or less characters.");
     }
     
     user = user._source;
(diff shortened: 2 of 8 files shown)


WilliamSalusky:
Can't wait to get home to have a look. (fistpump)

viewer/viewer.js
```diff
@@ -3580,6 +3585,204 @@ app.post('/updateUser/:userId', checkToken, function(req, res) {
...
+    var fragment = {
+      body: file.buffer.toString(),
+      name: fragmentName,
+      type: req.body.fragmentType,
+      filename: file.originalname
+    };
+    if (typeof user.fragments !== 'object') {
+      user.fragments = {};
+    }
+    user.fragments[fragment.filename] = fragment;
```
awick: can't use fragment.filename, since that may have a dot.  Shouldn't it be keyed off of fragmentName?

viewer/viewer.js
```diff
@@ -3580,6 +3585,204 @@ app.post('/updateUser/:userId', checkToken, function(req, res) {
   });
 });
 
+app.post('/uploadFragment', checkWebEnabled, fragmentUpload, function(req, res) {
+  Db.getUser(req.user.userId, function(err, user) {
```
awick: all these functions that call error need a 

  function error(text) {
    return res.send(JSON.stringify({success: false, text: text}));
  }

viewer/views/layout.jade
```diff
@@ -154,6 +154,14 @@ html
         - if (molochClusters)
           each key in Object.keys(molochClusters)
             command.sendSessionButton(cluster=key,label="Send Session to #{molochClusters[key].name}")
+        - if (titleLink == "sessionsLink")
+          span 
+            b Reports
```
awick: Maybe move b Reports inside the if, so it doesn't show up?

viewer/views/layout.jade
```diff
@@ -166,6 +174,13 @@ html
         - if (molochClusters)
           each key in Object.keys(molochClusters)
             command.sessionSendSessionButton.sendSessionAction(cluster=key,label="Send Session to #{molochClusters[key].name}")
+        span 
+          b Reports
+        - if (user.fragments)
```
awick: should b Reports be inside the if

viewer/viewer.js
DesertStorm: I went with filename because that could be more unique, I would guess that session/query fragment names could clash more often than filenames, although that's pure estimation. Also, it can be used, in JavaScript you can have an array with a key which has a dot in it, as long as you use the bracket syntax to access it.

viewer/views/layout.jade
DesertStorm: Good idea, but since the type of the fragment lies in the object itself (and JS currently has no predicate/LINQ-like query method to check whether the collection contains an object with certain property with a certain value) I could do separate collections for query fragments and session fragments, so it would make the checking on the template easier. Question is, is it worth it?

viewer/viewer.js
DesertStorm: What kind of functions do you mean? The ones that return errors instead of doing res.send?

awick: yes any method that has error(), either needs the function error defined or do the res.send yourself

viewer/viewer.js
awick: javascript yes, ES no.

DesertStorm:
I will make a new pull request with the specified changes.

viewer/public/common.js
```diff
@@ -785,6 +785,42 @@ $(document).ready(function() {
     });
     return false;
   });
+  
+  //////////////////////////////////////////////////////////////////////////////////
+  // Reports
+  //////////////////////////////////////////////////////////////////////////////////
+  $("#queryReport").click(function (e) {
```
linphoebe: when more then one report exists, only the first one can be used
ex: report2, report3 can't be used
report
  report1
  report2
  report3

recommend:
$(".queryReport").click(function (e) {

viewer/public/common.js
```diff
@@ -785,6 +785,42 @@ $(document).ready(function() {
...
+    }, function(qs, ids, filename) {
+      if (ids) {
+        qs.push({name: "ids", value: ids});
+      }
+      window.location = "queryReport/" + $(that).attr("data-filename") + "/" + filename + "?" + $.param(qs);
+    });
+    return false;
+  });
+  
+  $("#sessionReport").click(function (e) {
```
linphoebe: when more then one report exists, only the first one can be used
ex: report2, report3 can't be used
report
--report1
--report2
--report3

recommend:
$(".sessionReport").click(function (e) {

viewer/views/layout.jade
```diff
@@ -154,6 +154,13 @@ html
         - if (molochClusters)
           each key in Object.keys(molochClusters)
             command.sendSessionButton(cluster=key,label="Send Session to #{molochClusters[key].name}")
+        - if (user.fragments && user.fragments.query)
+          - var keys = Object.keys(user.fragments.query).sort()
+          - if (titleLink == "sessionsLink" && keys.length > 0)
+            span 
+              b Reports
+            each key in keys
+              command#queryReport(label="#{user.fragments.query[key].name}", data-filename="#{user.fragments.query[key].identifier}")
```
linphoebe: when more then one report exists, only the first one can be used
ex: report2, report3 can't be used
report
--report1
--report2
--report3

recommend:
command(class="queryReport", label="#{user.fragments.query[key].name}", data-filename="#{user.fragments.query[key].identifier}")

viewer/views/layout.jade
```diff
@@ -166,6 +173,13 @@ html
         - if (molochClusters)
           each key in Object.keys(molochClusters)
             command.sessionSendSessionButton.sendSessionAction(cluster=key,label="Send Session to #{molochClusters[key].name}")
+        - if (user.fragments && user.fragments.session)
+          - var keys = Object.keys(user.fragments.session).sort()
+          - if (keys.length > 0)
+            span 
+              b Reports
+            each key in keys
+              command#sessionReport(label="#{user.fragments.session[key].name}", data-filename="#{user.fragments.session[key].identifier}")
```
linphoebe: when more then one report exists, only the first one can be used
ex: report2, report3 can't be used
report
--report1
--report2
--report3

recommend:
command(class="sessionReport", label="#{user.fragments.session[key].name}", data-filename="#{user.fragments.session[key].identifier}")

PR #477, opened 2016-05-22
Fixes tag displaying in session detail

TNogisto:
Filename displaying seems to be still broken.


awick:
Is this needed because tags are broken sometimes?

Definitely need to support old sessions written out before the new field with something like

if (session.tags-terms)
  the new way
else
  the old way

razuz:
made a fix with old and new way ... testing it still a bit, but will push it here once we've confirmed it works

PR #512, opened 2016-08-03
support external cache, cache now at service level

awick:
* move caching from individual sources to service level
* support redis or memory for caching now
* jshint cleanup


31453:
👍 (except it could use some code comments 😄)

awick:
there are a few in there :)

PR #513, opened 2016-08-12
Record stats every 10 min too, and allow stat time period selection

awick:


viewer/views/stats.jade
```diff
@@ -767,8 +771,10 @@ block content
         } );
       }
 
-      $('#statsGraphSelect').change(function() {
-        makeStatsGraph($('#statsGraphSelect').val());
+      $('#statsGraphSelect,#statsTimeSelect').change(function() {
+        var interval = + $('#statsTimeSelect').val();
```
31453: I'd use `parseInt` here instead of `+`. Only because (1) more people understand parseInt, and (2) it's a teensy tiny bit faster in most browsers. Also, `+` is technically closer to `parseFloat` since it also accepts decimals. But really, for this case, it doesn't matter because we only have a few known values.

31453: 👍  LGTM

31453:
👍

PR #516, opened 2016-08-18
Update README.rst

necrose99:
https://github.com/VirusTotal/yara/releases/tag/v3.5.0 , yara's been long on GitHub since google-codes basically on historical freezer mode.  ,  also (part of Pentoo, / Gentoo / Sabayon perhaps forking a copy to  easy-button-Sabayon/Gentoo (if equo exisits (Sabayon get the binary ) Else emerge and build them)  will be on my too-do list latter. ) 
https://github.com/VirusTotal/yara-python also recommendable. 
however work loves Redhat

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -167,7 +167,7 @@ Building Capture
 
      + `yara <https://github.com/VirusTotal/yara>`_ version 1.6 or higher::
 
-            wget https://github.com/VirusTotal/yara/archive/v3.5.0.tar.gz -O yara-3.5.0.tar.gz [//]: # Fix file versioning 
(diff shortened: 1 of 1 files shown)


awick:
Have you tested building Moloch with 3.5?

necrose99:
I've used newer RPM's based on github , so far Easy-button builds on Centos. forensics repo /epel 
I can try from all sources since I have to migrate ome of our security vm's for production anyhow. 

for Gentoo / Pentoo I can make a run at home , in VM, if I ever can get time I could make an ebuild for Moloch (package file in bash)  econfgiure emake 
most of the deps are rather common in Gentoo or [@Pentoo ](https://github.com/pentoo/)  
Yara 3.x or has been in Pentoo for at least a year or more.

PR #525, opened 2016-09-07
Handle extra semicolons in pcapDir

MattCarothers:
This skips empty elements in the pcapDir list to gracefully handle the case
where pcapDir has extraneous semicolons.  E.g.:

pcapDir=/dir1;/dir2;


awick:
can you resubmit with just the change?  right now this seems to have your other pull request in it

PR #529, opened 2016-09-08
Adds two new algorithms for selecting pcap directories

MattCarothers:
Selecting pcap directories by round robin results in data expiring before it really needs to when the directories are on differently-sized filesystems.  This adds two options to select the directory based on available space.

Adds a new config file key "pcapDirAlgorithm."  Valid values:

- round-robin: uses each directory in sequence and then starts over at the beginning of the list
- max-free-percent: chooses the directory on the filesystem with the highest percentage of free disk space
- max-free-bytes: chooses the directory on the filesystem with the highest number of available bytes

The default is "round-robin," which is the original behavior.

### DIFF
--- a/capture/config.c
+++ b/capture/config.c
@@ -380,6 +380,14 @@ void moloch_config_load()
         exit(1);
     }
 
+    config.pcapDirAlgorithm = moloch_config_str(keyfile, "pcapDirAlgorithm", "round-robin");
+    if (strcmp(config.pcapDirAlgorithm, "round-robin") != 0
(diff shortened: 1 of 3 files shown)


awick:
awesome!

 few comments but I'll push anyway :)
1) I would probably convert the possible values to a int so that moloch_db_create_file can be just a case instead of strcmp everytime
2) would have to think about how much work it would be but FOO-weight options might be useful where it also somehow take into account both disk space AND how many other threads are writing to that disk at once.  (Right now if you add a new disk all the threads will write to it.)
3) Can you add to https://github.com/aol/moloch/wiki/Settings

PR #541, opened 2016-10-07
redirect / to new ui

awick:


31453:
LGTM

Soon, we should probably add links to the other pages in the navbar. They'll go to the old UI. Then, the "Sessions" link on the old UI, will go to the new UI.

PR #547, opened 2016-10-19
Make multiple capture units easier with systemd

jpvlsmv:
With this change, monitoring multiple network interfaces (on a systemd-based init) is easier:

Add a node in config.ini:

```
[moloch-enp0s3]
interface=enp0s3
```

and `systemctl add-wants multi-user.target molochcapture@enp0s3.service`

The changes look big, but that's mostly renaming the service file and replacing MOLOCH_INTERFACE with %i


awick:
Definitely want to take parts of this but to be honest I'm
1) Undecided if installing the package should auto enable, maybe that could be a Configure question.  If we do it, it needs to do for centos also.  Agree the documents could have those commands.
2) Undecided if it should always use systemd journal, maybe again that could be a Configure question.

jpvlsmv:
I would leave both those decisions to /data/moloch/bin/Configure.  I'll add that to this pull when I have time

PR #566, opened 2016-12-12
update path to ethtool

Joeyn414:
I've been deploying this script on a ubuntu 16.04 machine and it continually fails to start because the new path to ethtool is in /sbin/ not in /usr/sbin/ So with this simple fix the start script works again.


awick:
thanks

PR #578, opened 2017-01-05
Add TLSv1.3 and TLSv1.3-draft parsing

pwo:
- Add support for parsing upcoming TLSv1.3 and TLSv1.3-draft versions, able to extract protocol version and cipher selected by the server.
- Also fixes compile issue on FreeBSD with clang 3.9.1

The test-case is created blindfolded since I haven't figured out the test system yet.


awick:
Thanks! To generate the test file next time you just need to do "./tests.pl --make pcap/file.pcap"

PR #614, opened 2017-02-01
Fix viewer authentication user error handling

csash:
Right now, the general case error handling code is above the specific case error handling code and prevents it from being reached. This PR flips the order to allow the more specific (and helpful) error messages to be shown where possible.


awick:
Can you give an example of when this can happen or helpful?
What version of ES?
My concern is if the ES is down, then err should be set, but you will get some of the other errors first.

PR #632, opened 2017-02-15
Improve easybutton-build for non-root on Debian-based

jpvlsmv:
If run as a non-root user, easybutton-build.sh will fail due to the call to apt-get.  I'd rather not run the build as root, especially since I've configuration-management'd all of the required packages in place anyway.  This pull will check that all the requisite packages are installed, and bypass the call to apt-get if it not needed.


awick:
Easybutton is on life support and will be removed soon.  We want folks using the rpms/debs, probably should have something that help builds it if folks don't want to use it, but it won't be easy button.

PR #657, opened 2017-03-06
Ansible fixes for automatic builds of packages

jpvlsmv:
Revise Ansible configuration to successfully build .deb and .rpm for the 4 platforms.

* Install the RVM role only for this build's use (does not require root)
* Centos-6 is the only one that needs to upgrade Ruby (ships with 1.8, fpm requires >1.9)
* The Ruby development packages are required to build the native code for fpm

With this, `vagrant up` (or `vagrant provision`) successfully creates the packages.

### DIFF
--- a/release/build.yml
+++ b/release/build.yml
@@ -1,5 +1,6 @@
 ---
-# Requires running: sudo ansible-galaxy install rvm_io.ruby
+# Requires running: ansible-galaxy install -p `pwd`/roles rvm_io.ruby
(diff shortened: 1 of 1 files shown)


jpvlsmv:
Apparently ansible attempts to parse/execute a role even if the when: clause isn't satisfied.  This patch is neither necessary nor sufficient.

Delete the "check_mode: yes" line (line 19ish) of roles/rvm_io.ruby/tasks/rubies.yml

PR #692, opened 2017-05-16
Allow on-the-fly updates of the GeoIP databases

bladeswords:
This change should allow for on-the-fly updates to the GeoIP databases.  If the database is updated then the memory cache is reloaded.

https://github.com/maxmind/geoip-api-c/blob/0cfb181ff20ad9297e745318fdb9f484ba86cd0c/README.md#memory-caching-and-other-options

### DIFF
--- a/capture/db.c
+++ b/capture/db.c
@@ -2285,7 +2285,7 @@ void moloch_db_init()
     moloch_add_can_quit(moloch_db_can_quit, "DB");
 
     if (config.geoipFile) {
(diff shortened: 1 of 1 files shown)


awick:
Unfortunately GEOIP_CHECK_CACHE isn't thread safe.  Could probably make a set of databases per thread.

bladeswords:
Yeah...I suck at C... this might take a while...

awick:
gonna close for now

PR #712, opened 2017-07-06
Pcapfile pkt count limit

jmd1212:
This pull request adds the --pktcnt option with intended behaviour to be the same as tcpdump's -c option. When reading from pcap files, capture should read --pktcnt number of packets before stopping. Not specifying this option or when --pktcnt 0 all the packets in a file will be read.


awick:
Thanks, used some of your ideas in 6c6fdf6ff456f3c65b98bbcad065f7bcc51c9b62

PR #715, opened 2017-07-13
Save SPI View Fields

31453:
Allows a user to save the open fields in the SPI View tab for later use.


awick:
seems to work maybe

PR #717, opened 2017-07-17
Update to include extra required debian packages

bladeswords:
Needed to add some additional packages to get my vanilla Debian 9 to complete the easybutton-build.sh process.

### DIFF
--- a/easybutton-build.sh
+++ b/easybutton-build.sh
@@ -51,6 +51,8 @@ do
   esac
 done
 
+# Check the existance of sudo
+command -v sudo >/dev/null 2>&1 || { echo >&2 "MOLOCH: sudo is required to be installed"; exit 1; }
 
 MAKE=make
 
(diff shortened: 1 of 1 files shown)


awick:
Its to late to install sudo in a line that uses sudo.  autoreconf shouldn't be required since configure is already included.  Is something trying to run it?

bladeswords:
Ah, yes.  I didn't realize the sudo command was being run (I am running as root).  I might add it as a check to make sure it is there before hand.

As for the autoreconf - it failed when building yara.

```
2017-07-17 16:42:32 (313 KB/s) - 'yara/yara-3.6.1.tar.gz' saved [505302]

./bootstrap.sh: 2: ./bootstrap.sh: autoreconf: not found
./easybutton-build.sh: 122: ./easybutton-build.sh: ./configure: not found
make: *** No targets specified and no makefile found.  Stop.
```

bladeswords:
Ok, made some changes.  Should be good now.

PR #720, opened 2017-07-18
Add checks for location of node

bladeswords:
Added some additional checks for the location of node.  The default install didn't have node installed in the moloch directory, perhaps this happens in the package installs.

### DIFF
--- a/release/moloch_add_user.sh
+++ b/release/moloch_add_user.sh
@@ -1,3 +1,12 @@
 #!/bin/sh
 
-BUILD_MOLOCH_INSTALL_DIR/bin/node BUILD_MOLOCH_INSTALL_DIR/viewer/addUser.js -c BUILD_MOLOCH_INSTALL_DIR/etc/config.ini "$@"
+# Check the existance of a local copy of node and preference that local version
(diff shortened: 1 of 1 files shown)


bladeswords:
This issue is actually way larger.  There is lots of references to `MOLOCH_INSTALL_DIR/bin/node` which doesn't work if you installed nodejs yourself.  Would it be better to make this a change in the configure script?

awick:
Are you trying to fix doing a local "make;make install; make config"?   My assumption has been folks doing that are already creating their own scripts and such.  I would probably change the "make config" step to create soft links in /data/moloch/bin to node/npm, although it might be too late at that point since make install calls npm.

bladeswords:
Yes, I am doing what you say the "make;make install; make config" - which is why I am having troubles along the way.  So far the only issue (disregarding the previous pull requests) that I have had is with the scripts generated in the "make config" step.  They are all pointing at the wrong node - I think that a symlink might be the answer in this case.

awick:
so easybutton now installs node if not there, hopefullly will fix your issue.

PR #729, opened 2017-08-18
Fixes to readme and Configure script

ivernus:
Minor fix to Configure documentation and addition of CentOS7/ip compatible interface listing code.

### DIFF
--- a/release/Configure
+++ b/release/Configure
@@ -20,8 +20,7 @@ fi
 ### Ask config questions
 if [ -z "$MOLOCH_INTERFACE" ]; then
     echo -n "Found interfaces: "
-    if ! [ -f /sbin/ifconfig ];
-    then
+    if [ ! -f /sbin/ifconfig ]; then
         ip -o link | cut -f2 -d: | tr '\n' ' ' | sed 's/ \+/ /ig' | sed 's/ //1' | sed 's/[[:blank:]]*$//' | tr ' ' ';' | sed 's/$/\n/'
     else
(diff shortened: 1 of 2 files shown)


release/README.txt
```diff
@@ -5,8 +5,8 @@ Please report and bugs or feature requests by opening an issue at https://github
 Basic Moloch Installation steps:
  1) Download a Moloch build for 64bit Ubuntu 14.04, Ubuntu 16.04, Centos 6, or Centos 7 from http://molo.ch/index.html#downloads
  2) Install package
- 3) Configure basic moloch items (only need to do once)
-     /data/moloch/bin/Configure
+ 3) Configure basic moloch items by running the Configure script (this needs to be done only once)
+     . /data/moloch/bin/Configure
```
awick: Why the ". "?  Configure should be executable already

ivernus:
Hmm I'm not sure it wasnt on my system but I'll double check

release/Configure
```diff
@@ -20,7 +20,15 @@ fi
 ### Ask config questions
 if [ -z "$MOLOCH_INTERFACE" ]; then
     echo -n "Found interfaces: "
-    /sbin/ifconfig | grep "^[a-z]" | cut -d: -f1 | cut -d" " -f1 | paste -s -d";"
+    if [ -f /etc/redhat-release ];
```
awick: instead of checking for redhat should it just check for ifconfig so it might work on any OS without ifconfig?

ivernus:
Yep that makes sense @awick just made the changes now as requested

awick:
thanks!

Issue #740, opened 2017-10-04
left click extra menu options don't show up when date=X or start/stop missing


awick:
If you try and use the left click menu options and dont have date= or start/stop= in url then the extra items don't show up.  Add the date=1 and they do show up.

PR #742, opened 2017-10-05
Fix right click menus not appearing without a date=X param

MattCarothers:
This died if urlParams.date didn't exist:

isostart.setHours(isostart.getHours() - parseInt(urlParams.date));


awick:
fixes issue #740

PR #744, opened 2017-10-13
Add support for user specified histogram interval in API calls

MattCarothers:
e.g. https://viewhost/spigraph.json?interval=3600


awick:
Thanks, but already implemented similar but different.

PR #749, opened 2017-10-20
Wrap noisy non-errors under config.debug

jpvlsmv:
I think I remember talking about this on slack, but I can't find it right now.

Without this, moloch-capture is very chatty about every connection made or closed to Elasticsearch, generating lines like:

``Oct 20 09:11:46 capturehost moloch-capture[25686]: Oct 20 09:11:46 http.c:519 moloch_http_curl_close_callback(): Close 94/29 - eshost:9200   46348->146.xx.xx.xx:9200 fd:12 removed: true
Oct 20 09:11:46 capturehost moloch-capture[25686]: Oct 20 09:11:46 http.c:421 moloch_http_curl_watch_open_callback(): Connected 93/29 - eshost:9200   3966->146.xx.xx.xx:9200 - fd:12
``

This is just noise in normal operation, and should only be logged if debugging is on.

### DIFF
--- a/capture/http.c
+++ b/capture/http.c
@@ -411,7 +411,8 @@ static gboolean moloch_http_curl_watch_open_callback(int fd, GIOCondition condit
         strcat(remoteIp, "]");
     }
 
-    LOG("Connected %d/%d - %s   %d->%s:%d - fd:%d",
+    if (config.debug)
+        LOG("Connected %d/%d - %s   %d->%s:%d - fd:%d",
             server->outstanding,
             server->connections,
             server->names[0],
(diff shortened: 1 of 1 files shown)


awick:
They are very useful for debugging for new folks, may need a different var or something.  Are you really getting a lot of them?

jpvlsmv:
These are the only lines left if I exclude ERROR and WARNING, over 16 hosts, ~2.5M entries in 24h.  They are certainly useful for debugging.  Maybe tie them in with logESRequests, although they might be buried in that output.

awick:
added a new setting logHTTPConnections

PR #759, opened 2017-11-01
Systemd unit improvement

jpvlsmv:
2 more-or-less independent changes.  The NODE_ENV change is the more important


awick:
I like the constant restarting without a pause so I don't know the right solution.  closing for now

PR #772, opened 2017-11-21
Optionally index HTTP request bodies

ahodgen:


awick:
thanks, couple comments/questions

the db field needs to end with -term and then you can remove the raw section. 
Still debating if it should be more generic or just for http.
Instead of having off/on switch, what if we just use the number of chars to save, and default to 0 instead.  thoughts?

ahodgen:
I've corrected the db field name. I agree with the option change, and have amended the commit.

PR #792, opened 2018-01-09
Fix reversed icmp type and code

MattCarothers:
The ICMP type and code bytes were reversed.

¯\_(ツ)_/¯


awick:
it looks like this branch other stuff in it or something, github is confusing.  anyway, i will go ahead and put the fix in, thanks for finding

PR #806, opened 2018-02-23
allow reading of truncated packets for live captures

winstonliu1111:
Hi, I noticed that this fix was there for the offline capture case, but not for the live capture case, which was causing the program to exit whenever there is a truncated packet even when all the hw offload features are turned off.

### DIFF
--- a/capture/reader-libpcap.c
+++ b/capture/reader-libpcap.c
@@ -48,17 +48,21 @@ int reader_libpcap_stats(MolochReaderStats_t *stats)
 /******************************************************************************/
 void reader_libpcap_pcap_cb(u_char *batch, const struct pcap_pkthdr *h, const u_char *bytes)
 {
-    if (unlikely(h->caplen != h->len)) {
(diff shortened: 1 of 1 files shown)


awick:
I don't really want to add this.  Are you running in a VM?  did you turn off everything on the physical interface too?

winstonliu1111:
i'm running this within a docker container, where I have no access to the underlying docker host/infrastructure.

it's not turned on by default, it's an option, the default behavior is still the same as previously.

awick:
Yes, but people will find the option, and turn it on instead of fixing their deployment and then complain when it doesn't work right.  With truncated packets moloch will not be able to decode things correctly.

winstonliu1111:
yeah, i see your point about how people will turn it on without realizing the consequences and complain.  I was under the assumption that people understands the potential that certain sessions/flows will not be decoded correctly where truncated packets occur.  for someone who has no control over the underlying infrastructure (this can be VM on top of a physical machine in public cloud environments), this seemed like an acceptable tradeoff.  are most of the deployments/users today have access to their underlying infrastructure?  I'm just wondering if I'm the snowflake?

PR #817, opened 2018-03-13
fix moloch_field_define quic.version in moloch 1.0.0

jakup:
Updated moloch_field_define call for quic.version.  Removed "-term" suffix from field name.


awick:
thanks

PR #831, opened 2018-04-04
new:Clean up field values to remove URL information from queries on http.uri or host.* fields

MattCarothers:
Clean up user input by removing the http:// from http.uri queries and all URL information from host.* queries.  E.g.

http.uri = http://somehost.com/path becomes http.uri = //somehost.com/path
host.dns = http://somehost.com/path becomes host.dns = somehost.com

This helps new users, but more importantly it allows us to create right click pivots from URL fields.  For example, we can follow a redirect chain by pivoting from http.location to http.uri.

### DIFF
--- a/viewer/molochparser.js
+++ b/viewer/molochparser.js
@@ -511,9 +511,20 @@ function formatQuery(yy, field, op, value)
     throw "Invalid operator '" + op + "' for " + field;
   }
 
-  // Remove the "http:" from http.uri queries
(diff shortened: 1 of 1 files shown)


awick:
cool idea, couple things
* 1.0 also removed the //
* need to actually make changes to the jison file, that js file is autogenerated
* need to add some tests (i can help with that)

awick:
ok made it work with 1.0 with some other changes.  Thanks.

PR #832, opened 2018-04-05
support larger queries in /buildQuery.json

jakup:
Calling GET /buildQuery.json with a large query string (>100KB) returns a 400 Bad Request error.

This patch allows posting to the buildQuery.json route and increases the bodyParser.urlencoded size limit from 100KB to 5MB.


awick:
wow, what do you have that is > 100kb?  Amazing that even will work.

It didn't pass the lint tests above, can you use === and !== please and then I can merge.

PR #882, opened 2018-07-20
Update Configure

johnklasa:
Updated the Configure script to handle moloch installation device not having Internet connection


awick:
thanks for doing a PR, I needed a few more changes so I went ahead and just did it instead of having a conversation.

PR #902, opened 2018-08-23
fix Filename is empty

Eric-L90:
I use upload command to upload a test.pcap file and add a tag.But when i add Filename to visible columns in Sessions, the Filename is empty. So i try to add some codes and then solved this problem.


Eric-L90:
So when will Moloch show the file name ?

awick:
if you do a moloch-capture with either -r or -R and WITHOUT --copy

Eric-L90:
I use "upload command:  /data/moloch/bin/moloch-capture -r /tmp/9d5a0cc99db5c3ca0be4f939a5d95a29  --tag up2" to upload a pcap file.
But the file name  is still empty.

awick:
I just tried it and it worked fine.

PR #913, opened 2018-09-13
Packet Search

31453:


db/db.pl
```diff
@@ -2053,9 +2165,13 @@ sub dataNodes
         esDelete("/${PREFIX}tags", 1);
 
         checkForOld5Indices();
-    } elsif ($main::versionNumber <= 51) {
+    } elsif ($main::versionNumber < 52) {
+        createNewAliasesFromOld("users", "users_v6", "users_v5", \&usersCreate);
         sessions2Update();
         checkForOld5Indices();
+        fieldsUpdate();
```
awick: You need a huntsCreate() here.   Please test by running `./tests.pl -viewer` on master and then run `/db.pl http://localhost:9200 upgrade` and see if hunts work right.

viewer/viewer.js
```diff
@@ -4665,11 +4667,11 @@ app.post('/user/list', logAction('users'), function(req, res) {
     for (let i = 0, ilen = users.hits.hits.length; i < ilen; i++) {
       var fields = users.hits.hits[i]._source || users.hits.hits[i].fields;
       fields.id = users.hits.hits[i]._id;
-      fields.expression = fields.expression || "";
+      fields.expression = fields.expression || '';
       fields.headerAuthEnabled = fields.headerAuthEnabled || false;
       fields.emailSearch = fields.emailSearch || false;
       fields.removeEnabled = fields.removeEnabled || false;
-      fields.userName = safeStr(fields.userName || "");
+      fields.userName = safeStr(fields.userName || '');
```
awick: need a fields.packetSearch = fields.packetSearch || false

viewer/viewer.js
```diff
@@ -4784,14 +4787,15 @@ app.post('/user/update', logAction(), checkCookieToken, postSettingUser, functio
...
+    user.packetSearch = req.body.packetSearch === true;
 
     // Can only change createEnabled if it is currently turned on
     if (req.body.createEnabled !== undefined && req.user.createEnabled) {
       user.createEnabled = req.body.createEnabled === true;
     }
 
     Db.setUser(req.body.userId, user, function(err, info) {
-      console.log("setUser", user, err, info);
+      // console.log("setUser", user, err, info);
```
awick: lets wrap with a if debug since it used to be there

viewer/viewer.js
```diff
@@ -4969,138 +4973,606 @@ app.post('/removeTags', logAction(), function(req, res) {
 });
 
 //////////////////////////////////////////////////////////////////////////////////
-//// Search Session Add/Remove Tags
+//// Packet Search
 //////////////////////////////////////////////////////////////////////////////////
+function packetSearch (packet, options) {
+  if (options.searchType === 'asciicase') {
```
awick: switch? :)

viewer/viewer.js
```diff
@@ -4969,138 +4973,606 @@ app.post('/removeTags', logAction(), function(req, res) {
...
+  } else {
+    console.error('Invalid hunt search type');
+    return false;
+  }
+
+  return false;
+}
+
+function sessionHunt (sessionId, options, cb) {
+  function packetLoop (packets) {
```
awick: I gave you wrong info.  This src is even/dst is odd only works for reassembled.  For non reassembled you have to check each packet which it is. :(  see reassemble_* how it does a key and checks each packet.

viewer/viewer.js
```diff
@@ -4969,138 +4973,606 @@ app.post('/removeTags', logAction(), function(req, res) {
...
+      }
+
+      // do sessions query
+      Db.search('sessions2-*', 'session', query, function getMoreUntilDone (err, result) {
+        if (err || result.error) {
+          pauseHuntJobWithError(huntId, hunt, { value: `Hunt error searching sessions: ${err}` });
+          return;
+        }
+
+        let hits = result.hits.hits;
```
awick: We should check if the results total doesn't match the total we previously had in the hunt, and if they don't match update the hunt so the percent will work right.  (or you can just always update when you update the stats if easier)

viewer/viewer.js
```diff
@@ -4969,138 +4973,606 @@ app.post('/removeTags', logAction(), function(req, res) {
...
+    console.log('HUNT - processing hunt jobs');
+  }
 
-    var doit = false;
-    if (fields.tags) {
-      for (let i = 0, ilen = tags.length; i < ilen; i++) {
-        if (fields.tags.indexOf(tags[i]) === -1) {
-          doit = true;
-          break;
+  if (!internals.proccessHuntJobsInitialized) {
```
awick: don't need these 3 lines

viewer/viewer.js
```diff
@@ -4969,138 +4973,606 @@ app.post('/removeTags', logAction(), function(req, res) {
...
-      searchAndTagList(req, tags, list, function () {
-        return res.send(JSON.stringify({success: true, text: "Tags added successfully"}));
-      });
+});
+
+app.delete('/hunt/:id', logAction('hunt/:id'), checkCookieToken, function (req, res) {
+  if (Config.get('multiES', false)) { return res.molochError(401, 'Not supported in multies'); }
+  if (!req.user.packetSearch) { return res.molochError(403, 'Need packet search privileges'); }
+
+  function deleteHunt (id) {
```
awick: if the hunt is running should we still allow delete?

viewer/viewer.js
```diff
@@ -4969,138 +4973,606 @@ app.post('/removeTags', logAction(), function(req, res) {
...
+        return res.molochError(500, error);
+      });
+  } else { // admin can delete anyone's hunt
+    return deleteHunt(req.params.id);
+  }
+});
+
+app.put('/hunt/:id/pause', logAction('hunt'), checkCookieToken, function (req, res) {
+  if (Config.get('multiES', false)) { return res.molochError(401, 'Not supported in multies'); }
+  if (!req.user.packetSearch) { return res.molochError(403, 'Need packet search privileges'); }
```
awick: should non admin be allowed to pause other folks hunts?

viewer/viewer.js
```diff
@@ -4969,138 +4973,606 @@ app.post('/removeTags', logAction(), function(req, res) {
...
+  updateHuntStatus(req, res, 'paused', 'Paused hunt item successfully', 'Error pausing hunt job');
+});
+
+app.put('/hunt/:id/play', logAction('hunt'), checkCookieToken, function (req, res) {
+  if (Config.get('multiES', false)) { return res.molochError(401, 'Not supported in multies'); }
```
awick: should non admin be allowed to play other folks hunts?

viewer/viewer.js
31453: I make sure that the hunt job process is paused before deletion. And I have a check for processing a hunt that might be deleted. So there are checks to make sure it works without error. However, from a user interaction perspective, maybe you should not be able to delete a running session? Then, a user would have to pause, THEN delete the session?

db/db.pl
31453: I think it also needs usersUpdate() because of the addition of the packetSearch flag

awick: the createNewAliasesFromOld("users", "users_v6", "users_v5", \&usersCreate); should take care of the users table

db/db.pl
```diff
@@ -2053,9 +2165,15 @@ sub dataNodes
...
+        usersUpdate();
```
awick: so remove the usersUpdate

db/db.pl
```diff
@@ -1181,9 +1288,9 @@ sub usersCreate
   }
 }';
 
-    print "Creating users_v5 index\n" if ($verbose > 0);
-    esPut("/${PREFIX}users_v5", $settings);
-    esAlias("add", "users_v5", "users");
+    print "Creating users_v6 index\n" if ($verbose > 0);
```
awick: Actually when adding a field, you don't need to increment version number.  This should stay v5

PR #922, opened 2018-09-27
Changing CONTRIBUTING.md with more specific details on how to start 

srinathman:
Hi Andy, Elyse,

Since I've run into some issues when trying to start a dev environment here, I changed CONTRIBUTING.md with more specific details on how to start taking into account the problems I run into and with some tips you guys gave me on slack.

Hopefully it will make like easier for other potential contributors.


awick:
Thanks for the PR, however I've been thinking for a while that we should just make a new make target that does all the installs.

Also on what system are you having to do PERL5LIB settings?

srinathman:
Yes, that makes a lot of sense.

I'm on a ubuntu 18.04, but I believe I had the same issue on a 16.04.

PR #955, opened 2018-11-05
Adding environment variable to systemd unit file

srinathman:
- Modified parliament systemd file to load environment variables and asded $OPTIONS to command line.
  This will allow users to specify startup options, like --dashboardOnly, without having to change the systemd unit file

- Fixed the name on parliament service description (was wise)


awick:
Is it standard to have 2 EnvironmentFile lines?  Should we just do all the systemd files at once with this change?

srinathman:
For the 2 env. files, I can't say if it's standard but I've seen it in many other places.
It covers both rpm and debian based systems. 

And for other systemd unit files, I can submit the changes for them too if you want.

PR #958, opened 2018-11-14
bug find

kennafray:
On around like 284 
    export PATH=$TDIR/bin:$PATH
    sudo -E make install
echo "MOLOCH: Installed, now type sudo make config'"
else
    echo "MOLOCH: Now type 'sudo make install' and 'sudo make config'"
fi

That three part code uses sudo combined with export. Do you see that?
Please read man sudo 
Look at the -E flag 
Seems like sudo doesn't pass environment variables. 

Can you all look at this? Open a bug and fix it?


awick:
I don't understand what your issue is.  Can you open an issue please.

PR #964, opened 2018-11-20
prevent packaging log/raw dirs, conditionally handing in Configure instead

### LINKED ISSUE #963: prevent overwriting "raw" directory permissions in rpm
Currently, the RPM releases overwrite any permissions on the `/data/moloch/raw/` directory.  It would be helpful if any changes that have been made to this and the `/data/moloch/logs/` directories were left intact during an upgrade.

I don't believe marking these with `%config` is the right method though.  They might be better conditionally created via a post-install scriptlet, only if they don't already exist.

It doesn't look like the spec file is in the repo anywhere, but I'd be happy to take a look, figure out the syntax, and submit a PR if the file is added.

philhagen:
Modifies all `fpm` commands to exclude the `logs` and `raw` directories, which are created in the administrative `Configure` process anyway.

This _should_ fix #963, but I do not have an fpm environment to test.

### DIFF
--- a/release/Configure
+++ b/release/Configure
@@ -132,9 +132,14 @@ else
 fi
 
 ################################################################################
-mkdir -p $MOLOCH_INSTALL_DIR/logs $MOLOCH_INSTALL_DIR/raw
-chown nobody $MOLOCH_INSTALL_DIR/logs $MOLOCH_INSTALL_DIR/raw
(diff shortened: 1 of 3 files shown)


yahoocla:

philhagen:
I've signed the CLA and clicking the Heroku URL reflects that - do I need to do something to trigger a re-check?

PR #979, opened 2018-12-18
Viewer Notifiers

31453:
Add the ability to notify via email, twilio, and slack for cron query matches.


db/db.pl
```diff
@@ -2292,6 +2307,9 @@ sub dataNodes
         sessions2Update();
         checkForOld5Indices();
         usersUpdate();
+    } elsif ($main::versionNumber <= 56) {
+        usersUpdate();
+        queriesUpdate();
```
awick: You need to add queriesUpdate for all the other if cases that don't have one already, also every if should have the sessions2Update() so instead of making a new section probably just change the if from <= 55 to <= 56

PR #990, opened 2019-01-09
Hunt notify

31453:
Issue a notification via the configured notification service when a hunt completes or pauses with an error.

Also remove history logging for the hunt/list endpoint.


31453:
Did I do the db.pl stuff correctly ❓ ❓ ❓

PR #999, opened 2019-01-18
Better connections

31453:
Upgrade to d3 v5!


awick:
The old version seemed to do a better job keeping labels from being on top of each other.  Is that possible with this?

31453:
I'll see what I can do.

PR #1006, opened 2019-01-29
Initial tcp session seq numbers with first packet values.

### LINKED ISSUE #1004: Capture does not process payload of TCP session with no SYN Flag
Hello,
moloch-capture does not process payload of a TCP session, which is in the PCAP file attached. 
I did debug and saw it is because session neither start with nor have a packet with SYN flag. 
I modified packet.c file and I think the problem is solved. You may decide better. Also, I did regression test and just single-packets test failed and it seems the new result is more correct.

**Elasticsearch version**:

**Moloch version**:
1.7
**OS name and version**:
Ubuntu 18.04
**How was Moloch built/installed:** (rpm, deb, easybutton, ...)
easybutton
**Provide logs, stack traces and steps to reproduce**:

[p1-session1.zip](https://github.com/aol/moloch/files/2803047/p1-session1.zip)
[resolve suggest.zip](https://github.com/aol/moloch/files/2803058/resolve.suggest.zip)

aminlatifi:
* Fixes #1004

### DIFF
--- a/capture/packet.c
+++ b/capture/packet.c
@@ -633,6 +633,10 @@ LOCAL void *moloch_packet_thread(void *threadp)
                     session->port1 = ntohs(tcphdr->th_sport);
                     session->port2 = ntohs(tcphdr->th_dport);
                 }
+
+                session->tcpSeq[0] = ntohl(tcphdr->th_seq);
+                session->tcpSeq[1] = ntohl(tcphdr->th_ack);
+
                 if (moloch_http_is_moloch(session->h_hash, sessionId)) {
                     if (config.debug) {
                         char buf[1000];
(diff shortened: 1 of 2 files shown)


yahoocla:

aminlatifi:
> Unfortunately this won't work if you get packets out of order. Guess will have to add some test cases for that if we try and support this.
> 
> In order to deal with this I think the fix will either
> 
>     * have to be in moloch_packet_tcp_free where it checks for a single packet and instead does more
> 
>     * probably better in the maxTcpOutOfOrderPackets where instead of giving up it just pretends it the first packet it has is the first packet.

Yes, thanks.

PR #1020, opened 2019-02-21
User time limit

### LINKED ISSUE #1017: Limit per user max search time period
Add a "Extra Setting" per user of the max search period that user is allowed to use.  
* UI should use the same values the search uses (for when we make that configurable in the future)
* Default to all
* The UI for the user should then not show the values the user isn't allowed to do
* viewer.js should enforce that a query isn't too wide
* If they do too big a search, in viewer change startTime to be stopTime - maxAllowed.

For example if I say bob can do 1 month queries then then last 6 months and All would disappear from their menu, and even if they played with the url it would not do a search that big.

31453:
Disallow users to exceed a set time range for queries

fixes #1017

### DIFF
--- a/db/db.pl
+++ b/db/db.pl
@@ -54,6 +54,7 @@
 # 57 - hunt notifiers
 # 58 - users message count and last used date
 # 59 - tokens
+# 60 - users time query limit
 
 use HTTP::Request::Common;
 use LWP::UserAgent;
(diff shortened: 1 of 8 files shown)


31453:
@awick can you review these things:

1. the db.pl stuff
2. the server side check for user time limit (see the beginning of the buildSessionQuery function)
3. try to break it!

please and thank youuuu

PR #1022, opened 2019-02-26
Remove old users

### LINKED ISSUE #1021: Auto disable moloch user if not used
Now that we have users lastUsed time, add an optional ability to auto disable accounts when they haven't used moloch for X days.

* New config.ini setting that specifies X days of nonuse to auto disable
* When setting is set, in the cron enabled viewer, periodic check all enabled non admin users with a lastUsed date set, and set any that haven't used for X days as disabled
* Only check once a day
* A person that gets reenabled will have to use it right away, or they will get disabled the next run

31453:
fixes #1021

### DIFF
--- a/db/db.pl
+++ b/db/db.pl
@@ -132,6 +132,8 @@ sub showHelp($)
     print "    --shardsPerNode <shards>   - Number of shards per node or use \"null\" to let ES decide, default shards*replicas/nodes\n";
     print "  optimize                     - Optimize all indices in ES\n";
     print "    --segments <num>           - Number of segments to optimize sessions to, default 1\n";
+    print "  rmusers <days>               - Remove users that have not been active\n";
+    print "      days                     - Number of days of inactivity (integer)\n";
     print "\n";
     print "Backup and Restore Commands:\n";
     print "  backup <basename>            - Backup important indices into a file per index, filenames start with <basename>\n";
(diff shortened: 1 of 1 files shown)


31453:
Halp! Unknown territory! 🆘

db/db.pl
```diff
@@ -2083,6 +2085,35 @@ sub parseArgs {
     }
     esPost("/_flush/synced", "", 1);
     logmsg "\n";
+    exit 0;
+} elsif ($ARGV[1] eq "rmusers") {
+    showHelp("Invalid number of <hours>") if (!defined $ARGV[2] || $ARGV[2] !~ /^[+-]?\d+$/);
```
awick: should say days not hours

db/db.pl
```diff
@@ -2083,6 +2085,35 @@ sub parseArgs {
...
+
+    my $users = esGet("/${PREFIX}users/_search?size=1000");
+    my $rmcount = 0;
+
+    foreach my $hit (@{$users->{hits}->{hits}}) {
+        if (exists $hit->{_source}->{lastUsed} && !$hit->{_source}->{createEnabled} && $hit->{_source}->{enabled}) {
```
awick: this works, but you could also just have es do this with something like (not sure totally right), but either way is fine.

/${PREFIX}users/_search?size=1000&q=_exists_:lastUsed+AND+createEnabled:1+AND+enabled:1

https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-query-string-query.html

db/db.pl
31453: but duh!

db/db.pl
31453: 👍

PR #1025, opened 2019-03-01
Install Java when using "easybutton"

hackacad:
Installation of Elasticsearch failed on CentOS/RHEL due to missing java (not a dependency for elasticsearch rpm).
Just added the installation of openjdk.


yahoocla:

awick:
We used to do this and then people who had other java installed didn't like it.  Maybe put an if around it?

hackacad:
Thought of that before but decided that the yum install prompt (another version of java already installed) might be okay as well.
Added a check for java

PR #1044, opened 2019-04-09
Rollback

codesniffer:
- The Backup feature is enhanced such that it can export settings, mappings, templates, and aliases in addition to documents.

- A Rollback feature is implemented to restore settings, mappings, templates, and documents of a specific DB version.

### DIFF
--- a/db/db.pl
+++ b/db/db.pl
@@ -137,6 +137,7 @@ sub showHelp($)
     print "\n";
     print "Backup and Restore Commands:\n";
     print "  backup <basename>            - Backup important indices into a file per index, filenames start with <basename>\n";
+    print "  export <index> <basename>    - Save a single index into a file, fielname starts with <basename>\n";
     print "  restore <filename>           - Restore single index\n";
     print "  rollback <basename> [<opts>] - Rollback to a particular version; filenames of settings, mappings, templates, and documents start with <basename>\n";
     print "    --skipupgradeall           - Do not upgrade Sessions\n";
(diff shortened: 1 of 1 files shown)


yahoocla:

db/db.pl
```diff
@@ -2461,6 +2497,147 @@ sub dataNodes
         usersCreate();
         queriesCreate();
     }
+} elsif ($ARGV[1] =~ /^rollback$/) {
+
+    logmsg "It is STRONGLY recommended that you stop ALL moloch captures and viewers before proceeding.\n";
+
+    waitFor("ROLLBACK", "do you want to rollback?");
```
awick: * Can you verify that at least a few of the files exist before even prompting.
* Can you change to something like "do you want to rollback?  This will delete ALL data but sessions and history and restore from backups"

db/db.pl
```diff
@@ -2461,6 +2497,147 @@ sub dataNodes
...
+    esDelete("/_template/${PREFIX}template_1", 1);
+    esDelete("/_template/${PREFIX}sessions_template", 1);
+    esDelete("/_template/${PREFIX}sessions2_template", 1);
+    esDelete("/_template/${PREFIX}history_v1_template", 1);
+
+    my @indexes = ("users", "sequence", "stats", "queries", "hunts", "files", "fields", "dstats");
+    logmsg "Importings settings...\n";
+    foreach my $index (@indexes) { # import settings
+        if (-e "$ARGV[2].$index.settings.json") {
+            open(my $fh, "<", "$ARGV[2].$index.settings.json") or die "cannot open > $ARGV[2].$index.settings.json: $!";
```
awick: Thoughts about removing the die and change to warn or something?

db/db.pl
codesniffer: Displayed the list of files used for the rollback operation.

codesniffer: Updated the message according to the suggestion.

db/db.pl
codesniffer: Removed the “die” statement as there is defensive codes to check a file is found in the directory before to open it.

db/db.pl
```diff
@@ -137,7 +137,10 @@ ($)
     print "\n";
     print "Backup and Restore Commands:\n";
     print "  backup <basename>            - Backup important indices into a file per index, filenames start with <basename>\n";
+    print "  export <index> <basename>    - Save a single index into a file, fielname starts with <basename>\n";
```
awick: fielname mispell

db/db.pl
```diff
@@ -1910,8 +1914,29 @@ sub parseArgs {
...
+        } else {
+            print $fh "{}\n";
+        }
+    }
+    close($fh);
+    exit 0;
 } elsif ($ARGV[1] =~ /^backup$/) {
-    foreach my $index ("users", "sequence", "stats", "queries", "hunts", "files", "fields", "dstats") {
+    my @indexes = ("users", "sequence", "stats", "queries", "files", "fields", "dstats");
+    @indexes = (@indexes, "hunts") if ($main::versionNumber > 51);
```
awick: this really should be push(@indexes, "hunts")

db/db.pl
```diff
@@ -2461,6 +2517,159 @@ sub dataNodes
...
+    logmsg "It is STRONGLY recommended that you stop ALL moloch captures and viewers before proceeding.\n";
+
+    dbCheckForActivity();
+
+    my @indexes = ("users", "sequence", "stats", "queries", "hunts", "files", "fields", "dstats");
+    my @filelist = ();
+    foreach my $index (@indexes) { # list of data, settings, and mappings files
+        @filelist = (@filelist, "$ARGV[2].$index.json\n") if (-e "$ARGV[2].$index.json");
+        @filelist = (@filelist, "$ARGV[2].$index.settings.json\n") if (-e "$ARGV[2].$index.settings.json");
+        @filelist = (@filelist, "$ARGV[2].$index.mappings.json\n") if (-e "$ARGV[2].$index.mappings.json");
```
awick: these should be push?

PR #1049, opened 2019-04-18
Handle RAW IPV4 pcap layertype

### LINKED ISSUE #661: LINKTYPE_RAW value
**Elasticsearch version**: 2.4.0

**Moloch version**:0.18.0-GIT

**OS name and version**: Ubuntu 16.04

**How was Moloch built/installed:**: easybutton-singlehost

In packet.c, I believe case 12 (RAW) for the switch(pcapFileHeader.linktype) should be instead case 101.

The link below is from tcpdump.org specifying that LINKTYPE_RAW has value 101. 
http://www.tcpdump.org/linktypes.html

Regards

pavja2:
Problem: moloch-capture fails to load a pcap file with the link layer header type 228 (RAW IPv4). However, the handler link layer header type 101 (RAW) works as written. See comments on #661 

Fix: Added a switch for handling pcap link layer header type 228 which is semantically the same as RAW but only contains IPv4 packets instead of both IPv4 and IPv6 packets.

Fixes #661

### DIFF
--- a/capture/packet.c
+++ b/capture/packet.c
@@ -1869,6 +1869,9 @@ void moloch_packet_batch(MolochPacketBatch_t * batch, MolochPacket_t * const pac
     case 127: // radiotap
         rc = moloch_packet_radiotap(batch, packet, packet->pkt, packet->pktlen);
         break;
+    case 228: //RAW IPv4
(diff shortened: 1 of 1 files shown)


yahoocla:

PR #1056, opened 2019-04-30
fix for the prefix restore issue

codesniffer:
**Problem statement:** Backup and restore features do not work when index PREFIX is provided
**Solution:** Backup and restore methods are enhanced to handle PREFIX 
**Changes:** The restore command is renamed to import and the rollback command to restore

### DIFF
--- a/db/db.pl
+++ b/db/db.pl
@@ -2004,9 +2004,9 @@ if ($ARGV[1] =~ /^(users-?import|import)$/) {
         logmsg "The index is empty\n";
         exit 0;
     }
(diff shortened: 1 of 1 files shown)


awick:
In general I like using ${PREFIX} in strings instead of just $PREFIX since its "safer".  
So is the prefix in all the file names now too?

codesniffer:
> In general I like using ${PREFIX} in strings instead of just $PREFIX since its "safer".
> So is the prefix in all the file names now too?

- will change the $PREFIX syntax to ${PREFIX}
- Yea the prefix will be appended with the basename of all the file names: basename.prefix_index.json

PR #1065, opened 2019-05-22
Add string fields for TLS JA3 and JA3S

binkboyd:
Add raw string fields for JA3 and JA3S

this patch adds new field which are the strings which the MD5 for JA3 and JA3S is computed from. This allows e.g removing/adding the SNI field or making your own hashes based on different fields in the HELLOs.

This can be enabled or disabled by setting the ja3Strings option in the config file to true/false. It's disabled by default.


yahoocla:

capture/parsers/tls.c
```diff
@@ -874,12 +884,24 @@ void moloch_parser_init()
         MOLOCH_FIELD_TYPE_STR_GHASH,  MOLOCH_FIELD_FLAG_CNT,
         (char *)NULL);
 
+    ja3StrField = moloch_field_define("tls","lotermfield",
```
awick: Can you move the 2 field creations inside an if with config.ja3Strings

binkboyd: sure; commited.

awick:
thanks

PR #1066, opened 2019-05-22
Lookups

### LINKED ISSUE #1042: Lookup tables/Variables
Allow the creation of lists of values that can be used in queries as variables.  So for example you could have a list of ips stored and then you could say ip.src == $MY_IPS or ip.src != $MY_IPS.  Won't support other operators nor arrays.
We should support whatever https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-terms-query.html#query-dsl-terms-lookup supports.
These lists should be sharable.

- [x] db.pl to create new index/table/alias of lookups_v1, lookup, lookups with mapping of at least something like name, ip, string, number, creator, shared, description and it should have 1 shard and replicate to all nodes
- [x] UI to edit create/list/delete lists . (Probably have a name, description, values, sharable attributes, list type)
- [x] UI to view, add, delete values in lists
- [x] Support 3 list types: ip, string, number
- [x] Lists should be sharable
- [x] Support autocomplete on $ in sessions query
- [x] Change DSL to support == $LISTNAME and generate correct term query
- [x] Add a command to db.pl that can create/update a lookup table

31453:
fixes #1042


db/db.pl
```diff
@@ -1511,6 +1522,63 @@ sub huntsUpdate
...
+################################################################################
+sub lookupsCreate
+{
+  my $settings = '
+{
+  "settings": {
+    "index.priority": 30,
+    "number_of_shards": 1,
+    "number_of_replicas": 0,
+    "auto_expand_replicas": "0-3"
```
awick: documentation suggests using 0-all

db/db.pl
```diff
@@ -1954,11 +2022,11 @@ sub parseArgs {
             $pos++;
             $SEGMENTS = int($ARGV[$pos]);
         } elsif ($ARGV[$pos] eq "--nooptimize") {
-	    $NOOPTIMIZE = 1;
+	         $NOOPTIMIZE = 1;
```
awick: What craziness happen to all of these?  should be 4 spaces

db/db.pl
```diff
@@ -1973,6 +2041,15 @@ sub parseArgs {
         } elsif ($ARGV[$pos] eq "--warmafter") {
             $pos++;
             $WARMAFTER = int($ARGV[$pos]);
+        } elsif ($ARGV[$pos] eq "--shared") {
+            $pos++;
```
awick: no pos++ is needed since shared doesn't have an argument right?

db/db.pl
```diff
@@ -139,6 +143,13 @@ ($)
     print "    --segments <num>           - Number of segments to optimize sessions to, default 1\n";
     print "  disableusers <days>          - Disable user accounts that have not been active\n";
     print "      days                     - Number of days of inactivity (integer)\n";
+    print "  setshortcut <name> <userid> <file> [<opts>]  - Create/update shortcut\n";
+    print "       name                                    - Name of the shortcut (no special characters except '_')\n";
+    print "       userid                                  - UserId of the user to add the shortcut for\n";
+    print "       file                                    - File that includes a comma or newline separated list of values\n";
+    print "    --type                                     - Type of shortcut = string, ip, number, default is string\n";
```
awick: should be ```--type <type>``` and
```--description <description>```

File: db/db.pl [hunk 5]
```diff
@@ -2267,6 +2345,71 @@ sub parseArgs {
...
+      }
+      if ($existingShortcut->{_source}->{shared}) {
+        $newShortcut->{shared} = $existingShortcut->{_source}->{shared};
+      }
+    }
+    if ($DESCRIPTION) {
+      $newShortcut->{description} = $DESCRIPTION;
+    }
+    if ($SHARED) {
+      $newShortcut->{shared} = \1;
```
awick: what does \1 even do?  does just 1 work?

31453: \1 translates to true. "true" is a string, which we don't want and 1 is a number, which we don't want either
https://stackoverflow.com/questions/43867553/how-to-send-boolean-value-from-perl-script-without-converting-them-into-string

PR #1078, opened 2019-06-07
Update tags

31453:


viewer/db.js
```diff
@@ -399,6 +399,78 @@ exports.refresh = function (index, cb) {
   return internals.usersElasticSearchClient.indices.refresh({index: fixIndex(index)}, cb);
 };
 
+exports.addTagsToSession = function (index, type, id, tags, node, cb) {
```
awick: lets remove type from these apis and just assume 'session' in the code.  ES is moving away from types in ES 7.

PR #1094, opened 2019-06-23
check filenane permissions for possible viewer access issue

bf31415:
check filename permissions for possible view access issue

it's possible that when one ingests a capture but doesn't set the --copy flag, that permissions on the path to the capture flag might not be accessible when viewer runs as dropUser and dropGroup.  the purpose of this code change is to generate a warning if path permission issues might be present which would result in the viewer not being able to open the capture file.

1088

not relevant

ran ./tests.pl and all tests passed

## License

### DIFF
--- a/capture/reader-libpcap-file.c
+++ b/capture/reader-libpcap-file.c
@@ -182,9 +182,14 @@ LOCAL void reader_libpcapfile_init_monitor()
 LOCAL int reader_libpcapfile_process(char *filename)
 {
     char         errbuf[1024];
-    char 				 path [256];
-    char         *p;
-    struct 			 stat stats;
+    char         path[PATH_MAX];
+    struct       stat stats;
+    char         *token;
+    char         *save_ptr;
+    char         tmpFilename[PATH_MAX];
+    struct group *gr;
+    struct passwd *pw;
+
 
     if (!realpath(filename, offlinePcapFilename)) {
         LOG("ERROR - pcap open failed - Couldn't realpath file: '%s' with %d", filename, errno);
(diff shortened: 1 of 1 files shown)


capture/reader-libpcap-file.c
```diff
@@ -179,6 +182,9 @@ LOCAL void reader_libpcapfile_init_monitor()
 LOCAL int reader_libpcapfile_process(char *filename)
 {
     char         errbuf[1024];
+    char 				 path [256];
```
awick: Lets use PATH_MAX, and style nic pick path should line up with errbuf

capture/reader-libpcap-file.c
```diff
@@ -179,6 +182,9 @@ LOCAL void reader_libpcapfile_init_monitor()
...
+    char         *p;
+    struct 			 stat stats;
```
awick: style nic pick, please change to "struct stat    stats", where the stats lines up with the p

capture/reader-libpcap-file.c
```diff
@@ -196,6 +202,51 @@ LOCAL int reader_libpcapfile_process(char *filename)
         return 1;
     }
 
+    // check to see if viewer might have access issues to non-copied pcap file
+    if (config.copyPcap == 0) {
+
+      if (strlen (filename) > 255) {
```
awick: PATH_MAX

capture/reader-libpcap-file.c
```diff
@@ -196,6 +202,51 @@ LOCAL int reader_libpcapfile_process(char *filename)
...
+        // filename bigger than path buffer, skip check
+      } else if ((config.dropUser == NULL) && (config.dropGroup == NULL)) {
+        // drop.User,Group not defined -- skip check
+      } else if (strncmp (filename, "/", 1) != 0) {
+        LOG("WARNING using a relative path may make pcap inaccessible to viewer");
+      } else {
+
+    		path[0] = 0;
+
+        p = strtok (filename, "/");
```
awick: need to use strtok_r

bf31415: thanks for comments.  making changes.  i also realized I had these two vars defined in the middle of the function-- do you want the definitions at the top at the function level?

          if (stat(path, &stats) != -1) {
            struct group *gr = getgrgid (stats.st_gid);
            struct passwd *pw = getpwuid (stats.st_uid);

awick: nope that is fine there, other than the gr/pw should be lined up. :)

PR #1098, opened 2019-06-26
Enable implicit creation of users and group-authorization from headers

schenksj:
Addresses https://github.com/aol/moloch/issues/1095.   Enables implicit creation of users from HTTP headers when using HTTP header auth.  Assumes that authorization is performed upstream, and allows creation of a "default" user profile whose accesses can be overridden using the customary user management tools provided with moloch.  To enable this functionality, a user would enable header auth and provide a JSON document that describes the user as below:

```
# Enable implicit user creation and auth from header
userNameHeader=http_auth_http_user
userAutoCreateTmpl={"userId": "${this.http_auth_http_user}", "userName": "${this.http_auth_mail}", "enabled": true, "webEnabled": true, "headerAuthEnabled": true, "emailSearch": true, "createEnabled": false, "removeEnabled": false, "packetSearch": true }
```

Also added the capability to deny access to the application if a required Header w/a specified value is not present.  This is useful when the external authentication source (via. apache/sso) is also passing in a authorization group (like LDAP group membership, etc...).  The below sample make the app require a header `UserGroup: MOLOCH_ACCESS` before allowing header-based authentication to occur.
```
# Require that the UserGroup: MOLOCH_ACCESS HTTP header is present
requiredAuthHeader="UserGroup"
requiredAuthHeaderVal="MOLOCH_ACCESS"
```

## License

### DIFF
--- a/viewer/viewer.js
+++ b/viewer/viewer.js
@@ -291,6 +291,9 @@ if (Config.get("passwordSecret")) {
                    }
                  });
               }
+              if (suser) {
+                  userCleanup(suser._source);
+              }
               return next();
            });
 
--- a/viewer/viewer.js
+++ b/viewer/viewer.js
@@ -65,6 +65,7 @@ var internals = {
   elasticBase: Config.get("elasticsearch", "http://localhost:9200").split(","),
   esQueryTimeout: Config.get("elasticsearchTimeout", 300) + 's',
   userNameHeader: Config.get("userNameHeader"),
+  userAutoCreateTmpl: Config.get("userAutoCreateTmpl"),
   httpAgent:   new http.Agent({keepAlive: true, keepAliveMsecs:5000, maxSockets: 40}),
   httpsAgent:  new https.Agent({keepAlive: true, keepAliveMsecs:5000, maxSockets: 40, rejectUnauthorized: !Config.insecure}),
   previousNodesStats: [],
@@ -274,6 +275,27 @@ if (Config.get("passwordSecret")) {
     if (internals.userNameHeader !== undefined) {
       if (req.headers[internals.userNameHeader] !== undefined) {
         var userName = req.headers[internals.userNameHeader];
+        // Auto-creation of users from headers
+        if (internals.userAutoCreateTmpl !== undefined) {
+
+           // Check to see if we need to auto-create.
+           Db.getUserCache(userName, function(err, suser) {
+              if (err) {return res.send("ERROR - getUser - user: " + userName + " err:" + err);}
+              if (!suser || !suser.found) {
+                 var nuser = JSON.parse(new Function("return `" + internals.userAutoCreateTmpl + "`;").call(req.headers));
+                 var res = Db.setUser(userName, nuser, (err, info) => {
+                   if (err) {
+                     console.log("Elastic search error adding user: (" +  userName + "):(" + JSON.stringify(nuser) + "):" + err);
+                   } else {
+                     console.log("Added user:" + userName + ":" + JSON.stringify(nuser));
+                   }
+                 });
+              }
+              return;
+           });
+
+        }
+
         Db.getUserCache(userName, function(err, suser) {
           if (err) {return res.send("ERROR - getUser - user: " + userName + " err:" + err);}
           if (!suser || !suser.found) {return res.send(userName + " doesn't exist");}
--- a/viewer/viewer.js
+++ b/viewer/viewer.js
@@ -289,7 +289,7 @@ if (Config.get("passwordSecret")) {
                 authorized = true;
              }
           });
-          if(!authorized) {
+          if (!authorized) {
               return res.send("Not authorized");
           }
         }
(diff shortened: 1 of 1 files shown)


viewer/viewer.js
```diff
@@ -274,6 +275,27 @@ if (Config.get("passwordSecret")) {
...
       if (req.headers[internals.userNameHeader] !== undefined) {
         var userName = req.headers[internals.userNameHeader];
+        // Auto-creation of users from headers
+        if (internals.userAutoCreateTmpl !== undefined) {
+
+           // Check to see if we need to auto-create.
+           Db.getUserCache(userName, function(err, suser) {
+              if (err) {return res.send("ERROR - getUser - user: " + userName + " err:" + err);}
+              if (!suser || !suser.found) {
+                 var nuser = JSON.parse(new Function("return `" + internals.userAutoCreateTmpl + "`;").call(req.headers));
```
awick: Can you explain what this is doing?  Do you have a sample value you are trying to use so we can understand better?

viewer/viewer.js
```diff
@@ -274,6 +275,27 @@ if (Config.get("passwordSecret")) {
...
+                 var res = Db.setUser(userName, nuser, (err, info) => {
+                   if (err) {
+                     console.log("Elastic search error adding user: (" +  userName + "):(" + JSON.stringify(nuser) + "):" + err);
+                   } else {
+                     console.log("Added user:" + userName + ":" + JSON.stringify(nuser));
+                   }
+                 });
+              }
+              return;
```
awick: You need to actually call the userCleanup/next here too, since because of race conditions it might not hit the one below.  Actually I would probably switch the order around where you do the getUserCache always and then have the if internals.userAutoCreateTmpl !== undefined inside the results if the user isn't found.

viewer/viewer.js
awick: Oops, I see the example in the PR request.  let me think about it

schenksj: Do you want me to move the creation into the block with the original getUserCache?  I figured the penalty for calling getUserCache was minimal given that it is just a hashtable lookup when its a hit.

viewer/viewer.js
schenksj: Agreed, though wouldn't i need to call getUserCache again recursively in the cb from inside getUserCache?  Is there a good way to avoid code duplication (I'm admittedly new to javascript)

awick: I might not be following your question but the issue with the current code is when the user doesn't exist in cache its going to call ES twice at the same time to get the user, and the second one will fail, before the first one can create it.

What you have now is
```
if () {
  getUserCache() {  #1
    createUser
    return
  }
}
getUserCache () { #2
  //the user won't exit here
}

Comment (schenksj) [reply viewer/viewer.js]: oh, i see.  I will fix that...  I didn't realize that would go async like that...  i have some node learning to do i suppose

Comment (schenksj) [reply viewer/viewer.js]: I rewrote the algorithm now that i understand the `next()` concept.  Happy to make more modifications.

File: viewer/viewer.js [hunk 3]
```diff
@@ -274,7 +275,8 @@ if (Config.get("passwordSecret")) {
     if (internals.userNameHeader !== undefined) {
       if (req.headers[internals.userNameHeader] !== undefined) {
         var userName = req.headers[internals.userNameHeader];
-        Db.getUserCache(userName, function(err, suser) {
+
+        var ucb = function(err, suser) {
```
awick: this can just be `function ucb(err, suser)`

viewer/viewer.js
```diff
@@ -283,13 +285,34 @@ if (Config.get("passwordSecret")) {
           userCleanup(suser._source);
           req.user = suser._source;
           return next();
+        };
+
+        Db.getUserCache(userName, function(err, suser) {
+          if (internals.userAutoCreateTmpl === undefined) {
+             return ucb(err, suser);
+          } else if ((err && (err+"").includes("Not Found")) ||
```
awick: `err.toString().includes(` is considered cleaner even though the `err+""` will work

viewer/viewer.js
schenksj: Done.  I agree that the whole JSON in the ini file sounded better than it looks...  Functional but ugly...

viewer/viewer.js
schenksj: Updated.  I wasn't sure if err was always an object that supports toString(), but I'll take your word for it since this is the first time I've written JavaScript since my HTML4.0 class in 1998.

schenksj:
Hey @awick.  Do you want to eliminate the template in favor of a bunch of variables defining the values of the automatically creates user?

schenksj:
FYI: Added a capability to protect "implicit access" by requiring a defined HTTP header that contains a users LDAP group, etc...

schenksj:
Hello @31453 and @awick .  I have some other PR's to send your way, but I did this on my master branch... Should I close this PR and re-submit a fresh one against a branch so we can have more time to review?

awick:
you can do new PR if you want.  As we posted on slack we will be in and out the next few weeks for vacation.  Have you joined slack yet?

schenksj:
will re-open on a branch

PR #1104, opened 2019-06-28
fix mpls protocol parsing error

JackLiar:
Fix MPLS protocol parsing problem

It is the **third byte** that determines whether current mpls layer is at the stack bottom, not the fourth.

## License


awick:
thanks

PR #1120, opened 2019-07-08
Enable implicit creation of users and group-authorization from headers 

schenksj:
Addresses https://github.com/aol/moloch/issues/1095.   Enables implicit creation of users from HTTP headers when using HTTP header auth.  Assumes that authorization is performed upstream, and allows creation of a "default" user profile whose accesses can be overridden using the customary user management tools provided with moloch.  To enable this functionality, a user would enable header auth and provide a JSON document that describes the user as below:

```
# Enable implicit user creation and auth from header
userNameHeader=http_auth_http_user
userAutoCreateTmpl={"userId": "${this.http_auth_http_user}", "userName": "${this.http_auth_mail}", "enabled": true, "webEnabled": true, "headerAuthEnabled": true, "emailSearch": true, "createEnabled": false, "removeEnabled": false, "packetSearch": true }
```

Also added the capability to deny access to the application if a required Header w/a specified value is not present.  This is useful when the external authentication source (via. apache/sso) is also passing in a authorization group (like LDAP group membership, etc...).  The below sample make the app require a header `UserGroup: MOLOCH_ACCESS` before allowing header-based authentication to occur.
```
# Require that the UserGroup: MOLOCH_ACCESS HTTP header is present
requiredAuthHeader="UserGroup"
requiredAuthHeaderVal="MOLOCH_ACCESS"
```

## License

### DIFF
--- a/viewer/viewer.js
+++ b/viewer/viewer.js
@@ -65,6 +65,9 @@ var internals = {
   elasticBase: Config.get("elasticsearch", "http://localhost:9200").split(","),
   esQueryTimeout: Config.get("elasticsearchTimeout", 300) + 's',
   userNameHeader: Config.get("userNameHeader"),
+  requiredAuthHeader: Config.get("requiredAuthHeader"),
+  requiredAuthHeaderVal: Config.get("requiredAuthHeaderVal"),
+  userAutoCreateTmpl: Config.get("userAutoCreateTmpl"),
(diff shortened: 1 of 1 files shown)


31453:
@schenksj can you document those new config options at https://molo.ch/settings#advanced-settings (probably in the advanced section at https://github.com/aol/molochweb/blob/master/settings.html) and maybe `release/config.ini.sample`?

schenksj:
My apologies, I just noticed that this got merged!  I will submit the update you requested!

PR #1121, opened 2019-07-08
Add clustering capabilities for Myricom SNF

schenksj:
Current Myricom SNF plugin causes Moloch to fail to start if multiple processes are assigned to a given capture board even when environment variables and flags are set properly.  This adds support for clustered multi-process listeners gaining parity with the features in `pfring` and `tpacketv3`/`AF_PACKET`.

The design of this PR requests rings for all 3 processes (which is required in SNF) within each process, but then binds each process to only its rings.  This is accomplished by multiplying the required number of rings by the number of cluster members, and then binding each cluster member to its own slice of rings.

Further details are in the README changes.

Addresses: https://github.com/aol/moloch/issues/1118

## License

### DIFF
--- a/capture/plugins/snf/README.md
+++ b/capture/plugins/snf/README.md
@@ -1,9 +1,18 @@
 The snf reader plugin now works directly with Myricom SNF
-
+  
 Optional settings
 * ```snfNumRings``` number of rings and threads to use per interface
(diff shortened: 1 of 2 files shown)


awick:
thanks!  Could you also do a PR for aol/molochweb and add to the settings.html

PR #1130, opened 2019-07-24
Include Zeek Logs wise data source

mmguero:
Include Zeek Logs wise data source

Contributed from Malcolm (https://github.com/idaholab/Malcolm).

Changes:
* as zeek creates a LOT of fields, increase max field count from 512 to 1024 (moloch.h, wise.c)
* increase width of field name column in session detail (SessionDetail.vue)
* added commented-out section for enabling zeek log data source in wise.ini.sample
* added source.zeeklogs.js for wise data source creation

Data may be populated with Malcolm's Zeek Logstash filters:
  * https://github.com/idaholab/Malcolm/tree/master/logstash
  * https://github.com/idaholab/Malcolm/blob/master/logstash/pipeline-main/11_zeek_logs.conf

## License


mmguero:
One option for running Logstash configured with the filters i'm using to populate this data would be to just run Malcolm's Logstash docker container standalone:

[Dockerfile](https://github.com/idaholab/Malcolm/blob/master/Dockerfiles/logstash.Dockerfile)
[context](https://github.com/idaholab/Malcolm/tree/master/logstash)

awick:
Looks like these changes cause the wise tests to fail.  Are there other patches missing?

mmguero:
Hmmmm. No, nothing missing that I can see. I'll investigate.

mmguero:
```
internal/buffer.js:555
    throw new ERR_OUT_OF_RANGE('value', `>= ${min} and <= ${max}`, value);
    ^

RangeError [ERR_OUT_OF_RANGE]: The value of "value" is out of range. It must be >= 0 and <= 255. Received 479
    at writeU_Int8 (internal/buffer.js:555:11)
    at Buffer.writeUInt8 (internal/buffer.js:565:10)
    at WISESource.encode (/root/moloch/wiseService/wiseSource.js:309:11)
    at FileSource.WISESource.parseTagger (/root/moloch/wiseService/wiseSource.js:194:68)
    at fs.readFile (/root/moloch/wiseService/source.file.js:76:10)
    at FSReqWrap.readFileAfterClose [as oncomplete] (internal/fs/read_file_context.js:53:3)
```

Looks like the issue is the tests aren't expecting any data sources with more than 255 fields created by default, maybe, when the max should be higher (I set it to 1024 in the code elsewhere). I'll update the pull request with a fix.

mmguero:
Actually the issue isn't with my code, it's with the test framework.

Some time ago you increased the field size to 512 (see commit 328d1d2 and commit d1fd430). This pull request increases it further to 1024. However, that's besides the point: even though capture and wise are both fine with the increased number of fields, the tests.pl which is calling `WISESource.encode` is writing the field count with a `writeUInt8` which won't work. So yes, this pull request does fail the tests, but any other WISE data source that tries to create > 255 fields would have hit the same problem, even though it works fine in production.

iiiusky:
When is this feature expected to be supported? After all, the only shortcoming of zeek is that it cannot export pcap packets, while moloch can.

Issue #1132, opened 2019-08-02
Feature request: honor caTrustFile in capture


MattCarothers:
Attempting to use a TLS-enabled Elasticsearch cluster with internally-signed certificates results in an error:

http.c:237 moloch_http_send_sync(): Retry https://user:password@server.com:9200/_template/sessions2_template?filter_path=**._meta error 'Peer certificate cannot be authenticated with given CA certificates'

The caTrustFile configuration directive works for the viewer but not for capture.

PR #1133, opened 2019-08-02
Update capture to honor the caTrustFile config option.

MattCarothers:
This updates moloch-capture to use the caTrustFile specified in the config file.


MattCarothers:
Addresses #1132

PR #1145, opened 2019-08-22
Cancel request

### LINKED ISSUE #1100: Feature Request: Cancel Pending Query
Add a cancel button in the loading overlay to cancel the currently running query.

Also, cancel the running query if a new query is issued or the page changes.

31453:
Adds the ability to cancel requests made from Sessions, Spigraph, Connections, and Hunt pages.

Cancels the request from the client and the created ES task.

Fixes #1100


31453:
1. I didn't want to pollute the global scope. You're right, filter is used mostly for stuff within the view. I should probably move this somewhere else. I could create a new util file with functions like this so they can be imported as needed. Thoughts?
2. Will do! Good idea. Opaque is not very clear :)

PR #1157, opened 2019-09-08
Single packet session support

bf31415:
enable single packet sessions

for  number of network use cases, performing session aggregation results in loss of important data needed in the analytic space.   this PR allows the user to identify a subset of protocol and/or protocol+port values where default session aggregation is disabled.  Instead each packet which matches one of the protocol and/or protocol+port rules is treated as a single packet session.

1149

no.  no UI changes.

ran ./tests.pl, all passed.

## License

### DIFF
--- a/capture/config.c
+++ b/capture/config.c
@@ -25,6 +25,8 @@
 
 extern MolochConfig_t        config;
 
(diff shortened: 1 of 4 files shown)


awick:
Added most to 2.3.0, thanks!

PR #1161, opened 2019-09-12
Ethernet support

bf31415:
Enable moloch to capture, parse and store ethernet protocol data into elastic search.

Core to my uses cases is the need to support ISIS-- ISIS is an ethernet based IGP commonly used in the service provider space.   This PR enables the acceptance and light touch parsing of a handful of ethernet based protocols including ISIS, LLDP and ARP.   It will also accept other unknown ethernet protocols so these don't go under the radar screen.

Note this PR assumes the presence of the single packet session PR/capability.

1150

Yes, looks like three tests failed-- mpls-basic, gre-erspan and vxlan.   did a quick check of the mpls-basic pcap, lots of other stuff in that pcap besides mpls, including other ethernet protocols (eg loop, cdp, etc.) which i'm guessing the ethernet enablement is now accepting and causing unexpected results for these three test cases.

## License

### DIFF
--- a/capture/config.c
+++ b/capture/config.c
@@ -25,6 +25,8 @@
 
 extern MolochConfig_t        config;
 
+extern MolochSinglePacketSession_t *sps;
+
 LOCAL GKeyFile             *molochKeyFile;
 
 /******************************************************************************/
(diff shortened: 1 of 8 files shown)


awick:
Thanks for the patch but I don't think this is the way we are going to go, definitely want to make it generic and easy for new parsers to be created plus need to add viewer support.  This will include a refactoring of parsers.c/packet.c so you can register for ethernet types. For folks that need something right away they can use this for now.  Also for future patches make sure there are no tabs in the file so a github file comparison looks nice and aligned.  Thanks for the hard work and ideas, keep them coming.

awick:
Added most to 2.3.0, thanks!

PR #1175, opened 2019-09-25
added right-click replacer for %DBFIELD%

mmguero:
When using right-click replacers (see https://molo.ch/settings#right-click) it was possible to use `%FIELD%` to get a field replacement with the internal Moloch field name `ip.src`.

This pull request adds a `%DBFIELD%` replacer which instead is replaced with the elasticsearch field name (`srcIp` instead of `ip.src`).

Very minimal patch. Tested the patch to `master` using latest release branch 2.0.1.

## License


31453:
Thanks! ✨

Issue #1178, opened 2019-09-27
S3 Writer Issues


awick:
From Phillip 

>In viewer/viewer.js at around line 5220, you need to add packetLen into the default list of fields to retrieve. This gets the packetLen into the plugin
However, you then run into another problem in the plugin writer-s3/index.js where it is using (at around line 110) async.eachLimit to fetch the packets from S3. Unfortunately, the first invocation (that finds out the filename) has to complete before the second invocation has a chance of working (the params structure is not yet initialized). (edited) 
The cheap workaround is to change the limit to be 1 — but that serializes all the fetches which is not ideal.
I’m not a node expert so I don’t know what the best approach to fixing that second problem is….

Definitely some issues around having more then one file and order of ops.  Might have this issue for disk stuff, or see how we fixed there.  Could prefetch all the files.

pjsg:
I'm certainly going to look at this this week as well. In particular, I'm also going to look at adding support for compressed files. Yes -- sort of tricky, and will take advantage of the fact that JS supports 53 bit integers (and so can encode more than just the packet offset into the packetPos array). Also, some care needed when writing out the compressed files so that (the required level of) seeking can be supported.

awick:
So looking at the code more, processSessionId is supposed to be preloading the file cache, so probably just need to add a new routine that just returns whats in the filecache instead of using a callback.

As far as encoding more the position more information, moloch used to do that for the file num before it started using the negative numbers.  (btw that means you only have 52 bits since we use 53 to mean its a file).

Probably just move this part of the discussion to #728

PR #1180, opened 2019-09-30
Fix reading files from S3 n the writer-s3 plugin #1178

pjsg:
The issue turned out to be that the lookups to get the filename (in S3) was being done in parallel with the actual fetches of the data from S3. It is not clear how this ever worked. This was reported in #1178 

This version now is careful to get the filenames *before* doing the fetches. Further, it tries to reduce the number of fetches by coalescing fetches for multiple packets into a single fetch. This is based on the fact that the latency to fetch a single byte of data is much larger than fetching a small amount of extra data. Currently set to allowing fetching 32k of useless data. 

This PR also fixes an issue with ordering (I think) of the packets. The old code would order them in the order that the responses came back. The new code maintains the original order.

`npm run lint` passed clean. Also, I ran eslint against the `index.js` (which wasn't covered before) and fixed all the issues so it is now clean.

All tests pass as a result of `make check` (though I did get a capture test failure due to a difference in AS numbers for VeriSign).

## License

### DIFF
--- a/capture/plugins/writer-s3/index.js
+++ b/capture/plugins/writer-s3/index.js
@@ -94,14 +94,18 @@ function processSessionIdS3(session, headerCb, packetCb, endCb, limit) {
   function readyToProcess () {
     var itemPos = 0;
 
-    function process(data, ipos, nextCb) {
-      // console.log("NEXT", data.params);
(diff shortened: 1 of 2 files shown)


awick:
thanks!

Issue #1184, opened 2019-10-03
Add support for storing compressed PCAPs in S3 from writer-s3


pjsg:
We are intending to store PCAPs for long periods of time and want to reduce the S3 cost by only storing compressed PCAPs. The writer-s3 will be changed to write out block-compressed PCAPs that can be read efficiently when extracting uncompressed packets.

I will be posting a PR to provide this capability shortly.

PR #1185, opened 2019-10-03
Add support for s3WriteGzip option to gzip PCAPs on S3.

pjsg:
This PR adds support for two new configuration settings and addresses #1184. There will be another PR in the molochweb repo.

`s3WriteGzip=true`     this forces new files that are created in S3 (as a result of --copy) to be compressed. The files can still be decompressed by `gunzip`, but files compressed by `gzip` cannot be used. This is because there is careful control of the compressed block sizes to allow reading of individual packets without having to decompress the entire file. This defaults to false.

`s3StorageClass=STANDARD`   this sets the storage class used when storing objects in S3. The old default (`REDUCED_REDUNDANCY`) was worse in every way (including cost) than `STANDARD`. A plausible option might be `INTELLIGENT_TIERING`. It defaults to `STANDARD`.

Linting was run on the index.js (though it wasn't easy to do).

All the tests still pass.

## License


awick:
looks like there are some conflicts that need to be resolved.  Also the documentation lives in github.com/aol/molochweb so that will be another pull request.

awick:
this looks very cool, definitely want to look at adding it to normal pcap writing, not sure if there should be any shared code or not.

pjsg:
I think that it probably ought to be imported into the regular copy-to-disk path. However, I think that we want to get some time running this code to make sure that nothing untoward happens before adding it to the majority use case. 

In particular, it won't work without `--copy` in offline mode. My use case is all around offline mode.

Issue #1210, opened 2019-10-23
The s3-writer plugin ought to support EC2 IAM roles.


pjsg:
You currently have to configure the s3-writer with the S3 credentials. This is tiresome as you shouldn't have to configure credentials statically.

[I'll have a PR to fix it]

pjsg:
Merged

PR #1212, opened 2019-10-23
Add support for using AWS EC2 IAM Credentials for accessing S3

pjsg:
This PR allows the S3 writer (and reader) to use EC2 IAM Roles to access S3. This means that the S3 keys do not need to be configured in the `config.ini`. This improves security. #1210 

This PR also enables caching for the `pug` rendered which dramatically improves the performance when opening a session to see the details.

It also fixes a bug (that I introduced) where downloading the PCAP of a session would return packets that didn't match. It turned out that the packet callback (when creating a PCAP to download) actually uses the entire buffer supplied rather than the first packet (like the display does). Anyway, the code now only passes over a single packet buffer.

`npm run lint` returns no errors

`./tests.pl --viewer` -- all tests pass

## License

### DIFF
--- a/capture/plugins/writer-s3.c
+++ b/capture/plugins/writer-s3.c
@@ -287,7 +287,7 @@ void writer_s3_request(char *method, char *path, char *qs, unsigned char *data, 
              (s3Token?";x-amz-security-token":""),
              (specifyStorageClass?";x-amz-storage-class":""),
              bodyHash);
-    //LOG("canonicalRequest: %s", canonicalRequest);
+    LOG("canonicalRequest: %s", canonicalRequest);
 
     g_checksum_reset(checksum);
     g_checksum_update(checksum, (guchar*)canonicalRequest, -1);
@@ -303,7 +303,7 @@ void writer_s3_request(char *method, char *path, char *qs, unsigned char *data, 
              datetime,
              s3Region,
              g_checksum_get_string(checksum));
-    //LOG("stringToSign: %s", stringToSign);
+    LOG("stringToSign: %s", stringToSign);
 
     char kSecret[1000];
     snprintf(kSecret, sizeof(kSecret), "AWS4%s", s3SecretAccessKey);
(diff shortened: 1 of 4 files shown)


viewer/viewer.js
```diff
@@ -5474,6 +5474,7 @@ function localSessionDetailReturnFull(req, res, session, incoming) {
   if (req.packetsOnly) { // only return packets
     res.render('sessionPackets.pug', {
       filename: 'sessionPackets',
+      cache: true,
```
awick: please remove this file from PR

capture/plugins/writer-s3.c
```diff
@@ -607,15 +623,49 @@ void writer_s3_init(char *UNUSED(name))
...
         exit(1);
     }
 
-    if (!s3AccessKeyId) {
-        printf("Must set s3AccessKeyId to save to s3\n");
-        exit(1);
+    if (!s3AccessKeyId || !s3AccessKeyId[0]) {
+        // Fetch the data from the EC2 metadata service
+        size_t rlen;
+        void *metadataServer = moloch_http_create_server("http://169.254.169.254", 10, 10, 0);
```
awick: What is this ip?  Should it be configurable?

pjsg: This is the AWS metadata URL -- all EC2 machines can access this URL to get metadata about the environment.

viewer/viewer.js
pjsg: Did you want the whole file removed or just this line? Do you want this is a separate PR? [I found that it made a big difference in performance]

pjsg:
I'll update the documentation repo when this is merged.....

viewer/viewer.js
awick: In general 1 thing per PR, although letting the db.c slide.  Also in general we like to leave caching off because there are many issues with new installs and not restarting etc that the slight performance hit fixes.  Finally setting NODE_ENV=production many times turns on this type of caching, although I don't know for certain in this case.  So yes a separate PR, probably using NODE_ENV=production if it doesn't use that already would be the right way to go.

PR #1213, opened 2019-10-23
isis refactor

bf31415:
update isis parser to use new field (msgType) as means to express msg type in each isis packet

isis test fails.  expected.  cloud-arp fails which has just arp and icmp.  not sure why.

## License

### DIFF
--- a/capture/parsers/isis.c
+++ b/capture/parsers/isis.c
@@ -24,7 +24,7 @@ LOCAL int isisMProtocol;
 LOCAL int typeField;
 
 /******************************************************************************/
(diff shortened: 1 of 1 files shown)


capture/parsers/isis.c
```diff
@@ -78,4 +119,11 @@ void moloch_parser_init()
                                              isis_pre_process,
                                              isis_process,
                                              NULL);
+
+
+    typeField = moloch_field_define("isis","uptermfield",
```
awick: looks like this should be lotermfield since everything is lowercase

bf31415: k made these changes

PR #1220, opened 2019-10-24
add noCacheJson everywhere and make first middleware

awick:
thoughts?


31453:
There's also a nocache middleware provided by helmet: https://helmetjs.github.io/docs/nocache/. You can use `app.use(helmet.noCache())` instead of putting `noCacheJson` everywhere. Not sure which is better. It looks like the `noCacheJson` sets a few more restrictions than the default `noCache` helmet.

PR #1227, opened 2019-10-25
Slack notifiers should only allow Slack webhook URLs

anshumanbh:
Allowing all values for the Slack webhook URL introduces potential of attacks like SSRF. The only value allowed here should be a valid Slack webhook URL. 

HackerOne report # 721512

## License


awick:
Thanks, but we want to stay compatible with mattermost

PR #1228, opened 2019-10-25
Added permission check for deleting cron queries

inhibitor181:
Do no allow a user to delete another user's cron query

**Relevant issue number(s) if applicable** h1 # 721614

## License


awick:
Thanks! We will verify.

awick:
Thanks! Accepted a PR that fixed for both update and delete

PR #1229, opened 2019-10-25
Update viewer.js

JesseClarkND:
Add permission check to /estask/cancelById

Users can create/cancel/etc tasks when they have the createEnabled property enabled on their user account.
Permissions checks are done on POST receivers in the view.js.
There is a missing check for createEnabled at the /estask/cancelById controller.

**No relevant issue number**

## License


31453:
Implemented using middleware https://github.com/aol/moloch/pull/1273

PR #1232, opened 2019-10-25
Fix POST body parsing DoS

MrTuxracer:
wiseService parses the POST body and uses the first byte in call to buf.readUInt16BE eventually reading outside of the buffer leading to a crash of node.

Relevant issue number(s): HackerOne/#722898

## License


31453:
Fixed in https://github.com/aol/moloch/pull/1265

PR #1235, opened 2019-10-25
Maintain a list of active ES cluster

codesniffer:
**Problem Statement**: The MultiViewer cannot display session data if a node provided in the multiESNodes setting is down. The MultiViewer should work if at least one of the nodes from multiESNodes setting is active/up.  

****Solution**:  Maintain a list of active ES nodes by periodically submitting a ping request (HEAD /) to every multiESNodes. The MultiViwer displays data from the active nodes by excluding inactive nodes from the query.

** `npm run lint`**: Shows no error

## License

### DIFF
--- a/viewer/multies.js
+++ b/viewer/multies.js
@@ -795,50 +795,39 @@ nodes.forEach((node) => {
 activeESNodes = nodes.slice();
 
 // Ping (HEAD /) periodically to maintian a list of active ES nodes
-function pingESNodes(client, node, cb) {
-  client.ping({
-    requestTimeout: 3*1000 //ping usually has a 3000ms timeout
-  }, function (error) {
-    if (error) {
-      cb({value: false, node: node}); // ES cluster is down
-    } else {
-      cb({value: true, node: node}); // ES cluster is up
-    }
-  });
-}
-
-function getPingPromise(client, node){
+function pingESNode(client, node){
   return new Promise((resolve, reject) => {
-      pingESNodes(client, node, (result) => {
-        if(result.value === true) { // ES cluster is up
-          resolve(result.node);
-        } else { // ES cluster is down
-          // remove credential from the url
-          var host = result.node.split("://");
-          host = host[host.length > 1 ? 1 : 0 ].split("@"); // user:pass@elastic.com:9200
-          host = host [host.length > 1 ? host.length-1 : 0];
-          console.log("Elasticsearch is down at ", host);
-          resolve(null);
-        }
+      client.ping({
+        requestTimeout: 3*1000 //ping usually has a 3000ms timeout
+      }, function (error) {
+        resolve({isActive: error ? false : true, node: node});
       });
     });
 }
 
-function enumarateActiveNodes(){
+function enumerateActiveNodes(){
   var ping_tasks = [];
   for (var i = 0; i < nodes.length; i++) {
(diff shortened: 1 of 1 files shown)


viewer/multies.js
```diff
@@ -785,6 +792,54 @@ nodes.forEach((node) => {
...
+          host = host[host.length > 1 ? 1 : 0 ].split("@"); // user:pass@elastic.com:9200
+          host = host [host.length > 1 ? host.length-1 : 0];
+          console.log("Elasticsearch is down at ", host);
+          resolve(null);
+        }
+      });
+    });
+}
+
+function enumarateActiveNodes(){
```
awick: enumerate

viewer/multies.js
```diff
@@ -785,6 +792,54 @@ nodes.forEach((node) => {
...
+    var active_nodes = [];
+    for (var i = 0; i < values.length; i++) {
+      if(values[i]) { active_nodes.push(values[i]); }
+    }
+    activeESNodes = active_nodes.slice();
+  });
+}
+
+setInterval(enumarateActiveNodes, 1*60*1000); // 1*60*1000 ms
+
```
awick: should initialize activeESNodes = nodes so it has something the first minute.

viewer/multies.js
```diff
@@ -117,7 +117,14 @@ function node2Prefix(node) {
   return "";
 }
 
+var activeESNodes = [];
+
+function getActiveNodes(){
+  return activeESNodes.slice();
```
awick: why slice?

viewer/multies.js
```diff
@@ -785,6 +792,54 @@ nodes.forEach((node) => {
...
 });
 
+activeESNodes = nodes.slice();
+
+// Ping (HEAD /) periodically to maintian a list of active ES nodes
+function pingESNodes(client, node, cb) {
+  client.ping({
+    requestTimeout: 3*1000 //ping usually has a 3000ms timeout
+  }, function (error) {
+    if (error) {
```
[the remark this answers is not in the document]
codesniffer: Refactored the code according to the suggestion

viewer/multies.js
codesniffer: Made correction

viewer/multies.js
codesniffer: This initialization has already been implemnted

viewer/multies.js
codesniffer: **activeESNodes.slice()** returns the elements (values) of an array instead of the reference to the array. The statement **activeESNodes.slice()** will make the list activeESNodes Thread-safe. If the statment **return activeESNodes** is used then it is possible that the list activeESNodes is getting updated by the **enumerateActiveNodes** method while the list is being used by the **SimpleGather** or **Msearch** method.

viewer/multies.js
```diff
@@ -785,6 +792,43 @@ nodes.forEach((node) => {
...
 
+activeESNodes = nodes.slice();
+
+// Ping (HEAD /) periodically to maintian a list of active ES nodes
+function pingESNode(client, node){
+  return new Promise((resolve, reject) => {
+      client.ping({
+        requestTimeout: 3*1000 //ping usually has a 3000ms timeout
+      }, function (error) {
+        resolve({isActive: error ? false : true, node: node});
```
awick: Why not just do !error?

PR #1240, opened 2019-10-26
Fix XSS in connections error handling

securitybites:
For H1 report #721517


awick:
Thanks! Already fixed in #1220

PR #1243, opened 2019-10-28
Fix improper null termination (CWE-170)

Cache-Money:
H1 #723849

## License


awick:
Good find, went with a different fix, btw `groupbuf[buflen] = 0` would overflow

PR #1247, opened 2019-10-28
Fix oracle padding

c0rvax:
https://hackerone.com/reports/723098

## License

### DIFF
--- a/viewer/config.js
+++ b/viewer/config.js
@@ -103,18 +103,37 @@ exports.store2ha1 = function(passstore) {
 
 exports.obj2auth = function(obj, secret) {
   secret = secret || exports.getFull("default", "serverSecret") || exports.getFull("default", "passwordSecret", "password");
+
   var c = crypto.createCipher('aes192', secret);
   var e = c.update(JSON.stringify(obj), "binary", "hex");
   e += c.final("hex");
-  return e;
+
+  var h = crypto.createHmac('sha256', secret);
+  h.update(e, 'hex');
+  var countedSignature = h.digest('hex');
+
+  return e + '.' + countedSignature;
 };
 
 exports.auth2obj = function(auth, secret) {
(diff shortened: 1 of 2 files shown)


awick:
Looks good, however as currently written if you have many servers it would require them all to upgrade at once since auth2obj is used for server2server also.  Thoughts on adding a parameter to enforce the signature for the x-moloch-cookie check but not for the x-moloch-auth (since its server2server and usually over https)

So would have something like `Config.auth2obj(req.headers['x-moloch-auth'], true);` and change the function to be e`xports.obj2auth = function(obj, sigCheckOptional, secret) {`

c0rvax:
Agreed, that is reasonable, you want me to update the fix or you will handle that?

awick:
go for it if you have time

c0rvax:
Actually if we leave `x-moloch-auth` header without that signature, oracle padding would be still in place. It would be exploited against that method which parses and checks `x-moloch-auth`.

I will take some time and think about a more portable fix

c0rvax:
Oh, may be we just hide the errors. Basically there are two types of errors: JSON decode error and bad decryption. The former means the padding is ok, the latter means padding is not OK. Seing those errors plaintext allows finding the padding.

If we just catch both errors and return a common one, we get a fix.

awick:
Agree, but I was thinking s2s is potentially more risk acceptable unlike c2s, at least while folks roll out upgrades, and could be removed in a future version.

Hiding the errors for both and doing the full fix for c2s seems the way to go?  And later can do the full fix for s2s also.

c0rvax:
Agreed, let's do it that way. Preparing the commits

c0rvax:
I have fixed that. Looks like all tests are clean

awick:
* Looks like a second change got added mixed in
* I think it should always sign, its only on the checking side that matters right?  So if there is a . then check, if there is not a dot then if signIsOptional thats ok, otherwise its an error.

PR #1255, opened 2019-10-28
Fix buffer overflow (CWE-120)

Cache-Money:
H1 #724239

## License


awick:
Good find, went with a different fix, btw `lastGroup[buflen] = 0` would overflow

PR #1256, opened 2019-10-29
Fix for 724469

c0rvax:
Fix for https://hackerone.com/reports/724469

## License


31453:
Parliament is an exclusively internal tool that is run behind a reverse proxy that provides authentication. The Parliament admin should be trusted.

PR #1266, opened 2019-10-31
[DISTURBANCE] Fix Report 725223

spaceraccoon:
Fix Report 725223.

Tested on local instance and follows the `check*` middleware pattern established in https://github.com/spaceraccoon/moloch/blob/master/viewer/viewer.js#L854 and the `verify*` naming pattern for Parliament. Resolves issue for both endpoints.

## License

### DIFF
--- a/parliament/parliament.js
+++ b/parliament/parliament.js
@@ -323,6 +323,19 @@ function verifyToken (req, res, next) {
   });
 }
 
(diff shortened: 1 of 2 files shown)


31453:
Parliament is an exclusively internal tool that is run behind a reverse proxy that provides authentication. The Parliament admin is trusted.

PR #1273, opened 2019-11-01
Permission middleware

31453:
check endpoint permissions via middleware function


awick:
1) webEnabled needs to be added to I think the following
* /user.css
* /user/current
* /user/settings
* /user/columns
* /user/spiview/fields

2) i think we should add comments before the app.get that no perm are required to
/parliament.json
/eshealth.json

Issue #1272, opened 2019-11-01
Enable caching in pug if running in a production environment


pjsg:
**Moloch version**: master

**OS name and version**: ubuntu 16.04

If you start moloch with `NODE_ENV = "production"` then `express` runs in production mode. Unfortunately, `pug` does not. The documentation for express & pug seems to indicate that a production express environment ought to enable pug caching -- but this doesn't work, and the code to do this does not appear to be present in either express or pug.

I'll stick up a trivial PR to resolve this for moloch.

PR #1274, opened 2019-11-01
When running a production express environment, enable pug caching

pjsg:

The solution is to check when the app is running in production mode and, if so, enable pug caching (and turn off the debug compile). This (for me) makes a significant difference in the performance of the viewer when looking at the details of a session.

#1272 

yes

yes

## License

### DIFF
--- a/viewer/viewer.js
+++ b/viewer/viewer.js
@@ -114,6 +114,10 @@ if (internals.elasticBase[0].lastIndexOf('http', 0) !== 0) {
   internals.elasticBase[0] = "http://" + internals.elasticBase[0];
 }
 
+function isProduction() {
+  return app.get('env') === 'production';
+}
+
 function userCleanup(suser) {
   suser.settings = suser.settings || {};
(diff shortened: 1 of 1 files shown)


viewer/viewer.js
```diff
@@ -114,6 +114,10 @@ if (internals.elasticBase[0].lastIndexOf('http', 0) !== 0) {
   internals.elasticBase[0] = "http://" + internals.elasticBase[0];
 }
 
+function isProduction() {
+  return app.get('env') === 'production';
```
awick: shouldn't this be NODE_ENV instead of env?

awick: oh does express change it from NODE_ENV to just env?

pjsg: According to: https://expressjs.com/en/guide/migrating-4.html

```The app.configure() function has been removed. Use the process.env.NODE_ENV or app.get('env') function to detect the environment and configure the app accordingly.```

PR #1276, opened 2019-11-02
Prevent Many-Time Pad Attack

cablej:
Reference: https://hackerone.com/reports/727284

## License


awick:
Thanks! For now we are going to leave xor support, the code already warns, and will update the doc more that it isn't secure.  We want to have multiple encodings for testing purposes.

PR #1277, opened 2019-11-03
[h1-213] Fix for #727952	

tabahi90:
added permission check for "update cron query"


awick:
Thanks! Accepted a PR that fixed for both update and delete

PR #1278, opened 2019-11-03
[h1-213] Fix for #724985

tabahi90:
added permission check !


awick:
Not sure what this fix is doing?

PR #1281, opened 2019-11-04
fix for #729267

tabahi90:
added permission check!


awick:
Not sure the point of this fix, but any change should be in removeTagsList

PR #1284, opened 2019-11-07
Fix null-pointer dereference

ajxchapman:
Fixes HackerOne reports 726769

## License

### DIFF
--- a/capture/parsers/ssh.c
+++ b/capture/parsers/ssh.c
@@ -47,6 +47,7 @@ LOCAL void ssh_parse_keyinit(MolochSession_t *session, const unsigned char *data
 
     BSB_IMPORT_u32(bsb, len);
     BSB_IMPORT_ptr(bsb, value, len);
+    if (BSB_IS_ERROR(bsb)) return;
     BSB_EXPORT_ptr(hbsb, value, len);
     BSB_EXPORT_u08(hbsb, ';');
 


awick:
Yep this would work, but decided to add a check in EXPORT_ptr in https://github.com/aol/moloch/commit/41d8630b566b56833622eec700f9784c564973e3 to catch any of these.  Let me know if you think it won't work

ajxchapman:
@awick, whilst that fixes the null pointer dereference,  I believe the initial proposed fix is more robust for two reasons. 

1. Not checking the return value from `BSB_IMPORT_ptr` is akin to not checking the return value from `malloc`. In future if /capture/parsers/ssh.c decides to parse the available SSH keys this issue could arise again, e.g.:
```c
int key_exchange_len;
BSB_IMPORT_ptr(bsb, value, len);
key_exchange_len = data[3] << 24 | data[2] << 16 | data[1] << 8 | data[0];
```
2. The suggested fix is inline with the rest of the source file, e.g.
https://github.com/aol/moloch/blob/8c7feafe7a71da4b559a45a72a21cdb7145fc647/capture/parsers/ssh.c#L58-L60
https://github.com/aol/moloch/blob/8c7feafe7a71da4b559a45a72a21cdb7145fc647/capture/parsers/ssh.c#L67-L69
https://github.com/aol/moloch/blob/8c7feafe7a71da4b559a45a72a21cdb7145fc647/capture/parsers/ssh.c#L76-L78
etc.

PR #1287, opened 2019-11-08
added checkCookieToken middleware to removeTags endpoint

yashrsm:
https://hackerone.com/reports/732333 for all info


31453:
Went this route instead: https://github.com/aol/moloch/pull/1305

PR #1289, opened 2019-11-08
add setcookie middleware

31453:
apply it to main get requests for each page
apply check cookie middleware to state endpoint


31453:
Where else should I check the cookie? It's applied to most post and delete endpoints already

PR #1290, opened 2019-11-08
Fetch S3 dependencies over HTTPS

cablej:
Reference: report #724250

## License


awick:
Thanks! We went ahead and fixed all of them in these files instead of just files.molo.ch

PR #1295, opened 2019-11-08
[MALA FAMA TEAM] Fix auth issue for HackerOne report 723429

mala-fama:
## License


viewer/viewer.js
```diff
@@ -6384,6 +6384,10 @@ app.put('/user/:userId/acknowledgeMsg', [noCacheJson, logAction(), checkCookieTo
     return res.molochError(403, 'Message number required');
   }
 
+  if (req.params.userId != req.user.userId) {
```
31453: Can you use `!==` here instead of `!=`

mala-fama: Our bad, fixed! Thanks

PR #1297, opened 2019-11-09
[MALA FAMA TEAM] Improve coverage of the fuzzer 

mala-fama:
Hi!

This one is related to HackerOne reports 725149 and 725151.

The reason behind this PR is that the fuzzer is only fuzzing the ipv6/ipv4 packets (see 
 https://github.com/aol/moloch/blob/master/capture/packet.c#L1916). If we modify the
linkType to be 1 we are going to get to the ether packet level, meaning that we will fuzz *everything*. Depending on the type, will select the type of packet that we want and trigger so we will end up with much more coverage of the program. On our tests we got around 15% more coverage and found some new crashes doing so.

## License


awick:
Thanks! Great idea!  I placed in another location, but helpful.

PR #1298, opened 2019-11-09
Patch for bug report #732585 - team ☐

smiegles:
## License

### DIFF
--- a/capture/main.c
+++ b/capture/main.c
@@ -43,6 +43,7 @@ unsigned char          moloch_hex_to_char[256][256];
 
 extern MolochWriterQueueLength moloch_writer_queue_length;
 extern MolochPcapFileHdr_t     pcapFileHeader;
+uint32_t                hashSalt;
 
 MOLOCH_LOCK_DEFINE(LOG);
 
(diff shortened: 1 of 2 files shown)


capture/main.c
```diff
@@ -43,6 +43,7 @@ unsigned char          moloch_hex_to_char[256][256];
 
 extern MolochWriterQueueLength moloch_writer_queue_length;
 extern MolochPcapFileHdr_t     pcapFileHeader;
+uint32_t                hashSalt;
```
awick: can you move all of this to session.c and make it a LOCAL var just for session.c

capture/session.c
```diff
@@ -26,6 +26,7 @@ extern time_t                lastPacketSecs[MOLOCH_MAX_PACKET_THREADS];
 /******************************************************************************/
 
 LOCAL int                   protocolField;
+extern uint32_t              hashSalt;
```
awick: make sure the h lines up with the above p, and since you'll be moving the hashSalt init to session.c change extern to LOCAL

PR #1304, opened 2019-11-12
Arch linux installer support

rnbwdsh:
Adds a MacOS like installer (without thirdparty stuff) configured to use system libs for everything.

Also add docker installer + an intermediate commit to test if install works and to determine requirements for proper pacman package.

Check out the [previous commit](https://github.com/aol/moloch/commit/2b326a3cc0af93b6ce3abaf5f8ef65e9df68e3d0) to run tests locally.


awick:
thanks, took pieces and now we have a arch build

PR #1306, opened 2019-11-13
mutex and tests fixes

awick:
* setLastUsed is now an update instead of setting whole user
* Add a mutex for lookups so you can't add same name twice
* Rename getSettingUser/postSettingUser to getSettingUserCache/getSettingUserDb
* getSettingUserDb now handles the error for caller if doesn't exist
* Fix views to not edit cached user obj
* improve users tests and added views comments

## License

### DIFF
--- a/viewer/viewer.js
+++ b/viewer/viewer.js
@@ -659,30 +659,30 @@ function arrayZeroFill(n) {
 
 // https://stackoverflow.com/a/48569020
 class Mutex {
-    constructor () {
-        this.queue = [];
-        this.locked = false;
-    }
+  constructor () {
+    this.queue = [];
+    this.locked = false;
+  }
 
-    lock () {
-        return new Promise((resolve, reject) => {
-            if (this.locked) {
-                this.queue.push([resolve, reject]);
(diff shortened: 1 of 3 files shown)


viewer/viewer.js
```diff
@@ -657,6 +657,34 @@ function arrayZeroFill(n) {
...
+            } else {
+                this.locked = true;
+                resolve();
+            }
+        });
+    }
+
+    unlock () {
+        if (this.queue.length > 0) {
+            const [resolve, reject] = this.queue.shift();
```
31453: Is reject necessary here? This will fail linting cause it is an unused variable.

viewer/viewer.js
```diff
@@ -657,6 +657,34 @@ function arrayZeroFill(n) {
   return a;
 }
 
+// https://stackoverflow.com/a/48569020
+class Mutex {
+    constructor () {
+        this.queue = [];
```
31453: Just a nitpick, but in the rest of the code, we use 2 spaces, not 4.

Issue #1312, opened 2019-11-19
When using Wise and the Tagger file, cannot import UTF-8 strings


pjsg:
**Elasticsearch version**:   git master

**Moloch version**: 2.1.0

**OS name and version**: ubuntu 18.04

**How was Moloch built/installed:** (rpm, deb, easybutton, ...) easybutton

If the tagger file includes (say):

#field:bs.info;kind:termfield;friendly:bsinfo;db:bs.info;help:The IP Information;shortcut:i
1.2.3.4:i=女宇農系人女族本職味協動公。失浮枝真公室霊奪疑奮発販。

and then Wise tags 1.2.3.4 -- the string is not shown correctly. This is becuase the UTF8 flag is not set on the field.

awick:
do you think it should be on by default and a new flag to turn off instead?

pjsg:
I'm always nervous about making backwards-incompatible changes. I can't *think* of any use for the 8-bit mode -- it isn't 8-bit clean in that if you put in (say) 0xa8, you don't get that back.

PR #1313, opened 2019-11-19
Add the utf8:true field option to support UTF8 strings

pjsg:

The problem is that wise did not set the UTF8 flag on the dynamically created field and hence the import process treated the text as being in 8-bit (not entirely clear which character set).

This just adds support for being able to define a field as having utf-8 values.

#1312 

n/a

n/a

I'll submit another PR for the documentation repo.

## License


awick:
went with the reverse, cross your fingers :)

PR #1320, opened 2019-11-24
Added SIGTERM handler for graceful termination

fj604:
Added SIGTERM handler for graceful service shutdown, equivalent to SIGINT

SIGTERM handler ensures that all queues are drained before the service is shut down, reducing chances of data loss caused by shutdown or restart.

N/A

N/A - no changes to viewer

N/A - no changes to viewer

## License

### DIFF
--- a/capture/main.c
+++ b/capture/main.c
@@ -310,6 +310,12 @@ void controlc(int UNUSED(sig))
     moloch_quit();
 }
 /******************************************************************************/
+void terminate(int UNUSED(sig))
+{
+    LOG("Terminate");
+    moloch_quit();
+}
+/******************************************************************************/
 void reload(int UNUSED(sig))
 {
     moloch_plugins_reload();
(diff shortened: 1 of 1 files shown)


awick:
Why is TERM need since its the same as INT?

fj604:
> Why is TERM need since its the same as INT?

When the server (physical or virtual) is shut down or rebooted, molochcapture.service receives SIGTERM and not SIGINT.

awick:
ok, please update to call exit on second call like controlc does

fj604:
> ok, please update to call exit on second call like controlc does

When the service is stopped (manually or upon shutdown or reboot), for some reason SIGTERM is sent twice. If I add the same update as in controlc then the second SIGTERM would cause immediate exit which is what I'm trying to prevent.

awick:
i think you need to figure out why that happens

fj604:
> i think you need to figure out why that happens

This has something to do with moloch-capture stopped by systemd.  If I kill the service with the `kill` command, I see only one `SIGTERM` sent, but `systemctl stop molochcapture.service` sends `SIGTERM` twice.

fj604:
Yes, definitely something to do with systemd.
If moloch-capture is started with `service molochcapture start` then either ` systemctl stop molochcapture.service` or `service molochcapture stop` sends one SIGTERM.
If moloch-capture is started with `systemctl atart molochcapture.service` then either `service molochcapture stop` or `systemctl stop molochcapture.service` sends SIGTERM twice.

PR #1322, opened 2019-11-27
S3 access changed to virtual hosted style 

fj604:
Changed S3 addressing from path-style to virtual-hosted style

Amazon has deprecated path-style access (`https://s3.amazonaws.com/bucket/key`) to S3 buckets. Buckets created after September 30, 2020, will support only virtual hosted-style requests (`https://bucket.s3.amazonaws.com/key`).
References:
[AWS Blog](https://aws.amazon.com/blogs/aws/amazon-s3-path-deprecation-plan-the-rest-of-the-story/)
[S3 Documentation](https://docs.aws.amazon.com/AmazonS3/latest/dev/UsingBucket.html)

Changing S3 access to virtual hosted-style to comply with the deprecation.
For bucket names with dots, added a fallback option:
`s3PathAccessStyle=true`
It will default to true only if the bucket name contains a dot, otherwise will be false.

N/A

N/A

N/A

## License


pjsg:
Does this work for bucket names with dots in them? I seem to recall that there are TLS certificate issues....

fj604:
> Does this work for bucket names with dots in them? I seem to recall that there are TLS certificate issues....

Yes indeed there is an issue with bucket names containing dots - upload fails due to hostname mismatch in SSL wildcard certificate:

For example, for bucket name `fj,moloch`:

```
Dec  5 18:48:19 http.c:584 moloch_http_curl_watch_open_callback(): Connected 1/0 - https://fj.moloch.s3-eu-west-2.amazonaws.com   35142->52.95.148.46:443 - fd:8
* Connected to fj.moloch.s3-eu-west-2.amazonaws.com (52.95.148.46) port 443 (#0)
* ALPN, offering http/1.1
* Cipher selection: ALL:!EXPORT:!EXPORT40:!EXPORT56:!aNULL:!LOW:!RC4:@STRENGTH
* successfully set certificate verify locations:
*   CAfile: /etc/pki/tls/certs/ca-bundle.crt
  CApath: none
* SSL connection using TLSv1.2 / ECDHE-RSA-AES128-GCM-SHA256
* ALPN, server did not agree to a protocol
* Server certificate:
*  subject: C=US; ST=Washington; L=Seattle; O=Amazon.com, Inc.; CN=*.s3.eu-west-2.amazonaws.com
*  start date: Nov  9 00:00:00 2019 GMT
*  expire date: May 20 12:00:00 2020 GMT
*  subjectAltName does not match fj.moloch.s3-eu-west-2.amazonaws.com
* SSL: no alternative certificate subject name matches target host name 'fj.moloch.s3-eu-west-2.amazonaws.com'
* Closing connection 0

```

This is [documented ](https://docs.aws.amazon.com/AmazonS3/latest/dev/VirtualHosting.html)by AWS:

> When using virtual hosted–style buckets with SSL, the SSL wild-card certificate only matches buckets that do not contain periods. To work around this, use HTTP or write your own certificate verification logic.

What should be done about it?

- Ignore deprecation and leave the code as is with path-style access? Or...
- Add a note about bucket names with dots not supported? Or...
- Disable certificate verification in libcurl call? Or...
- As per Amazon's suggestion, write custom certificate verification logic?

awick:
no custom cert logic please. :)  I would suggest we leave both methods in and just either add a switch for which method to use, or if there is a dot in the name choose the old method otherwise use the new method.

fj604:
> I would suggest we leave both methods in and just either add a switch for which method to use, or if there is a dot in the name choose the old method otherwise use the new method.

This means that uploads to buckets with dots created after September 30, 2020 would not work.

awick:
correct.  isn't it kind of a well known issue now to not use dots unless its for static host serving?

fj604:
I believe it is well-known. I'll see if I can add a new switch to the code. Still new to GitHub - this is only my second OSS contribution. Thank you for your help!

capture/plugins/writer-s3.c
```diff
@@ -261,9 +262,38 @@ void writer_s3_request(char *method, char *path, char *qs, unsigned char *data,
     g_checksum_reset(checksum);
     g_checksum_update(checksum, data, len);
     strcpy(bodyHash, g_checksum_get_string(checksum));
-    snprintf(canonicalRequest, sizeof(canonicalRequest),
+    if (s3PathAccessStyle) {
```
awick: Minor nit.  Instead of this big if, could you have just a fullpath var like below and use that in this url?

fj604: > Minor nit. Instead of this big if, could you have just a fullpath var like below and use that in this url?

I couldn't reuse the fullpath var as it also had the querystring appended to it, so I added a new var. Not ideal but works and looks nicer :)

capture/plugins/writer-s3.c
```diff
@@ -616,6 +649,7 @@ void writer_s3_init(char *UNUSED(name))
 
     s3Region              = moloch_config_str(NULL, "s3Region", "us-east-1");
     s3Bucket              = moloch_config_str(NULL, "s3Bucket", NULL);
+    s3PathAccessStyle     = moloch_config_boolean(NULL, "s3PathAccessStyle", FALSE);
```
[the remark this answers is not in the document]
fj604: Done and thanks!

awick:
seems ok to me, @pjsg look ok?

Issue #1303, opened 2019-11-12
Usage of non-exposed pcap function


rnbwdsh:
**Elasticsearch version**: irrelevant

**Moloch version**: 2.1

**OS name and version**: 5.3.8-1-MANJARO

**How was Moloch built/installed:** (rpm, deb, easybutton, ...): modified install for arch based systems: `./configure --with-libpcap=no --with-yara=no --with-glib2=no --with-pfring=no --with-curl=no --with-lua=no LIBS="-lpcap -lyara -llua -lcurl" GLIB2_CFLAGS="-I/usr/include/glib-2.0 -I/usr/lib/glib-2.0/include" GLIB2_LIBS="-lglib-2.0 -lgmodule-2.0 -lgobject-2.0 -lgio-2.0`

**Provide logs, stack traces and steps to reproduce**: (inside easybutton --install, while building capture)
``` 
gcc -rdynamic -ggdb -pthread main.o db.o yara.o http.o config.o parsers.o plugins.o field.o trie.o writers.o writer-inplace.o writer-disk.o writer-null.o writer-simple.o readers.o reader-libpcap-file.o reader-libpcap.o reader-tpacketv3.o reader-null.o packet.o session.o rules.o drophash.o pq.o -o moloch-capture \
            -u g_checksum_update -u g_hmac_update -u g_uri_unescape_segment \
    -lnl-genl-3 -lnl-3 \
    -lglib-2.0 -lgmodule-2.0 -lgobject-2.0 -lgio-2.0 -lmaxminddb -lnet -lrt -luuid -lpcre -lpcap -lyara -llua -lcurl thirdparty/http_parser.o thirdparty/js0n.o thirdparty/patricia.o -ldl -lssl -lcrypto -lyaml \
    -lm -lresolv -lmagic -lffi -lz
/usr/bin/ld: reader-libpcap-file.o: in function `reader_libpcapfile_opened':
/home/m/IdeaProjects/moloch/capture/reader-libpcap-file.c:562: undefined reference to `dlt_to_linktype'
/usr/bin/ld: reader-libpcap.o: in function `reader_libpcap_start':
/home/m/IdeaProjects/moloch/capture/reader-libpcap.c:97: undefined reference to `dlt_to_linktype'
collect2: error: ld returned 1 exit status
make[1]: *** [Makefile:41: all] Error 1
make[1]: Leaving directory '/home/m/IdeaProjects/moloch/capture'
make: *** [Makefile:329: all-recursive] Error 1
[...]

PR #1326, opened 2019-12-01
Fix pcap writing and link-layer DLT/LINKTYPE handling from issue #1303

rnbwdsh:
Rename pcapFileHeader.linktype to pcapFileHeader.dlt
Use DLT_EN10MB instead of 1 when using moloch_packet_set_linksnap
Use sizeof(MolochPcapFileHdr_t) instead of 24
Fix moloch package headers on write via copy + modify if needed + write copy
Add DLT_* and LINKTYPE_* defines, as pcap doesn't export them
Remove references to dlt_to_linktype as pcap doesn't export them (this broke .so compatibility with libpcap)
Fix broken calls to moloch_packet_set_linksnap( dlt_to_linktype | ..)

### DIFF
--- a/capture/main.c
+++ b/capture/main.c
@@ -644,7 +644,7 @@ gboolean moloch_ready_gfunc (gpointer UNUSED(user_data))
         }
     }
     moloch_reader_start();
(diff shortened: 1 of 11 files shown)


awick:
We merged in our own fix, but thank you for all the work.

Issue #1318, opened 2019-11-20
Improve HTTP callback integration into the lua plugin


pjsg:
I'd like to see more of the http parsing hooks exposed to the lua plugin. I think that all of the hooks should be exposed into the lua code.

For example:

MolochSession.on_http_cb(cb_type, function)

cb_type is one of "header_value", "header_field", "headers_complete", "message_begin", "url", "body", "message_complete".

The callback function would be invoked with the session object, the direction (0/1), data (if relevant).

Note that this will duplicate  (in one case) the existing register_body_feed. 

I'd also like to add a "header_field_raw" that returns the header field string before it is lower-cased. 

I'd appreciate comments on this before starting work.

Does anybody use the lua integration?

awick:
I'm not sure if anyone uses it.  We did do that sample creditcard.lua

31453:
Implemented in c43c659

PR #1336, opened 2019-12-10
Improve the LUA plugin for HTTP

pjsg:

This fixes the issue #1318. In particular, it allows access (from Lua) to all the phases of the http parsing -- including headers. It also fixes the nasty problem that could happen if you used MolochData objects incorrectly.

#1318 

This was tested with a simple lua app that looked at all the header field names and generated a hash code for the fields -- which allowed similar browsers / clients to be grouped together. 

## License


pjsg:
Note that no companion PR is needed in molochweb as the documentation for the LUA plugin is in this repo (and it was updated).

awick:
are you willing to share your sample app?

pjsg:
Yes -- I will share it. I'm actually going to track down a weird issue that I ran into, and fix that as well. [Adding a field to a session doesn't seem to work if the field name has a dot in it].

Issue #1350, opened 2020-01-02
moloch_update_geo.sh no longer works.


pjsg:
**Elasticsearch version**: N/A

**Moloch version**: master

**OS name and version**: ubuntu 16.04

**How was Moloch built/installed:** (rpm, deb, easybutton, ...) easybutton

**Provide logs, stack traces and steps to reproduce**:

The `moloch_update_geo.sh` no longer works as maxmind now requires an account id.

awick:
yep, still trying to decide what we are going to do.  such a pain.

gmctl:
do you have the accounts of maxmind . i also catch the 401 of this problem with not build the geoip-datebase. and i have made the old geoip2 with it.

awick:
Fixed in master and will be part of 2.2.  You'll have to sign up for an account :(

https://molo.ch/faq#maxmind

Trolldemorted:
Can we tell moloch to operate without any geo data?

PR #1357, opened 2020-01-09
Set GeoIP default database paths to empty

fj604:
Change the default GeoIP file location to empty, unless database files exist in /data/moloch/etc

As GeoIP now requires a license key, change the default setting not to use GeoIP unless database files exist.

#1350 is still broken but it will no longer cause the capture service to stop as it will no longer look for GeoIP databases

n/a

yes

## License


awick:
Thanks, went with a different solution.  

FYI, you can call moloch_config_str(keyfile... with NULL and it will do the right thing.

PR #1360, opened 2020-01-20
Capture tcp health check

fj604:
Capture plugin to respond to TCP health checks made by AWS Network Load Balancer.

Provide a simple TCP socket listener in the capture module so that AWS Network Load Balancer can confirm that the service is up and running. A new setting tcpHealthCheckPort is added to config.ini. The plugin starts a TCP server in a single thread that listens for connections on a socket and immediately closes them.

n/a

n/a

Yes

## License

### DIFF
--- a/capture/plugins/tcphealthcheck.c
+++ b/capture/plugins/tcphealthcheck.c
@@ -1,0 +1,95 @@
+/* tcphealthcheck.c  -- TCP Health Check plugin
+Listens on a TCP socket and immediately closes it to confirm that the capture service is running.
+ */
+
+
+#include <sys/socket.h>
+#include <netinet/in.h>
+#include <arpa/inet.h>
+#include <stdio.h>
+#include <string.h>
+#include <stdlib.h>
+#include <ctype.h>
+#include <pthread.h>
+#include "moloch.h"
+
+
+/******************************************************************************/
+
+extern MolochConfig_t        config;
+
+LOCAL  int                   tcp_port;
+
+/******************************************************************************/
+void server(void) {
+    int server_fd, client_fd, err;
+    struct sockaddr_in server, client;
+   
+    server.sin_family = AF_INET;
+    server.sin_port = htons(tcp_port);
+    server.sin_addr.s_addr = htonl(INADDR_ANY);
+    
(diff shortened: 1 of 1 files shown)


awick:
Thanks for the PR.  I'm not sure I understand the point.  Are you just using this for alerting?  Could parliament be changed to do what you need instead?  My main issue is, just because its listening doesn't mean packets are being processed and just because its down doesn't mean its not restarting.

fj604:
As you can see this single thread just plays whack-a-mole with any open sockets on a port which is enough to keep AWS Load Balancer TCP health check happy. The idea is that if the service is not running then the socket will not be listening and AWS would terminate the instance and replace it with a new one - auto scaling takes care of this.

With dedicated capture and viewer nodes, capture nodes will be under their own load balancer and auto scaler, separate from viewer nodes. There is no TCP listener on a capture node apart from SSH, and using the OpenSSH daemon as a health check listener does not tell the load balancer whether the capture process is running or not. The load balancer will not react to a single missed health check - it would take at least two health checks failing at a 30-second interval by default to mark the instance as unhealthy, so the capture process restart shouldn't cause this. Calling it healthy only when packets are processed might not be best as the service could just be idling for some time due to no traffic hitting it.

fj604:
To answer your question, this is not used for alerting but to signal to the AWS load balancer to take the instance out of service (stop routing traffic to it) and to the auto scaler to terminate it and replace with a new one.

awick:
You say "stop routing traffic to it", but what traffic is going thru the LB to the capture program?  I feel like I'm still missing something.  Can you send the vxlan traffic to a LB?  (I haven't used it yet, so I'm behind)

Are using the AWS load balancer not to actually balance load in this case but to just perform health checks and auto restart instances?

I still wonder if you are calling it healthchecks if it should also have configurable options to make sure its seeing traffic over certain time periods.  For example if it has seen 0 packets over 60 second you might want to kill the instance also.

fj604:
@awick as discussed, VXLAN traffic is indeed going through the LB to the capture program. You can see the [solution diagram:](https://github.com/fj604/moloch-vpc-cfn/blob/master/diagram.svg) - VXLAN traffic is sent to the Network Load Balancer at the bottom.

PR #1370, opened 2020-01-30
Fixes the problem where the s3 token expires during a capture

### LINKED ISSUE #1366: writer_s3_part_cb(): Bad Response: 400
**Moloch version**: V2.1.0

**OS name and version**: Linux 

**How was Moloch built/installed:** easybutton

It is configured to upload pcaps to S3. However, sometimes I get log messages like:

Jan 28 04:26:17 writer-s3.c:139 writer_s3_part_cb(): Bad Response: 400 s3://us-east-1/blah-blah/2020-01-ingest/a4-200128-164.pcap.gz

One problem is that the moloch-capture process still exits with return code 0 so that wrapper script thinks that the upload has been successful. Also, the data about the packets has been written into elastic search, so you can see the session in the web UI, but have no access to the pcap data. 

Another problem is that there is insufficient information displayed to reveal what the error actually is -- my suspicion is that the upload was too slow and S3 timed out.

pjsg:

If moloch-capture is running on AWS, pushing files to S3, and is using IAM Roles for access permissions, then the token can expire during the run. The fix is to refetch the token just before every request to upload a part of the resulting file to S3.

This fixes #1366 

N/A

N/A

## License

### DIFF
--- a/capture/plugins/writer-s3.c
+++ b/capture/plugins/writer-s3.c
@@ -189,9 +189,6 @@ void writer_s3_refresh_s3credentials(void)
         newS3AccessKeyId = moloch_js0n_get_str(credentials, rlen, "AccessKeyId");
         newS3SecretAccessKey = moloch_js0n_get_str(credentials, rlen, "SecretAccessKey");
         newS3Token = moloch_js0n_get_str(credentials, rlen, "Token");
-    } else {
-        printf("Cannot retrieve credentials from metadata service at %s -- no returned data\n", role_url);
-        exit(1);
     }
 
     if (newS3AccessKeyId && newS3SecretAccessKey && newS3Token) {
--- a/capture/plugins/writer-s3.c
+++ b/capture/plugins/writer-s3.c
@@ -65,6 +65,7 @@ LOCAL  SavepcapS3File_t      fileQ;
 
 
 LOCAL  void *                s3Server = 0;
+LOCAL  void *                metadataServer = 0;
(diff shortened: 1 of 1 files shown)


pjsg:
This works on the 2.2.1 tag -- just reverifying on master.....

awick:
Just curious, how expensive is the writer_s3_refresh_s3credentials call?  Would it better to use g_timeout_add_seconds and just call everytime it is about to expire?  I don't care either way, just wondering.

pjsg:
It isn't anything like as expensive as putting a chunk of data to S3.... I think that it only needs to be done once every five minutes, but the added complexity didn't seem worth it. I'm a little unclear on the threading/event model used in the capture process, so I wanted to disturb as little as possible.

fj604:
Maybe store expiration time when retrieving credentials and only attempt to refresh them if they are about to expire? This check can be made within writer_s3_refresh_s3credentials().

pjsg:
That makes sense. I'll add that.

awick:
Also, the void *metadataServer = moloch_http_create_server("http://169.254.169.254", 10, 10, 0); could just be done once into a global and remove all the other creates/frees(except at exit).  I missed this before. Not a huge deal, but its causing a bunch of things to happen that aren't required.

pjsg:
I'll let this run overnight to make sure that there is nothing unexpected....

pjsg:
It seemed to work fine overnight.... This is good to go.

PR #1372, opened 2020-01-31
Support custom s3 repositories like minio

Jc2k:

The s3 plugins do work with other implementations of the s3 protocol pretty much as is. We ran moloch against a Pithos s3 instance for over a year pretty successfully but had to carry a custom patch to force the "endpoint" setting for the viewer and to set s3Host in capture. Recently set this up again with a recent version of moloch and minio, and again it is working well.

This patch makes exposes a `s3Host` so that others can choose to use a non-s3 s3 too.

N/A

No errors for the viewer, and N/A for parliament.

Still working on getting these working in my env and its the end of the day here, but wanted to get this up for review in the meantime.

## License

### DIFF
--- a/capture/plugins/writer-s3/index.js
+++ b/capture/plugins/writer-s3/index.js
@@ -63,7 +63,9 @@ function makeS3 (node, region) {
     s3Params.endpoint = Config.getFull(node, 's3Host')
   }
 
-  if (Config.getFull(node, 's3PathAccessStyle') === 'true') {
+  var bucket = Config.getFull(node, 's3Bucket')
+  var bucketHasDot = bucket.indexOf('.') >= 0
+  if (Config.getBoolFull(node, 's3PathAccessStyle', bucketHasDot) === true) {
     s3Params.s3ForcePathStyle = true;
   }
 
--- a/capture/plugins/writer-s3/index.js
+++ b/capture/plugins/writer-s3/index.js
@@ -59,6 +59,14 @@ function makeS3 (node, region) {
     s3Params.secretAccessKey = secret;
   }
 
+  if (Config.getFull(node, 's3Host') !== undefined) {
+    s3Params.endpoint = Config.getFull(node, 's3Host')
+  }
+
+  if (Config.getFull(node, 's3PathAccessStyle') === 'true') {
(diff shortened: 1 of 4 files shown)


viewer/viewer.js
```diff
@@ -8868,8 +8868,12 @@ function main () {
     internals.previousNodesStats.push(info.nodes);
   });
 
-  expireCheckAll();
-  setInterval(expireCheckAll, 60*1000);
+  var pcapWriteMethod = Config.get("pcapWriteMethod");
```
awick: Good idea, but I think these 6 lines will need to be moved to below loadPlugins for it to work

capture/plugins/writer-s3/index.js
```diff
@@ -59,6 +59,14 @@ function makeS3 (node, region) {
     s3Params.secretAccessKey = secret;
   }
 
+  if (Config.getFull(node, 's3Host') !== undefined) {
+    s3Params.endpoint = Config.getFull(node, 's3Host')
+  }
+
+  if (Config.getFull(node, 's3PathAccessStyle') === 'true') {
```
awick: So if we want to match the capture side this would need to get the s3Bucket var and set default to true if there is a period

   s3PathAccessStyle     = moloch_config_boolean(NULL, "s3PathAccessStyle", strchr(s3Bucket, '.') != NULL);

viewer/viewer.js
Jc2k: Ah of course. I've done this in my latest commit.

capture/plugins/writer-s3/index.js
Jc2k: Added a getBool wrapper and updated this to match the C code.

awick: thanks

awick:
Do you have time to do a PR on https://github.com/aol/molochweb and add this setting to the writer s3 section?

Jc2k:
Ah i was wondering where that page lived - you mean [this one](https://github.com/aol/molochweb/blob/master/_wiki/s3.md) right?

awick:
So feel free to add some text there too, but I meant the https://github.com/aol/molochweb/blob/master/settings.html page around line 2400

Issue #1377, opened 2020-02-11
Multiple errors during execution of large Hunt-Jobs


mammo0:
**Elasticsearch version**:
7.5.1

**Moloch version**:
2.2.1

**OS name and version**:
Ubuntu 18.04
in Docker container (see: https://github.com/mammo0/docker-moloch)

**How was Moloch built/installed:** (rpm, deb, easybutton, ...)
deb

**Provide logs, stack traces and steps to reproduce**:

Hello,

in my Moloch environment I've got a large amount of network data to search through with a hunt job (~10.000.000 sessions). At first the job ran fine, but at around 30% it threw the following error:

```
Hunt error searching sessions: [search_context_missing_exception] No search context found for id [52501] (and) [...]
```

After a little search I found out that this kind of error is thrown by Elasticsearch if a timeout is hit... But I don't figured out what timeout exactly, and where it can be increased. Maybe you've got an idea.

After that I tried to simply rerun/continue the above hunt job. It immediately threw another error:
```
 Hunt error searching sessions: [illegal_argument_exception] Batch size is too large, size must be less than or equal to: [10000] but was [6222549]. Scroll batch sizes cost as much memory as result windows so they are controlled by the [index.max_result_window] index level setting. (and) [...]
```

I figured out that the above number `6222549` was calculated in the line
https://github.com/aol/moloch/blob/b2d8e312a89f658741c29c79b8f035d52c344859/viewer/viewer.js#L7260
I don't know why the query size is calculated here... If I simply comment out this line (resp. including the surrounding `if`-block), the job started normally again. Now the job is running and I will report later if another error occurs.

But until then you may have some ideas regarding the first error above and can explain why this query size calculation is done.

Thank you.

mammo0:
> Now the job is running and I will report later if another error occurs.

Immediately after I submitted this issue, the restarted process failed again with the same error message as the first time:
```
Hunt error searching sessions: [search_context_missing_exception] No search context found for id [759] (and) [...]
```

I saw that in viewer.js there is the value `esQueryTimeout`:
https://github.com/aol/moloch/blob/b2d8e312a89f658741c29c79b8f035d52c344859/viewer/viewer.js#L70

By default it is set to 300s resp. 5min. I increased the value in the config.ini file. The option `elasticsearchTimeout` didn't exist in the `[default]` section of the config.ini, so I added it with a value of 900 (15min).
I will report back if this helps.

mammo0:
[...]

PR #1385, opened 2020-03-03
Fix for #1377

mammo0:
Please have a look at #1377.

I described my problem there in detail (also see my comments in this issue) and what solutions I found. If you have additional questions, just ask them.
Thank you.

### DIFF
--- a/viewer/viewer.js
+++ b/viewer/viewer.js
@@ -68,6 +68,7 @@ var internals = {
   CYBERCHEFVERSION: '9.11.7',
   elasticBase: Config.getArray('elasticsearch', ',', 'http://localhost:9200'),
   esQueryTimeout: Config.get("elasticsearchTimeout", 300) + 's',
+  esScrollTimeout: Config.get("elasticsearchScrollTimeout", 600) + 's',
   userNameHeader: Config.get("userNameHeader"),
   requiredAuthHeader: Config.get("requiredAuthHeader"),
   requiredAuthHeaderVal: Config.get("requiredAuthHeaderVal"),
@@ -7242,7 +7243,7 @@ function runHuntJob (huntId, hunt, query, user) {
   let options = buildHuntOptions(huntId, hunt);
   let searchedSessions;
 
-  Db.search('sessions2-*', 'session', query, {scroll: '600s'}, function getMoreUntilDone (err, result) {
+  Db.search('sessions2-*', 'session', query, {scroll: internals.esScrollTimeout}, function getMoreUntilDone (err, result) {
     if (err || result.error) {
       pauseHuntJobWithError(huntId, hunt, { value: `Hunt error searching sessions: ${err}` });
       return;
@@ -7310,7 +7311,7 @@ function runHuntJob (huntId, hunt, query, user) {
 
       // There might be more, issue another scroll
       if (result.hits.hits.length !== 0) {
-        return Db.scroll({ body: { scroll_id: result._scroll_id }, scroll: '600s' }, getMoreUntilDone);
+        return Db.scroll({ body: { scroll_id: result._scroll_id }, scroll: internals.esScrollTimeout }, getMoreUntilDone);
       }
 
       Db.clearScroll({ body: { scroll_id: result._scroll_id } });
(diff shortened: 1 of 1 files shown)


awick:
Its not clear to me any of these changes are needed after the updates we've made.
What scroll timeout are you using?  Any reason to just not use that everywhere?

awick:
* Ah I missed the "get the size of the query if it is being restarted" change, that should be removed
* I'm not sure another setting for esScrollTimeout is useful, if you have a better number lets just use it
* the remote query stuff shouldn't use the esQueryTimeout since its not a es query.  We currently have it set to 20 minutes, which seems like it should be enough.

mammo0:
## 1)
> Ah I missed the "get the size of the query if it is being restarted" change, that should be removed

I removed these lines of code, because of the `Batch size is too large, size must be less than or equal
to: [10000]` error message. If this calculation is necessary, can you please explain me why?


## 2)
> I'm not sure another setting for esScrollTimeout is useful, if you have a better number lets just use it

In my current setup I set this to 1800s. But this value was needed because I performed a Hunt search through around 10TB of network data. This job took about 1,5 weeks... I don't think that this high value should be used as a default.
The ES docu says about a high value (https://www.elastic.co/guide/en/elasticsearch/reference/7.5/search-request-body.html#scroll-search-context):
> Keeping older segments alive means that more disk space and file handles are needed. Ensure that you have configured your nodes to have ample free file handles. See File Descriptors.

So I think a configurable parameter would make sense here...


## 3)
> the remote query stuff shouldn't use the esQueryTimeout since its not a es query. We currently have it set to 20 minutes, which seems like it should be enough.

I ran into this issue because I use Docker to run the Moloch viewer. Every time I created a new container (e.g. after an update) the hostname of the container changed to a random value.
Because of that the current viewer instance thought that a previously added Hunt job was added by a remote viewer and therefore performed the remote requests. And for those remote requests the timeout was I think only 60s.

In my case I fixed this issue by adding `--host localhost` parameter when launching the viewer. This overrides the random hostname of the Docker container. And now the viewer performs local queries, not remote ones.

I saw that you increased this timeout to a hard coded value of 20min. But as you see in 2) I'm not a fan of hard coded values ;)
So I used the `esQueryTimeout` value here. Does this remote call not initiate a query on the remote viewer? If so, why do not use the same timeout as for a local query here? Maybe a some extra time should be added to avoid latency problems.

Please let me know, what you think.

awick:
>

> ## 1)
> > Ah I missed the "get the size of the query if it is being restarted" change, that should be removed
> 
> I removed these lines of code, because of the `Batch size is too large, size must be less than or equal to: [10000]` error message. If this calculation is necessary, can you please explain me why?

Yes, I was agreeing with you that this part of the PR is needed, sorry it wasn't clear.

> 
> ## 2)
> > I'm not sure another setting for esScrollTimeout is useful, if you have a better number lets just use it
> 
> In my current setup I set this to 1800s. But this value was needed because I performed a Hunt search through around 10TB of network data. This job took about 1,5 weeks... I don't think that this high value should be used as a default.
> The ES docu says about a high value (https://www.elastic.co/guide/en/elasticsearch/reference/7.5/search-request-body.html#scroll-search-context):
> 
> > Keeping older segments alive means that more disk space and file handles are needed. Ensure that you have configured your nodes to have ample free file handles. See File Descriptors.
> 
> So I think a configurable parameter would make sense here...

What would it hurt setting to 1800s everywhere?  Just need to make sure we do a clearScroll.  What this value should be is just how long it takes to search 100 sessions.


> 
> ## 3)
> > the remote query stuff shouldn't use the esQueryTimeout since its not a es query. We currently have it set to 20 minutes, which seems like it should be enough.

> I saw that you increased this timeout to a hard coded value of 20min. But as you see in 2) I'm not a fan of hard coded values ;)

I'm not a fan just adding a million new config vars, plus defaults should be sane.  
It shouldn't be the same as esQueryTimeout because what is taking the majority of the time is the searching of the pcap, not going against ES.  Also I noticed there is a 10minute timeout on the receiving side that needs to be updated anyway.

mammo0:
> Yes, I was agreeing with you that this part of the PR is needed, sorry it wasn't clear.

Ok, misunderstood this point.

> What would it hurt setting to 1800s everywhere? Just need to make sure we do a clearScroll. What this value should be is just how long it takes to search 100 sessions.

How you mean. I didn't test a lower value in my setup. I just set this value and it worked. Maybe the point is, that I had a lot of sessions with the maximum packet size of 10.000. Maybe a value of 20min (1200s) works as well... So this would be consistent to the other 20min timeouts...

> I'm not a fan just adding a million new config vars, plus defaults should be sane.

Hehe :)
It's your software, so you decide ;) I just can make suggestions.

> It shouldn't be the same as esQueryTimeout because what is taking the majority of the time is the searching of the pcap, not going against ES. Also I noticed there is a 10minute timeout on the receiving side that needs to be updated anyway.

I'm not as deep as you in the whole Moloch source code. So I think you know more what you do than me.

awick:
Thanks, I merged the first 2 parts

Issue #1403, opened 2020-03-25
Feature Request: Collapsible Navbars (especially Connections Tab


amaddalena:
can we please have collapsible navbars, especially for the connections tab?

yay @31453

BryceV:
Hi @amaddalena, collapsible navbars added with #1412. Cheers!

PR #1412, opened 2020-04-07
Collapsible Navbars

BryceV:
For  #1403. Note that the settings page does not have any toolbars (edit: and the help, and upload pages) and the load indicator for spiview is inside their toolbar.

### DIFF
--- a/viewer/vueapp/src/components/connections/Connections.vue
+++ b/viewer/vueapp/src/components/connections/Connections.vue
@@ -1564,10 +1564,6 @@ export default {
 </script>
 
 <style scoped>
-.connections-page {
-  /* account for main navbar height */
-  margin-top: 36px;
-}
 
 .connections-graph {
   /* don't allow selecting text */
--- a/viewer/vueapp/src/components/connections/Connections.vue
+++ b/viewer/vueapp/src/components/connections/Connections.vue
@@ -329,106 +329,110 @@
         </div>
       </div> <!-- /popup area -->
 
-      <!-- zoom in/out -->
-      <div class="btn-group-vertical zoom-btns overlay-btns">
-        <span title="Zoom in"
-          v-b-tooltip.hover.lefttop>
-          <button type="button"
-            class="btn btn-default btn-sm"
-            :class="{'disabled':zoomLevel >= 4}"
-            @click="zoomConnections(2)">
-            <span class="fa fa-lg fa-search-plus">
-            </span>
-          </button>
-        </span>
-        <span title="Zoom out"
-          v-b-tooltip.hover.lefttop>
-          <button type="button"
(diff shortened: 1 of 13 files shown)


awick:
Very cool! 

Few minor comments
* The tool tip should start with capital letter.  Also not sure if open/close are the right words.  Maybe Hide/Unhide?
* The spacing between the (i) and the Collapse icon doesn't look right
* Anyway to make it so the toolbars disappear before/as everything slides up
* Help page shows the button even though no toolbar, while the settings page hides the button.  they should either both show or hide (i would vote for show, but ...)

31453:
This is awesome! ✨The connections page really benefits from this feature. The use of the store and the MolochCollapsible component make the code really clean and easy to understand. 

Also, I had read about `$nextTick`, but not used it yet. It's AWESOME! I think there are a few other places in the app where it would make sense to use this 👍(but I have to find them)

In reference to Andy's comments:
- I like Hide/Unhide for the wording of the title of the toggle button. I also think you might consider using a [tooltip](https://bootstrap-vue.js.org/docs/components/tooltip/) instead of just a title attribute.
- I don't mind the spacing. @awick what doesn't look right? too much space?
- I don't think making it so the toolbars disappear before/as everything slides up is big deal if it's difficult to implement. Honestly, I didn't even notice. You might consider adding a transition to the hiding/showing of the toolbar. Vue has a good [tutorial on transitions](https://vuejs.org/v2/guide/transitions.html) (and there are a few examples in the code).
- I vote for hiding the toggle on any page that doesn't have tool bars (to go to the help page, click the owl on the top left, it's kind of a hidden page IMO).

awick:
> * I don't mind the spacing. @awick what doesn't look right? too much space?

too much and not even.  Here is what I see.  I would say the spacing between each component should be the same, right now its kind of random.  At the very least the spacing between the T(i) and (I)^ should be the same.
image


> * I don't think making it so the toolbars disappear before/as everything slides up is big deal if it's difficult to implement. Honestly, I didn't even notice. 

Ya one of the first things I noticed :)  The slide up is animated, goes under the toolbars and then the tool bars just jump away.  I agree if its too hard don't change, but if the order could be reversed that would be awesome.

BryceV:
Good ideas. Here are the changes:

- Removed the collapse arrow from 'help' and also the 'upload' page which was over looked. The 'help' page needed a bit of logic for the navbar to recognize it as an activePage.
- Fixed spacing around the collapse arrow
- Edited the title and put it in a tooltip
- I did get the bootstrap animation to work better, but it might not be a worthy solution. I remove the position:fixed property momentarily so bootstrap can collapse them cleanly. It will override the position property though, regardless of how it was previously  set (for inline and scoped positions...which is not common). Let me know what you would prefer for this one.

PR #1417, opened 2020-04-08
toggle fixing the timeline/map visualizations to the top of the screen

### LINKED ISSUE #1081: Feature Request: keep graph but enable scrolling of table contents
the feature request is to adjust or augment how scrolling works.   on the session page, would it be possible to keep the graph and table header (and the info above like search bar etc.) and just scroll the table contents.

31453:
save whether the viz is fixed in localstorage (per page)
fix some logic in spigraph
fix some tooltip issues
fixes #1081

### DIFF
--- a/viewer/vueapp/src/components/spigraph/Spigraph.vue
+++ b/viewer/vueapp/src/components/spigraph/Spigraph.vue
@@ -143,6 +143,18 @@
       </form>
     </MolochCollapsible>
 
+    <!-- main visualization -->
+    <div v-if="spiGraphType === 'default' && mapData && graphData && fieldObj">
+      <moloch-visualizations
+        id="primary"
+        :graph-data="graphData"
+        :map-data="mapData"
+        :primary="true"
+        :timezone="user.settings.timezone"
+        @fetchMapData="cancelAndLoad(true)">
+      </moloch-visualizations>
+    </div> <!-- /main visualization -->
+
     <div class="spigraph-content">
 
       <!-- pie graph type -->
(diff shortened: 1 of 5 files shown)


viewer/vueapp/src/components/spigraph/Spigraph.vue
```diff
@@ -511,8 +511,9 @@ export default {
         respondedAt = Date.now();
         this.error = '';
         this.loading = false;
-        this.items = []; // clear items
-        this.processData(response.data);
+        this.items = response.data.items;
```
BryceV, line 514: Maybe we could add this to Connections.vue?

BryceV:
I think the sticky toggle button covers the timeline text. Might need a bit of padding there but otherwise it all looks good!

viewer/vueapp/src/components/spigraph/Spigraph.vue
```diff
@@ -511,8 +511,9 @@ export default {
...
+        this.mapData = response.data.map;
```
BryceV, line 515: Can we add these inits to Sessions .vue

31453: I'm not sure what you mean. Can you clarify?

BryceV: The initializing for mapData and graphData. They were set to undefined like this line previously.

31453: Where were they set to undefined? 

So in Sessions.vue the mapData and graphData are set the same way as in Spigraph.vue. I took out the processData step because it is not needed and I wanted to simplify the code. Before, the processData function was used to set the mapData and graphData. But now it's all done here, like in Sessions.vue.

I think I'm still misunderstanding you though.

BryceV: Oops I see the confusion. I was looking at this in the "Files changed" tab. This was placed right under where vue data was set, so I thought it was also setting up data. Never mind, this works :)

31453: 👍 thanks for clarifying. I find the diff to be super confusing sometimes

PR #1419, opened 2020-04-09
add info for action items in session detail

31453:


31453:
@awick want to provide better wording? Also, what do you think about the icons?

PR #1430, opened 2020-04-27
Sticky table header

### LINKED ISSUE #1416: Feature Request: fix table header
allow a user to fix the table header on the sessions page and scroll the body of the table

31453:
Make the sessions table header stick to the top of the page with the sticky visualization. The table is scrollable but the header will always stay at the top in the same position.

One caveat: a user cannot scroll the headers left/right. So if the table is wider than the window, and they scroll right, the table body cells no longer match the header columns.

Fixes #1416

### DIFF
--- a/viewer/vueapp/src/components/sessions/Sessions.vue
+++ b/viewer/vueapp/src/components/sessions/Sessions.vue
@@ -54,6 +54,7 @@
 
       <!-- sessions results -->
       <table class="table-striped sessions-table"
+        :class="{'sticky-header':stickyHeader}"
         :style="tableStyle"
         id="sessionsTable">
         <thead>
@@ -197,7 +198,7 @@
               <th v-for="header of headers"
                 :key="header.dbField"
                 class="moloch-col-header"
-                :style="{'width': header.width + 'px'}"
+                :style="{'width': `${header.width}px`}"
                 :class="{'active':isSorted(header.sortBy || header.dbField) >= 0, 'info-col-header': header.dbField === 'info'}">
                 <!-- non-sortable column -->
                 <span v-if="header.dbField === 'info'"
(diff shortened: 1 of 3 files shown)


BryceV:
Interesting! I see what you are doing here. A couple things I noticed:
- If you open a menu in the sticky state, then deselect sticky, the dropdown for the table headers will be squashed. 
- The headers act strangely in sticky mode when there are only a few columns (try 1 to 3). This is close to what you were saying about the caveat about the table stretching horizontally. 
- "Fit the window to the current table size" button is part of the vertical scroll on the thead > tr when sticky is on. It gets pushed far down and is hidden. 
- In sticky mode the overflow-x: scroll adds a border under the table header. Its a design choice but maybe overflow-x: auto works better if we want to remove that border.

31453:
- How do I reproduce this one? I'm clicking on a dropdown while in sticky mode, then unselecting sticky mode, then opening a dropdown again. I don't see any changes to the dropdown or the table headers. Which dropdown are you clicking? What are your exact steps to reproduce?
- This is a strange one! I think I might need to re-render the entire table when toggling sticky mode 😕 
- Nice find! 👍 
- Auto doesn't provide the same behavior in letting the dropdowns overflow the column header row container unfortunately, so scroll is required. I'm not seeing the bottom border. There is however a drop shadow to try to indicate that the header is stuck to the top.

31453:
Never mind on the first bullet point! I was able to reproduce it (I was closing the dropdown before clicking the unstick button).

BryceV:
Nice! I still see the fit button scrolling with the page and it crosses over the collapsable section. Also, the 310px top is not going to respond to the collapsing action. 

It looks like you did something with the thead bottom border. There is a shadow but also a border on chrome when its in sticky mode and  after a dropdown toggle. I think its added from the overflow-x.

31453:
I already forgot about that new feature 🤦 Thanks for the reminder!

PR #1433, opened 2020-04-30
fix styles in sticky sessions when navbar is collapsed

### LINKED ISSUE #1432: Sticky sessions need better styling when the navbar is collapsed
To see this collapse the navbar and open the detail of at least one session. Note:

- the sticky sessions button hovers over the map/graph
- when the sticky sessions are opened, they do not take up the full space and it looks cut off at the top

31453:
move the sticky session button up
move the sticky session detail up
fixes #1432

### DIFF
--- a/viewer/vueapp/src/components/sessions/Sessions.vue
+++ b/viewer/vueapp/src/components/sessions/Sessions.vue
@@ -728,6 +728,8 @@ export default {
         let index = this.stickySessions.indexOf(session);
         if (index >= 0) { this.stickySessions.splice(index, 1); }
       }
+
+      this.$store.commit('setStickySessionsBtn', !!this.stickySessions.length);
     },
     /**
      * Closes the session detail for a session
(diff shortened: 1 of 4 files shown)


BryceV:
Good find. I am also confused why this is the "sticky sessions button" instead of the thumbtack button. But semantics aside, there was only one thing I noticed:
- Close the toolbar
- Open a sessions detail
- From within the menu opened with the "sticky sessions" button" (It says "X open sessions" on the top") close all the open sessions. This hides the yellow button but keeps the same width on the navbar.

BryceV:
Also, I might be missing something, but why is 'setStickySessionsBtn' getting set to the store? Do we need it somewhere else?

31453:
Not sure what you mean by the thumbtack button. The thumbtack button is for the sticky visualizations and sticky sessions header. The sticky sessions button is the button for displaying the opened sessions in the body of the sessions table. 

Nice find! I'll fix that.

The sticky sessions button needs to be in the store for two reasons: 
1. I use it in `Navbar.vue` to add a class to the navbar so it knows to push the info to the left.
2. I set it in `Sessions.vue` to indicate that there are sessions opened.

PR #1436, opened 2020-05-02
Various ui fixes

BryceV:
Various fixes. Some of them existing and some found after visualization-additions. Review commit titles for details.

### DIFF
--- a/viewer/viewer.js
+++ b/viewer/viewer.js
@@ -1759,7 +1759,7 @@ app.post('/notifiers/:name/test', [noCacheJson, getSettingUserCache, checkCookie
 
 // gets a user's settings
 app.get('/user/settings', [noCacheJson, recordResponseTime, getSettingUserDb, checkPermissions(['webEnabled']), setCookie], (req, res) => {
-  let settings = Object.assign(settingDefaults, req.settingUser.settings);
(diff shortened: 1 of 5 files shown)


viewer/viewer.js
```diff
@@ -1759,7 +1759,7 @@ app.post('/notifiers/:name/test', [noCacheJson, getSettingUserCache, checkCookie
 
 // gets a user's settings
 app.get('/user/settings', [noCacheJson, recordResponseTime, getSettingUserDb, checkPermissions(['webEnabled']), setCookie], (req, res) => {
-  let settings = req.settingUser.settings || settingDefaults;
+  let settings = Object.assign(settingDefaults, req.settingUser.settings);
```
awick, line 1762: pretty sure this modifies target, so you'll need to clone settnigDefaults.

31453:
We actually don't want to remove the pie info popup when mouse leaves the graph slice. This is because the user can go to the popup and add hover over a value and add that value to the search expression.

Otherwise, looks great!

viewer/viewer.js
BryceV: Oh, yeah. Plus they both have arrays, so a clone of each is required.

BryceV:
Good catch, the pop up would be too fast. I removed the pop up for when the graph type changes instead (since that was a byproduct of mouseleave, I left it off before).

PR #1442, opened 2020-05-08
Use enum instead of #define

### LINKED ISSUE #1380: #define ⇒ enum?
Would you like to replace [more defines for constant values](https://github.com/aol/moloch/blob/664ffe25810380f12823941c210efad4e9881124/capture/moloch.h#L158 "MOLOCH_FIELD_TYPE_…") by [enumerations](https://en.wikipedia.org/wiki/Enumerated_type#C_and_syntactically_similar_languages "Description for the usage of enumerations") to stress their relationships?

komainu8:
Hello,

I fixed #1380 within the limits of my ability.

# License

### DIFF
--- a/capture/config.c
+++ b/capture/config.c
@@ -723,7 +723,7 @@ void moloch_config_load_header(char *section, char *group, char *helpBase, char 
                                                    NULL);
         snprintf(name, sizeof(name), "%s", keys[k]);
         int type = 0;
-        int t = 0;
+        MolochFieldTypes t = MOLOCH_FIELD_TYPE_INT;
         int unique = 1;
         int count  = 0;
         char *kind = 0;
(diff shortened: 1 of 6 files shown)


awick:
Everyplace that uses the enums needs to change from int to the new types that have been defined

komainu8:
Thank you for your comment.
You're right. I'll also modify them.

awick:
This is getting closer, thanks for working on it
1) can you remove the s at the name of the two types.  So MolochFieldType and MolochSessionType
2) There are several switch statements with errors now, mainly about missing MOLOCH_FIELD_TYPE_CERTSINFO you think you can fix those or want me to?

komainu8:
> 1. can you remove the s at the name of the two types. So MolochFieldType and MolochSessionType

I understand. I'm fixing them.

> 2. There are several switch statements with errors now, mainly about missing MOLOCH_FIELD_TYPE_CERTSINFO you think you can fix those or want me to?

Thank you for your comment.
I want to ask the revision of you. Because I don't understand this problem details.

PR #1443, opened 2020-05-08
Update copyright in moloch.h

komainu8:
Hello,

I've noticed that copyright is old in this file. Therefore, I fix it.

## License


awick:
Copyright dates actually don't need to be updated so we've stopped doing that

komainu8:
Thank you for your comment.
I see.

PR #1445, opened 2020-05-08
put eslint at top level

31453:
fix lint errors
remove lint from parliament and viewer dirs
update linting part of contributing file
update linting in screwdriver build


31453:
@awick is that the correct place to fix the linting for our builds?

PR #1453, opened 2020-05-20
Add an error handling when malloc failed

komainu8:
Hello,

We should exit when malloc failed. 
Because if we can't allocate memory, processing thereafter can't continue. 

## License


awick:
A pretty pointless check, if at startup can't malloc memory you are going to have many more issues.  If you are really interested in fixing all the mallocs, should look at using g_malloc which does the check for you (but need to make sure to use g_free)

komainu8:
Oh, I see.
Thank you for your comment.

Issue #548, opened 2016-10-20
WISE UI


awick:
Phase 1)
- [x] may need to add an api to return to the UI all the sources loaded (see internals.sources) and moloch types (see internals.types) loaded
- [x] allow querying from UI (show pull down of sources (add a ANY option by default) show pull down of moloch types (default to IP), and search box and then call the API /:type/:query)
- [x] dump sections in both json and csv format (maybe show all the configured sources with a dump button like /dump/:type api does)

Phase 2)
- [x] add a stats page, that calls the /stats and displays all the sources/types stats.  There are 2 different stats right now, types and sources, it might be good to have 2 tabs on the stats page one for each, since there might be more stats in the future, or we can do that latter. Each of the stats should be shows in table format with types/source names as row names, different stats as column headers.
- [x] use json for config format, so -c config.json
- [x] allow redis to be used for config storage, something like -c redis://host:port/db#/config.js (whatever is real redis url standard
- [ ] if redis is used poll ever so often for changes (or pub/sub?)
- [x] figure out auth

Phase 3)
- [x] add ability in UI to create csv/json/tagger formatted files
- [x] add new redis storage for "simple files", again with polling
- [x] add settings page that lets you CRUD right clicks that wise serves
- [x] some way to convert current config.ini to config.json

Phase 4)
- [ ] add help
- [ ] maybe add clear getting starting stuff when there is no config
- [ ] documentation about the UI itself on arkime.com

I don't remember these:
Example - right click for threatstream Id should allow you to open up wise UI that then calls the api to get datas.

PR #1495, opened 2020-07-22
Wise config support changes

BryceV:
#548 work
- Support for parsing json files 
- You can now store the config in JSON form in redis
- Started work for async pubsub when config it is stored in redis.  
- Created a simple README to log how to pass in a redis url for getting the config

### DIFF
--- a/tests/config.test.json
+++ b/tests/config.test.json
@@ -1,0 +1,242 @@
+{
+  "default": {
+    "elasticsearch": ", , 127.0.0.1:9200,",
+    "pcapDir": "/tmp",
+    "geoLite2ASN": "GeoLite2-ASN.mmdb",
+    "geoLite2Country": "GeoLite2-Country.mmdb",
+    "rirFile": "ipv4-address-space.csv",
+    "ouiFile": "oui.txt",
+    "parsersDir": "../capture/parsers;parsers",
+    "pluginsDir": "plugins;../tests/plugins;../capture/plugins",
+    "yara": "rules.yara",
+    "smtpIpHeaders": "X-Originating-IP:;X-Barracuda-Apparent-Source-IP:",
+    "keyFile": "",
+    "spiDataMaxIndices": "-1",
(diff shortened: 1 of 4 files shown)


wiseService/wiseService.js
```diff
@@ -847,12 +846,12 @@ app.get('/stats', [noCacheJson], function (req, res) {
   res.send(stats);
 });
 // ----------------------------------------------------------------------------
-if (getConfig('wiseService', 'regressionTests')) {
-  app.post('/shutdown', (req, res) => {
+app.post('/shutdown', (req, res) => {
```
awick, line 849: Why the change? I think not having api available is better.

tests/config.test.json
```diff
@@ -0,0 +1,242 @@
+{
```
awick, line 1: I would say until we have editing available we should should use the ini one, unless this is just for a test.

wiseService/package.json
```diff
@@ -17,7 +17,7 @@
     "build": "npm run bundle:min",
     "start": "node wiseService.js -c ../tests/config.test.ini",
     "prestart": "npm run bundle:min",
-    "dev": "nodemon wiseService.js -c ../tests/config.test.ini & webpack -d --watch --progress --hide-modules --display-error-details --config vueapp/build/webpack.dev.conf.js",
+    "dev": "nodemon wiseService.js -c ../tests/config.test.json & webpack -d --watch --progress --hide-modules --display-error-details --config vueapp/build/webpack.dev.conf.js",
```
awick, line 20: until editing is available lets use the ini one

tests/config.test.json
BryceV: I left the json version for testing

wiseService/wiseService.js
BryceV: 👍

wiseService/package.json
BryceV: Ok

PR #1501, opened 2020-08-03
Fix integer overflow

mmiszczyk:
This fixes a potential integer overflow bug in a TDS parser, which arises due to the use of signed integer for both comparison and memcpy, allowing for an overflow with very large values.

## License

### DIFF
--- a/capture/parsers/tds.c
+++ b/capture/parsers/tds.c
@@ -27,9 +27,10 @@ LOCAL  int userField;
 LOCAL int tds_parser(MolochSession_t *session, void *uw, const unsigned char *data, int remaining, int which)
 {
     TDSInfo_t *tds = uw;
+    size_t count = remaining;
 
-    remaining = MIN(remaining, TDS_MAX_SIZE - tds->pos[which]);
-    memcpy(tds->data[which] + tds->pos[which], data, remaining);
+    count = MIN(count, TDS_MAX_SIZE - tds->pos[which]);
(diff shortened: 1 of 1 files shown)


awick:
Was wondering if you could provide more info on the overflow? It should only happen when remaining is > 0x7fffffff, which shouldn't happen because the tcp layer won't buffer that many packets. Or I might be missing something obvious.

mmiszczyk:
Yes, this would require a packet with size larger than maximum signed int (specific value is platform dependent, might be 0x7fff on embedded but will probably be 0x7fffffff on servers, PCs etc.). I've run some fuzzers on just the parsing functions, not the entirety of Moloch, so it's possible that the overflow might be prevented on lower layers.

awick:
Ok. Are you running moloch on embedded? I thought I had checks that require 64bit, but maybe I don't. If you have a PR to do fuzzing just on the parsers that would be a good addition, right now I just have the one at the packet level.

mmiszczyk:
Didn't try running it on embded yet, only in VMs so far. My fuzzing code is not production-ready so far and not really integrated with the build system, it's a quick and dirty AFL-based solution which takes an input file, uses the first byte to pick a parser and passes the rest to the chosen parsing function.

awick:
So, since I don't think anything will happen on a 64bit system with this, I'm just going to close for now. Let me know if I'm missing something.

PR #1506, opened 2020-08-17
Add support for maxFileTimeM in simple writer

edynox:
Adding support for configuration parameter maxFileTimeM to the simple PCAP writer.

PCAP files are rotated based on time passed since the output file has been created if the parameter is specified. 

This has been supported by the now deprecated default writer.

## License

### DIFF
--- a/capture/writer-simple.c
+++ b/capture/writer-simple.c
@@ -382,9 +382,7 @@ LOCAL void writer_simple_write(const MolochSession_t * const session, MolochPack
             LOG("opened %d %s %d", thread, name, info->file->fd);
         g_free(name);
 
-        MOLOCH_LOCK(simpleQ);
         gettimeofday(&fileAge[thread], NULL);
-        MOLOCH_UNLOCK(simpleQ);
     }
 
     packet->writerFileNum = currentInfo[thread]->file->id;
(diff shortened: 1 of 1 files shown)


capture/writer-simple.c
```diff
@@ -380,6 +381,10 @@ LOCAL void writer_simple_write(const MolochSession_t * const session, MolochPack
         if (config.debug)
             LOG("opened %d %s %d", thread, name, info->file->fd);
         g_free(name);
+
+        MOLOCH_LOCK(simpleQ);
```
awick, line 385: Shouldn't need this LOCK/UNLOCK since writer_simple_check is called inside each packet thread just like writer_simple_check. I guess should add a comment to the top of writer_simple_check

capture/writer-simple.c
```diff
@@ -499,7 +504,11 @@ LOCAL void writer_simple_check(MolochSession_t *session, void *UNUSED(uw1), void
     if (now.tv_sec - lastSave[session->thread].tv_sec < 10)
         return;
 
-    writer_simple_process_buf(session->thread, 0);
+    if (config.maxFileTimeM > 0 && now.tv_sec - fileAge[session->thread].tv_sec >= config.maxFileTimeM*60) {
```
awick, line 507: I think you want to move this check above the 10 second idle check, otherwise it will only do this check if there have been no saves for 10 seconds.  Remove the else and leave that , 0 write down here though.

capture/writer-simple.c
edynox: Thanks! I wasn't sure. Fixed!

capture/writer-simple.c
edynox: Right. In the use case I was thinking about it didn't make a difference - avoiding situation, that the file is stuck unfinalized for days if there is no more traffic on the interface; the file can get removed by moloch viewer in the meantime and then new traffic is dropped on the floor. 

You are right, that in general, it should be checked anyway, so the rotation time is actually considered.
Fixed!

edynox:
Thanks for the review! Let me know if you see anything else.

Issue #1512, opened 2020-09-03
Hunts - handle nodes being down better


awick:
Currently if during a hunt a node is down or having an issue the hunt errors out and pauses. Instead we should save the session ids that failed with the hunt and then try them again at the end. My general thought is
* Do the hunt as normal, any sessionId that fails because of the remote node being down, save the session id in something like hunts.failedSessionIds array in write back to ES occasionally
* After the normal hunt has failed then go back thru all the failedSessionIds, deleting the ones that no longer fail
* Repeat until either no more failedSessionIds (success!) or a complete run thru the error with no changes (pause for error)
* May want to have a max failedSessionIds length (I think ES has a 32k array limit) in which case we should pause for error 
* If someone reruns it should delete the failedSessionIds

Bonus, the bar chart thingy should either 
a) not count the failed sessions as being done
b) should display 3 sections (failed, done, not done)

The details display should probably show how many are waiting to retry

PR #1515, opened 2020-09-04
Borked hunt

### LINKED ISSUE #1439: Feature Request - Indicate un-runnable hunts
Some hunts are beyond repair and cannot be rerun or replayed (for example: hunts that have bad regex). Show that these hunts are completely borked. Do not show the rerun, replay, or open buttons. The error should be prominent and indicate that this is a bad hunt and a user can do nothing with it except delete it and try again with a new hunt.

31453:
add new unrunnable flag to hunt index
set unrunnable flag when encountering bad regex
set unrunnable flag when encountering bad search expression
show unrunnable error in the hunt ui
remove rerun and replay buttons for unrunnable hunts

Screen Shot 2020-09-04 at 3 10 08 PM

fixes #1439


viewer/viewer.js
```diff
@@ -17,7 +17,7 @@
  */
 'use strict';
 
-const MIN_DB_VERSION = 62;
+const MIN_DB_VERSION = 65; // TODO ECR - need to up the version?
```
31453, line 20: @awick Do I need to up the db version here?

awick: Yes we probably do because the hunt_v2 is defined as static, if you try write a new field it will fail.

db/db.pl
```diff
@@ -2471,6 +2472,12 @@ sub huntsUpdate
       },
       "notifier": {
         "type": "keyword"
+      },
+      "unrunnable": {
+        "type": "boolean"
+      },
+      "failedSessionIds": {
```
BryceV, line 2479: Where is that getting used?

awick: I asked @31453 to add now so a new db version isn't required for #1512 1512

PR #1516, opened 2020-09-08
Update moloch_update_get.sh

sickcodes:
Update critical environmental variables.

Reduce `if $? == 0` to `&&`

Escapable environmental variables.

#976661

Shell code

No

## License

### DIFF
--- a/release/moloch_update_geo.sh
+++ b/release/moloch_update_geo.sh
@@ -6,21 +6,21 @@ TIMEOUT="${WGET_TIMEOUT:-30}"
 
 cd "${DEST_DIR}"
 
-# Remove ./ipv4-address-space.csv.tmp if it exists from a failed download, and then download the latest
-if [ -e ./ipv4-address-space.csv.tmp ]; then
-    rm -rf ./ipv4-address-space.csv.tmp
+# Remove ipv4-address-space.csv.tmp if it exists from a failed download, and then download the latest
+if [ -e ipv4-address-space.csv.tmp ]; then
+    rm -rf ipv4-address-space.csv.tmp
 fi
 
-wget -nv --timeout="${TIMEOUT}" -O ./ipv4-address-space.csv.tmp --no-check-certificate "https://www.iana.org/assignments/ipv4-address-space/ipv4-address-space.csv" \
-    && mv -f ./ipv4-address-space.csv.tmp ./ipv4-address-space.csv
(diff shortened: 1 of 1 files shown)


awick:
Sorry at this point I'm not going to accept a patch that changes directories used without more information.

sickcodes:
I can change it back to /tmp but that folder is publicly writable.

Previously, any non-privileged user could create a file called oui.txt or ipv4-address-space.csv and the script will fail.

sickcodes:
/opt is only readable and writeable by root

sickcodes:
Secondly, I changed cp to mv so the file is removed after use.

awick:
So the reason /tmp was used and the files were left around was so -N could be used. But since we no longer download the large geo files this isn't a big deal. If you don't want to write in /tmp then instead of writing in the random /top you could just write in the final location with a .tmp suffix before doing the move.

sickcodes:
I removed the paths altogether and added the .tmp suffix to the downloads.
I also added the force tag, `mv -f`, to ensure that the temporary file always goes over the old one, with zero "downtime"

sickcodes:
Cool with current working directory paths or do you still want final directory?

Would become pretty unreadable but I do it if you reckon it's a better place to download to.

release/moloch_update_geo.sh
```diff
@@ -2,25 +2,23 @@
 
 # Variables
 DEST_DIR="${MOLOCH_DIR:-BUILD_MOLOCH_INSTALL_DIR}/etc"
-TIMEOUT=${WGET_TIMEOUT:-30}
+TIMEOUT="${WGET_TIMEOUT:-30}"
```
awick, line 6: Just add a cd $DEST_DIR if you don't want to use $DEST_DIR/... everywhere

sickcodes: Resolved :)

release/moloch_update_geo.sh
```diff
@@ -2,25 +2,25 @@
...
 
-# Work on temp dir to not affect current working files
-cd /tmp
+cd "{$DEST_DIR}"
```
awick, line 7: i think it needs to be ${DEST_DIR}

sickcodes: Embarrassing typo resolved.

sickcodes: Working on ubuntu-16 fix

sickcodes:
A bit more going on that expected, working on it.

sickcodes:
Closed, in favour of other PR https://github.com/aol/moloch/pull/1518

PR #1518, opened 2020-09-09
Lesser update for moloch_update_geo.sh

sickcodes:
Lite version of https://github.com/aol/moloch/pull/1516

Fix escapable environmental variables

 #976661

No

No

## License

### DIFF
--- a/release/moloch_update_geo.sh
+++ b/release/moloch_update_geo.sh
@@ -7,16 +7,17 @@ TIMEOUT="${WGET_TIMEOUT:-30}"
 # Work on temp dir to not affect current working files
 cd /tmp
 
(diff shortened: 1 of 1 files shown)


awick:
Unfortunately this breaks one of the most important goals of the script, if the download fail don't break a running moloch. Also don't like the mkdir being there. What is actually the goal here? What is issue 976661?

sickcodes:
> Unfortunately this breaks one of the most important goals of the script, if the download fail don't break a running moloch. Also don't like the mkdir being there. What is actually the goal here? What is issue 976661?

Understood, I reverted the changes, just quoted the appropriate variables. Issue is via hackerone https://hackerone.com/reports/976661

PR #1524, opened 2020-09-16
handle nodes being down better

### LINKED ISSUE #1512: Hunts - handle nodes being down better

31453:
if remote node is down, skip the session and go back through skipped sessions later
consolidate redundant ui components for status and hunt details
don't display skipped hunts in progress bar
indicate number of skipped hunts in hunt details
if failed sessions exceed 10k, pause with an error

fixes #1512

### DIFF
--- a/viewer/viewer.js
+++ b/viewer/viewer.js
@@ -7701,15 +7701,14 @@ function continueHuntSkipSession (hunt, huntId, session, sessionId, searchedSess
   updateHuntStats(hunt, huntId, session, searchedSessions, cb);
 }
 
-let searchingFailedSessions = false; // TODO ECR - a better place to put this? internals?
 // if there are failed sessions, go through them one by one and do a packet search
 // if there are no failed sessions left at the end then the hunt is done
 // if there are still failed sessions, but some sessions were searched during the last pass, try again
 // if there are still failed sessions, but no new sessions coudl be searched, pause the job with an error
 function huntFailedSessions (hunt, huntId, options, searchedSessions, user) {
   let changesSearchingFailedSessions = false;
-  if (!searchingFailedSessions && hunt.failedSessionIds && hunt.failedSessionIds.length) {
-    searchingFailedSessions = true;
+  if (!options.searchingFailedSessions && hunt.failedSessionIds && hunt.failedSessionIds.length) {
+    options.searchingFailedSessions = true;
     // copy the failed session ids so we can remove them from the hunt
     // but still loop through them iteratively
     let failedSessions = JSON.parse(JSON.stringify(hunt.failedSessionIds));
@@ -7748,7 +7747,7 @@ function huntFailedSessions (hunt, huntId, options, searchedSessions, user) {
       }
 
       if (hunt.failedSessionIds && hunt.failedSessionIds.length === 0) {
-        searchingFailedSessions = false; // no longer searching failed sessions
+        options.searchingFailedSessions = false; // no longer searching failed sessions
         // we had failed sessions but we're done searching through them
         // so we're completely done with this hunt (inital search and failed sessions)
         hunt.status = 'finished';
@@ -7764,7 +7763,7 @@ function huntFailedSessions (hunt, huntId, options, searchedSessions, user) {
         // so keep going
         return continueProcess();
       } else if (!changesSearchingFailedSessions) {
-        searchingFailedSessions = false; // no longer searching failed sessions
+        options.searchingFailedSessions = false; // no longer searching failed sessions
         // there were no changes, we're still struggling to connect to one or
         // more renote nodes, so error out
         return pauseHuntJobWithError(huntId, hunt, {
@@ -7782,6 +7781,7 @@ function runHuntJob (huntId, hunt, query, user) {
 
   // look for failed sessions if we're done searching sessions normally
   if (hunt.searchedSessions === hunt.totalSessions && hunt.failedSessionIds && hunt.failedSessionIds.length) {
+    options.searchingFailedSessions = true;
     return huntFailedSessions(hunt, huntId, options, searchedSessions, user);
   }
 
@@ -7796,7 +7796,7 @@ function runHuntJob (huntId, hunt, query, user) {
       searchedSessions = hunt.searchedSessions || 0;
       // if the session query results length is not equal to the total sessions that the hunt
       // job is searching, update the hunt total sessions so that the percent works correctly
-      if (!searchingFailedSessions && hunt.totalSessions !== (result.hits.total + searchedSessions)) {
+      if (!options.searchingFailedSessions && hunt.totalSessions !== (result.hits.total + searchedSessions)) {
         hunt.totalSessions = result.hits.total + searchedSessions;
       }
     }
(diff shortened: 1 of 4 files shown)


viewer/viewer.js
```diff
@@ -7686,35 +7701,112 @@ function buildHuntOptions (huntId, hunt) {
...
     let hits = result.hits.hits;
-
     if (searchedSessions === undefined) {
       searchedSessions = hunt.searchedSessions || 0;
       // if the session query results length is not equal to the total sessions that the hunt
       // job is searching, update the hunt total sessions so that the percent works correctly
-      if (hunt.totalSessions !== (result.hits.total + searchedSessions)) {
-        hunt.totalSessions = result.hits.total + searchedSessions;
-      }
+      // TODO ECR - this borks rerunning through the failed sessions - the totalSessions gets incremented
```
31453, line 7796: @awick What was the original purpose of this? Why was the percent not working correctly? Because of the scroll stuff?

If I keep this in when there are failed sessions, it just keeps incrementing the searched sessions.

awick: If I remember right, sessions could be added during the scroll so previous totalSessions would be wrong.

31453: 👍 I left it in to see how it works with real data, since the test data never hits the scroll

viewer/viewer.js
```diff
@@ -7702,14 +7701,15 @@ function continueHuntSkipSession (hunt, huntId, session, sessionId, searchedSess
   updateHuntStats(hunt, huntId, session, searchedSessions, cb);
 }
 
+let searchingFailedSessions = false; // TODO ECR - a better place to put this? internals?
```
31453, line 7704: @awick thoughts on the best place to put this?

awick: Ya shouldn't be a global incase somehow multiple hunts are running or some other badness. I would look at having buildHuntOptions put it inside of options if that would work.

31453: 👍 I put it on the hunt options object

viewer/viewer.js
```diff
@@ -7686,24 +7686,124 @@ function buildHuntOptions (huntId, hunt) {
   return options;
 }
 
+// if we couldn't retrieve the seession, skip it but add it to failedSessionIds
+// so that we can go back and search for it at the end of the hunt
+function continueHuntSkipSession (hunt, huntId, session, sessionId, searchedSessions, cb) {
+  if (!hunt.failedSessionIds) {
+    hunt.failedSessionIds = [ sessionId ];
+  } else {
+    if (hunt.failedSessionIds.length > 10000) { // TODO ECR - is this a good limit? ES doesn't specifically have an array size limit, but it does get costly to update a document if there is a large field
```
31453, line 7695: @awick what do you think the limit should be?

awick: 10k seems a fine place to start

31453: 👍

viewer/viewer.js
```diff
@@ -7686,24 +7686,126 @@ function buildHuntOptions (huntId, hunt) {
...
+  return updateHuntStats(hunt, huntId, session, searchedSessions, cb);
+}
+
+// if there are failed sessions, go through them one by one and do a packet search
+// if there are no failed sessions left at the end then the hunt is done
+// if there are still failed sessions, but some sessions were searched during the last pass, try again
+// if there are still failed sessions, but no new sessions coudl be searched, pause the job with an error
+function huntFailedSessions (hunt, huntId, options, searchedSessions, user) {
+  let changesSearchingFailedSessions = false;
+  if (hunt.failedSessionIds && hunt.failedSessionIds.length) {
```
awick, line 7717: style thing, would prefer a ! with a return here, so the rest of the function isn't indented.

viewer/viewer.js
```diff
@@ -7686,24 +7686,126 @@ function buildHuntOptions (huntId, hunt) {
...
+    options.searchingFailedSessions = true;
+    // copy the failed session ids so we can remove them from the hunt
+    // but still loop through them iteratively
+    let failedSessions = JSON.parse(JSON.stringify(hunt.failedSessionIds));
+    // we don't need to search the db for session, we just need to search each session in failedSessionIds
+    async.forEachLimit(failedSessions, 3, function (sessionId, cb) {
+      Db.get(Db.sid2Index(sessionId), 'session', Db.sid2Id(sessionId))
```
awick, line 7724: this should use Db.getSession if possible

awick (commented):
minor stuff, feel free to ignore

viewer/viewer.js
31453: 👍

viewer/viewer.js
31453: Does Db.getSession get the session even if it's on a remote node? So I won't have to do the make request part too?

31453: or do something like this:

  ```
Db.getSession(sessionId, {}, (err, session) => {
      if (err) {
        return continueHuntSkipSession(hunt, huntId, session, sessionId, searchedSessions, cb); 
      }
      
      makeRequest(session.node, `${session.node}/hunt/${huntId}/remote/${sessionId}`, user, (err, response) => {
        if (err) {
          return continueHuntSkipSession(hunt, huntId, session, sessionId, searchedSessions, cb);
        }

        let json = JSON.parse(response);

        if (json.error) {
          console.log(`Error hunting on remote viewer: ${json.error} - ${path}`);
          return continueHuntSkipSession(hunt, huntId, session, sessionId, searchedSessions, cb);
        }

        // remove from failedSessionIds if it was found
        hunt.failedSessionIds.splice(hunt.failedSessionIds.indexOf(sessionId), 1);
        // there were changes to this hunt; we're making progress
        changesSearchingFailedSessions = true;

        if (json.matched) { hunt.matchedSessions++; }
        return updateHuntStats(hunt, huntId, session, searchedSessions, cb);
      });
    });
```

31453: And I don't need to pass it options because I don't need any fields. I just need the packets. Right?

awick: Oh i just realized it doesn't have a promise version :( And yes you can just do {} for options, although I think you just need the ['lastPacket', 'node', 'huntId', 'huntName', 'fileId'] fields.

31453: 👍 works perfectly. I forgot that function existed!

PR #1525, opened 2020-09-18
disable sending referer

awick:
Hoping to disable sending the referer header to our 3 apps.  https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referrer-Policy


31453 (approved):
👍 Works for me. I don't think this will break anything because we don't use the referrer header anywhere.

PR #1527, opened 2020-10-06
Share hunts between users

### LINKED ISSUE #1438: Feature Request - Share Hunts
- [x] Add the ability to share hunts between users:

Currently, a **non-admin** user can only view information, results, and rerun/repeat their **own** hunts. A hunt should be able to be exposed to non-admin users so that any user of a cluster can access a hunt. When shared, all users should be able to view the information for the hunt and view the results for the hunt.

---

- [ ] Add the ability to share hunts between clusters:

Currently, hunts are stored per cluster and cannot be shared between clusters. A user may want to run hunts against multiple clusters.

31453:
A user creating a hunt can share it with users if they know the user's user ID. A user can also share already completed and in progress hunts with other users.

fixes #1438


viewer/viewer.js
```diff
@@ -8116,13 +8155,43 @@ app.post('/hunt', [noCacheJson, logAction('hunt'), checkCookieToken, checkPermis
...
   };
 
-  Db.createHunt(hunt, function (err, result) {
-    if (err) { console.log('create hunt error', err, result); }
-    hunt.id = result._id;
-    processHuntJobs(() => {
-      return res.send(JSON.stringify({ success: true, hunt: hunt }));
+  function doneCb (hunt, invalidUsers) {
+    Db.createHunt(hunt, function (err, result) {
+      if (err) { console.log('create hunt error', err, result); }
```
awick, line 8160: should it return or something here?

viewer/viewer.js
```diff
@@ -8068,6 +8068,45 @@ function updateHuntStatus (req, res, status, successText, errorText) {
...
+      }
+    };
+
+    Db.searchUsers(query, (error, users) => {
+      if (error) {
+        reject('Unable to validate userIds');
+      }
+
+      let usersList = [];
+      for (let user of users.hits.hits) {
```
awick, line 8088: map or filter or something youngin

viewer/viewer.js
```diff
@@ -8068,6 +8068,45 @@ function updateHuntStatus (req, res, status, successText, errorText) {
...
+    const query = {
+      _source: ['userId'],
+      from: 0,
+      size: 10000,
+      query: { // exclude the shared user from results
+        bool: { must_not: { term: { userId: '_moloch_shared' } } }
+      }
+    };
+
+    Db.searchUsers(query, (error, users) => {
```
awick, line 8082: maybe handle somewhere below, but does this handle the case when there are no users because they run anonymously? or is an anonymous user created anyway right away.

viewer/viewer.js
```diff
@@ -8068,6 +8068,45 @@ function updateHuntStatus (req, res, status, successText, errorText) {
...
+      if (error) {
+        reject('Unable to validate userIds');
+      }
+
+      let usersList = [];
+      for (let user of users.hits.hits) {
+        usersList.push(user._source.userId);
+      }
+
+      let validUsers = [];
```
awick, line 8092: I think youngins would use sets and filter here :)

viewer/viewer.js
```diff
@@ -8224,6 +8297,87 @@ app.put('/hunt/:id/play', [noCacheJson, logAction('hunt/:id/play'), checkCookieT
   updateHuntStatus(req, res, 'queued', 'Queued hunt item successfully', 'Error starting hunt job');
 });
 
+app.delete('/hunt/:id/users/:user', [noCacheJson, logAction('hunt/:id/users/:user'), checkCookieToken, checkPermissions(['packetSearch']), checkHuntAccess], (req, res) => {
+  if (Config.get('multiES', false)) { return res.molochError(401, 'Not supported in multies'); }
+
+  Db.get('hunts', 'hunt', req.params.id, (err, hit) => {
```
awick, line 8303: probably should create Db.getHunt since we have all the other hunt stuff there already

viewer/viewer.js
```diff
@@ -8224,6 +8297,87 @@ app.put('/hunt/:id/play', [noCacheJson, logAction('hunt/:id/play'), checkCookieT
...
+    });
+  });
+});
+
+app.post('/hunt/:id/users', [noCacheJson, logAction('hunt/:id/users/:user'), checkCookieToken, checkPermissions(['packetSearch']), checkHuntAccess], (req, res) => {
+  if (Config.get('multiES', false)) { return res.molochError(401, 'Not supported in multies'); }
+
+  if (!req.body.users) { return res.molochError(403, 'You must provide users in a comma separated string'); }
+
+  Db.get('hunts', 'hunt', req.params.id, (err, hit) => {
```
awick, line 8339: Db.getHunt

awick (commented):
Minor stuff, feel free to ignore. I did wonder why there are new apis instead of just a updateHunt call. Maybe that would be more work anyway.

31453:
There's no update hunt API endpoint, so I needed a new API endpoint anyway. I could have called `updateHuntStats` and changed the function to accommodate the new user stuff. But I thought it was easier and more readable to just add 2 new endpoints that only updated the users array and set the hunt.

viewer/viewer.js
31453: :roll_eyes:

viewer/viewer.js
31453: 👍

viewer/viewer.js
31453: 👍

viewer/viewer.js
31453: :roll_eyes:

viewer/viewer.js
31453: Probably! 🤓

PR #1530, opened 2020-10-15
add capture process start time indication to sessions timeline graph

### LINKED ISSUE #271: visual indicator in session view graph for capture process restarts
A capture process restart loses protocol state awareness and restarts segment counts.  To avoid bad assumptions a visual vertical bar identifying a specific capture process had restarted at that time would be great.

31453:
only displayed on sessions page
can also view start time and running time fields on the capture stats page

Screen Shot 2020-10-15 at 4 24 51 PM
Screen Shot 2020-10-15 at 4 25 07 PM

fixes #271

### DIFF
--- a/viewer/vueapp/src/components/sessions/Sessions.vue
+++ b/viewer/vueapp/src/components/sessions/Sessions.vue
@@ -34,11 +34,11 @@
 
     <!-- visualizations -->
     <moloch-visualizations
-      v-if="mapData && graphData && capStartTime"
+      v-if="mapData && graphData && capStartTimes.length"
       :graph-data="graphData"
       :map-data="mapData"
       :primary="true"
-      :cap-start-time="capStartTime"
+      :cap-start-times="capStartTimes"
       :timezone="user.settings.timezone"
       :timelineDataFilters="timelineDataFilters"
       @fetchMapData="cancelAndLoad(true)">
@@ -600,7 +600,7 @@ export default {
       stickyHeader: false,
       tableHeaderOverflow: undefined,
       showFitButton: false,
-      capStartTime: undefined
+      capStartTimes: []
     };
   },
   created: function () {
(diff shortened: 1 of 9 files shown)


viewer/vueapp/src/components/sessions/Sessions.vue
```diff
@@ -1451,6 +1454,16 @@ export default {
         this.loading = false;
       });
     },
+    getCaptureStats: function () {
+      this.$http.get('stats.json')
+        .then((response) => {
+          // TODO ECR - which capture process?
+          this.capStartTime = response.data.data[0].startTime * 1000;
```
31453, line 1461: @awick So, this endpoint returns an array of capture stats. I'm assuming that we should use the first one. Is that a valid assumption?

viewer/vueapp/src/components/sessions/Sessions.vue
```diff
@@ -34,10 +34,11 @@
 
     <!-- visualizations -->
     <moloch-visualizations
-      v-if="mapData && graphData"
+      v-if="mapData && graphData && capStartTime"
```
31453, line 37: @awick The capture start time is only visible on the sessions page. Should this be added to all pages? I think it makes the most sense on the sessions page given the use case provided in #271

viewer/vueapp/src/components/sessions/Sessions.vue
awick: no, you would want to show a little line for each capture instance since they could all be different

awick: or maybe only show the most recent or least recent? hmm

viewer/vueapp/src/components/sessions/Sessions.vue
awick: ya i think anywhere there is sessions graph

viewer/vueapp/src/components/sessions/Sessions.vue
31453: Ah ok, so there could be multiple restarts. One for each capture at least. So it should probably say in the tooltip: "X capture restarted at this time"

viewer/vueapp/src/components/sessions/Sessions.vue
31453: 👍

31453: It is now displayed on all session timeline graphs! It is only shown on the main timeline graph in spigraph

Screen Shot 2020-10-16 at 8 23 33 AM

awick (approved):
LGTM

PR #1532, opened 2020-10-21
add business hour indication to sessions graph

### LINKED ISSUE #419: visual highlighting of business/peak hours per node per graph
Far reaching wishful thinking to render graphs with highlighting/shading that identify configurable periods of interest like normal business hours. This gets more interesting with node in many different timezones and the need to represent a sliding window of business hours.

31453:
can be configured in config

fixes #419

Screen Shot 2020-10-22 at 12 06 12 PM


release/config.ini.sample
```diff
@@ -252,6 +252,13 @@ packetThreads=2
 # and includes subdomains (the app must be served over https)
 #hstsHeader=true
 
+# Business Hours
```
31453, line 255: @awick good place to put this?

release/config.ini.sample
```diff
@@ -252,6 +252,13 @@ packetThreads=2
...
+# If set, displays a colored background on the sessions timeline graph during business hours
+# Business hours are assumed to be on a daily basis Monday through Friday
```
31453, line 257: Should business days be configurable too?

release/config.ini.sample
31453: @awick and does it make sense?

release/config.ini.sample
```diff
@@ -252,6 +252,21 @@ packetThreads=2
...
+# Values are set in hours from midnight (default is off)
+#businessDayStart=9
+#businessDayEnd=17
+
+# Business Days
```
31453, line 262: @awick make sense?

release/config.ini.sample
```diff
@@ -252,6 +252,21 @@ packetThreads=2
...
+# Comma separated list of days
+# If set, displays the business hours on only the days provided here
+# Business hours must be set for these to be of use
+# Values are the days of the week as numbers, the week starts at Sunday = 0 and ends on Saturday = 6
+# (default is Monday - Friday 1,2,3,4,5)
```
31453, line 267: @awick good default?

viewer/viewer.js
```diff
@@ -9449,7 +9449,10 @@ app.use(cspHeader, setCookie, (req, res) => {
     huntWarn: Config.get('huntWarn', 100000),
     huntLimit: limit,
     serverNonce: res.locals.nonce,
-    anonymousMode: !!app.locals.noPasswordSecret && !Config.get('regressionTests', false)
+    anonymousMode: !!app.locals.noPasswordSecret && !Config.get('regressionTests', false),
```
BryceV, line 9452: This context is never used?

viewer/vueapp/src/components/visualizations/Visualizations.vue
```diff
@@ -769,6 +772,40 @@ export default {
           }
         });
       }
+
+      // add business hours to graph if they exist
+      if (!this.$constants.MOLOCH_BUSINESS_DAY_START || !this.$constants.MOLOCH_BUSINESS_DAY_END) {
+        return;
+      }
+      let businessDays = this.$constants.MOLOCH_BUSINESS_DAYS.split(',');
```
BryceV, line 780: Safer if this was wrapped in a `if (this.$constants.MOLOCH_BUSINESS_DAY_START && this.$constants.MOLOCH_BUSINESS_DAY_END)` just in case code is added after this piece.

viewer/viewer.js
31453: The anonymous mode part? It's not used in this particular PR, but it is used elsewhere. The diff here is just showing that I added a comma to the end to add more variables.

viewer/vueapp/src/components/visualizations/Visualizations.vue
31453: Yeah, not a bad idea 👍 Maybe I should also make this it's own function and just call it here. That might be cleaner

BryceV (approved):
Looks good! I would add these new fields to "ini settings" (settings.html) on molochWeb if we leave them as part of the config file

31453:
@BryceV Good idea on updating the ini settings! 🧠 💡 Thanks for reminding me!

viewer/viewer.js
BryceV: Oh, no just the new business fields. I see you are using them by passing it into this.$constants

release/config.ini.sample
awick: Assume this is UTC, should mention that.  It would be could be a range in 1 variable instead of 2. not sure if we are using moment or something else that could do magic.

release/config.ini.sample
awick: yep, the only other common weekends is fri/sat

awick (approved):

release/config.ini.sample
31453: Yes! It is. I'll add a note about that

PR #1534, opened 2020-10-28
Package stuff

31453:
Add all client bundling and scripts to the top level to consolidate all of the common npm packages!

Also update documentation


package.json
```diff
@@ -60,11 +60,33 @@
     "lint-fix": "eslint --fix --ext .js,.vue parliament viewer notifiers wiseService",
     "lint-fix-viewer": "eslint --fix --ext .js,.vue viewer",
     "lint-fix-parliament": "eslint --fix --ext .js,.vue parliament",
-    "lint-fix-wise": "eslint --fix --ext .js,.vue wiseService"
+    "lint-fix-wise": "eslint --fix --ext .js,.vue wiseService",
+    "viewer:addtestuser": "cd viewer && node addUser.js -c ../tests/config.test.ini -n testuser admin admin admin --admin --packetSearch",
+    "viewer:build": "cd viewer && node vueapp/build/build.js",
+    "viewer:bundle": "webpack -d --progress --config viewer/vueapp/build/webpack.dev.conf.js",
+    "viewer:start": "cd viewer && node viewer.js -c ../config.ini",
```
BryceV, line 67: Does a --prefix flag work here? Then you dont have to run cd everytime

package.json
```diff
@@ -60,11 +60,33 @@
...
+    "previewer:start": "npm run viewer:build",
```
BryceV, line 68: Is this a hook to run before viewer:start? It doesnt look like we built before this previously

parliament/package.json
```diff
@@ -6,47 +6,22 @@
     "type": "git",
     "url": "https://github.com/aol/moloch.git"
   },
-  "scripts": {
```
BryceV, line 9: Part of me thinks it's nice to have this for developing. If working from root dir, the git diff's will be several levels deep. But then we are supporting 2 scripts... We could also have the root README scripts just point to the child README scripts.

viewer/viewer.js
```diff
@@ -9721,7 +9721,7 @@ function processCronQueries () {
 // Main
 // ----------------------------------------------------------------------------
 function main () {
-  if (!fs.existsSync('./vueapp/dist/index.html')) {
+  if (!fs.existsSync('./vueapp/dist/index.html') && app.settings.env !== 'development') {
```
BryceV, line 9724: How does the env account to whether index.html exists?

CONTRIBUTING.md
```diff
@@ -17,29 +17,34 @@ First, checkout the main [Moloch README](README.md) for information on how to bu
...
-  * run `tests.pl` and `tests.pl --viewer` in the tests directory
+  * run `tests.pl` and `tests.pl --viewer` in the tests directory (this loads PCAPs)
 
 > **Note:** this will only work if viewer is not already running.
 
 You should now have test data loaded, so let's **start the web app**:
 
-* Move to the Moloch viewer directory
+* Move to the viewer directory
 * Run `npm ci`
```
awick, line 32: do you still need this in the viewer directory?

viewer/viewer.js
31453: If the app is run in development mode, webpack blows away the vueapp/dist directory, so the index file is not there and this error always happens. If it's run in production, the build is run before running the app and the index file is already there.

package.json
31453: I don't think prefix works with the node command. Or maybe I'm not understanding what you're saying

parliament/package.json
31453: I'm not sure what the best way to do this is. I definitely don't want to support 2 scripts for everything, so I put everything at the top level. @awick what do you think?

awick: I say we try it for a bit and if we can't stand it we can add some back that point to the top level like the lint one did.

viewer/viewer.js
awick, line 9724: this should use the same join dirname maybe?

awick (commented):
minor comments, haven't actually tried yet

package.json
BryceV: Huh, guess it doesnt

CONTRIBUTING.md
31453: Yeah, there are still a few node modules in the viewer directory. I opted to leave them here because they are specific to viewer (no overlap with any of the other projects). But, we could take this even further and put everything at the top level. I thought that was unnecessary since we need a package.json for each file anyway.

viewer/viewer.js
31453: 👍 good idea. done

PR #1540, opened 2020-11-15
docs: fix simple typo, quiting -> quitting

timgates42:
There is a small typo in capture/plugins/netflow.c, capture/plugins/suricata.c, capture/plugins/tagger.c.

Should read `quitting` rather than `quiting`.


awick:
Do you want to fix the ones in the code also?

PR #1546, opened 2020-11-20
Update/Standardize APIs

### LINKED ISSUE #1535: Add POST version of expression APIs
Add POST versions of any APIs that support expression so that long expressions work better.  Then switch the UI to use them. Would be best if could use something like app.all instead of having app.get and.post that might make things get out of sync, but not sure if possible.

31453:
- add jsdoc documentation and ability to generate documentation using jsdoc2md
- document api doc workflow in contributing
- add POST version of APIs that use expressions (support GET still)
- organize and breaks out some APIs and helper functions from `viewer.js` into their own files
- organize `viewer.js`

fixes #1535


31453:
Check out the documentation of viewer APIs by running `npm run viewerdoc` from the top level. The output is at `/arkime/viewer/jsdoc/arkimeAPI.md`. I'd like to use this file instead of `/arkimeweb/_wiki/api.md`. Ignore the "members" section, I would like to remove it entirely and replace it with something similar to the TOC we have for the current `/arkimeweb/_wiki/api.md file.

package.json
```diff
@@ -60,7 +60,8 @@
     "lint-fix": "eslint --fix --ext .js,.vue parliament viewer notifiers wiseService",
     "lint-fix-viewer": "eslint --fix --ext .js,.vue viewer",
     "lint-fix-parliament": "eslint --fix --ext .js,.vue parliament",
-    "lint-fix-wise": "eslint --fix --ext .js,.vue wiseService"
+    "lint-fix-wise": "eslint --fix --ext .js,.vue wiseService",
+    "viewerdoc": "jsdoc2md --partial viewer/jsdoc/body.hbs viewer/jsdoc/returns.hbs viewer/jsdoc/sig-name.hbs viewer/jsdoc/global-index.hbs --files viewer/apiSessions.js viewer/apiConnections.js viewer/viewer.js > ../arkimeweb/_wiki/api_docs.md"
```
BryceV, line 64: Assumes where arkimeweb is installed?

viewer/jsdoc/returns.hbs
```diff
@@ -0,0 +1,13 @@
+{{#if returns}}
+{{#if returns.[0].description~}}
```
BryceV, line 2: returns.[0] ?

BryceV (approved):
Havent pulled it but good idea to standardize!

viewer/jsdoc/returns.hbs
31453: Something specific with the hbs stuff. I don't know the nuances. The only reason I've added this file is to override the default. I wanted to add a newline to each different return value, which the default puts on a single line

package.json
31453: Yeah, this is just for testing to make sure this is a viable solution. If you and @awick like it, I'm going to add an argument to pass in the location of the arkimeweb directory

awick: can you leave the default to be ../arkimeweb since sane people will use that :)

31453: Yeah, I think so!

PR #1547, opened 2020-11-21
Add http s3 connection

lionelfleury:
Allow insecure connections to S3 when deploying in docker (or Kubernetes) with S3 on-prem (MinIO e.g.) mainly.


awick:
The feature is fine, but I don't like the name, everywhere we use insecure to mean https but don't check the cert.  Maybe just s3UseHttp

lionelfleury:
Fair renaming. done ✅

PR #1550, opened 2020-11-29
Fix incorrect path of package-lock file in wiseService install target

michaelroland:
The path `$(VIEWERDIR)/package-lock.json` seems to be a left-over from copy-pasting the Makefile. This should be `$(WISEDIR)/package-lock.json` to clean-up the package-lock file of this build.

## License


31453:
Thanks!

PR #1551, opened 2020-12-01
Yara version 4 support - code clone edition

rnbwdsh:
Add yara 4 support as requested in
https://arkime.slack.com/archives/CC8AR721W/p1606744757044000

Sticking to the design of the yara.c file of wrapping the whole class in #if

Just a shorter version of
#1552


awick:
thanks

PR #1552, opened 2020-12-01
Yara version 4 support - inline edition

rnbwdsh:
Add yara 4 support as requested in
https://arkime.slack.com/archives/CC8AR721W/p1606744757044000

Just adding the 2 changes in a #if into the yara 3 code.

Just a shorter version of
https://github.com/arkime/arkime/pull/1551


awick:
use your other solution

PR #1553, opened 2020-12-01
Check old yara path or yara path for archlinux

rnbwdsh:
It's just 1 line and shouldn't have side effects.


awick:
doesn't seem like this is needed

PR #1560, opened 2020-12-08
rename molochconfig.h to arkimeconfig.h, include prefix in config.h

awick:
see if we can build both a /opt/arkime and /data/moloch and have life work out


31453 (approved):
🍑

awick:
the beginning of the end

PR #1566, opened 2020-12-13
first version of esProxy

awick:
@31453 let me know if there is anything you would like to change from a js standpoint

### DIFF
--- a/viewer/esProxy.js
+++ b/viewer/esProxy.js
@@ -24,7 +24,7 @@ const http = require('http');
 const https = require('https');
 const fs = require('fs');
 const basicAuth = require('basic-auth');
-var URL = require('url');
+const URL = require('url');
 
 // express app
 const app = express();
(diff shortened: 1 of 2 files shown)


viewer/esProxy.js
```diff
@@ -0,0 +1,280 @@
...
+  });
+}
+
+// Proxy
+
+function doProxy (req, res, cb) {
+  let result = '';
+  let url = elasticsearch + req.url;
+  console.log('URL', url);
+  var info = URL.parse(url);
```
31453, line 162: Any reason you're using `var` in some places and `let` in others??

viewer/esProxy.js
```diff
@@ -0,0 +1,280 @@
...
+
+// ============================================================================
+// handle the requests from the capture/viewer
+// ===========================================================================
+
+// Get requests
+app.get('*', (req, res) => {
+  const path = req.params['0'];
+
+  if (getExact[path]) {
```
31453, line 209: What is this part all about? Why doesn't each case have anything inside of it? Is it a stub for the future?

31453: And why not SWITCH!? 😄

31453: Maybe create a function for the parts that are the same for the POST and GET parts?

31453 (approved):

viewer/esProxy.js
awick: copy paste, fixing

viewer/esProxy.js
awick: they are empty because they aren't returning an error and will do success at bottom. I'll add a comment

31453 (approved):

PR #1570, opened 2020-12-17
simplify encoding, initSimple now does load

awick:
* results are always encoded now with leading count, no more {num:, buffer:}
* result2Str => result2JSON
* encode => encodeResult
* getRaw/putRaw => getSourceRaw/putSourceRaw
* initSimple now handles the initial load and reloading


31453 (approved):
:rage1:

PR #1579, opened 2021-01-05
support text area for fields and views

awick:
Support text areas for wise config and auto split and join before saving.


31453 (approved):
🍑

31453:
fix that linting error though 😄

PR #1582, opened 2021-01-08
add wise ui startup time to the stats page

### LINKED ISSUE #1580: display startup timestamp on wise stats page
Display the startup on the stats page. Maybe on the same line as Sources/Types but on the far right. The /stats endpoint returns it as startupTime element. For now display as localtime, although I'm slowly adding user support so maybe eventually it could use the same setting as viewer.

image

31453:
put moment-timezone in the top level package.json

Screen Shot 2021-01-08 at 11 12 33 AM

fixes #1580


31453:
@awick fixed the text. Also, if you want to figure out why `moment-timezone` is needed in the viewer `package.json`, please be my guest. I can't figure it out for the life of me. It's impossibly frustrating 😄

PR #1583, opened 2021-01-08
Users api

31453:
Put all users api endpoint functions into apiUsers.js!

Document and standardize endpoints.


viewer/apiUsers.js
```diff
@@ -0,0 +1,1493 @@
+'use strict';
+
+const fs = require('fs');
+const stylus = require('stylus');
+
+module.exports = (app, Config, Db, internals, ViewerUtils) => {
```
awick, line 6: would like to remove app

viewer/apiUsers.js
```diff
@@ -0,0 +1,1493 @@
...
+
+    const clone = {};
+
+    for (const prop of userProps) {
+      if (req.user.hasOwnProperty(prop)) {
+        clone[prop] = req.user[prop];
+      }
+    }
+
+    clone.canUpload = app.locals.allowUploads;
```
awick, line 204: move to internals

viewer/vueapp/src/components/settings/Settings.vue
```diff
@@ -2499,11 +2499,11 @@ export default {
 
           if (updateTheme) {
             let now = Date.now();
-            if ($('link[href^="user.css"]').length) {
-              $('link[href^="user.css"]').remove();
+            if ($('link[href^="api/user/css"]').length) {
```
awick, line 2502: still hurts deeply that it isn't api/user.css, at least we support both

awick (changes requested):

viewer/vueapp/src/components/settings/Settings.vue
31453: 💔

viewer/apiUsers.js
31453: 👍

viewer/apiUsers.js
31453: 👍

PR #1591, opened 2021-01-15
update entire shortcut (not just shortcut value)

31453:


31453:
@awick give it a quick test for me please ❤️

PR #1598, opened 2021-01-28
Wise cleanup

### LINKED ISSUE #1592: WISE UI Cleanup
- [x] On query page if you switch types and have never typed anything it gives an error about empty search box, it shouldn't
- [x] would be nice if it remembered which query type you last had selected and/or maybe default to ip (if not in url)
- [x] hitting enter in query box if there isn't currently a query going should resubmit (sometimes I know something has changed, or there has been an error, and no easy way to try again)

31453:
- don't show error message on wise query page if search never entered
- save wise query type and source in localstorage
- submit wise query on enter click from search input

fixes #1592

### DIFF
--- a/wiseService/vueapp/src/components/Query.vue
+++ b/wiseService/vueapp/src/components/Query.vue
@@ -216,10 +216,12 @@ export default {
     },
     sendSearchQuery: function () {
       if (!this.searchTerm) {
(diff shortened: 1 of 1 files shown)


awick:
lgtmuii

PR #1610, opened 2021-02-22
Packet Options Improvements

31453:
Combine packet options into a dropdown instead of a list of buttons
Make packet option buttons its own component for reuse


awick:
LGTM

PR #1612, opened 2021-02-24
Webpack 5 Upgrade

31453:
✨ AMAZING STUFF 🥳 🎉 

- Upgrade lots of packages 🆙 
- Move all client packages to devDeps 🚗 
- More strict linting and linting fixes. const time 🤩
- Remove redundant files and use toplevel gitignore, babelrc, postcss configs ❌ 
- Remove vueapp from nodemon watch so client changes don't restart the server 😸 
- Source maps work again, HOORAY! 🗺️ 

✨


parliament/vueapp/build/webpack.prod.conf.js
```diff
@@ -35,58 +36,38 @@ const webpackConfig = merge(baseWebpackConfig, {
...
+    new webpack.ProvidePlugin({
+      $: 'jquery',
+      jQuery: 'jquery'
     }),
     // generate dist index.html with correct asset hash for caching.
     // you can customize output by editing /index.html
     // see https://github.com/ampedandwired/html-webpack-plugin
     new HtmlWebpackPlugin({
-      title: 'Production Parliament',
+      title: 'Production Arkime',
```
awick, line 59: Why isn't this Parliament?

parliament/vueapp/build/webpack.prod.conf.js
```diff
@@ -35,58 +36,38 @@ const webpackConfig = merge(baseWebpackConfig, {
...
       filename: config.build.index,
       template: 'index.html',
       inject: true,
       minify: {
-        removeComments: true,
+        removeComments: false,
```
awick, line 64: bloat? :)

parliament/vueapp/config/index.js
```diff
@@ -1,44 +1,40 @@
...
+    assetsRoot: path.resolve(__dirname, '../dist'),
     assetsSubDirectory: 'static',
-    assetsPublicPath: '/',
+    assetsPublicPath: '',
     proxyTable: {},
 
     // Various Dev Server settings
     host: 'localhost', // can be overwritten by process.env.HOST
-    port: 8080, // can be overwritten by process.env.PORT, if port is in use, a free one will be determined
+    port: 8123, // can be overwritten by process.env.PORT, if port is in use, a free one will be determined
```
awick, line 21: is this the parliament port? Then the default should be 8008 right? Or this is just for testing?

viewer/config.js
```diff
@@ -98,19 +98,19 @@ exports.md5 = function (str, encoding) {
...
+    const iv = crypto.randomBytes(16);
+    const c = crypto.createCipheriv('aes-256-cbc', internals.passwordSecret256, iv);
     let e = c.update(m, 'binary', 'hex');
     e += c.final('hex');
     return iv.toString('hex') + '.' + e;
   } else {
     // Old style without IV: E
-    var c = crypto.createCipher('aes192', internals.passwordSecret);
-    var e = c.update(m, 'binary', 'hex');
+    const c = crypto.createCipheriv('aes192', internals.passwordSecret);
```
awick, line 112: this can NOT change

viewer/config.js
```diff
@@ -119,17 +119,17 @@ exports.pass2store = function (userid, password) {
...
-      let c = crypto.createDecipheriv('aes-256-cbc', internals.passwordSecret256, Buffer.from(parts[0], 'hex'));
+      const c = crypto.createDecipheriv('aes-256-cbc', internals.passwordSecret256, Buffer.from(parts[0], 'hex'));
       let d = c.update(parts[1], 'hex', 'binary');
       d += c.final('binary');
       return d;
     } else {
       // Old style without IV: E
-      var c = crypto.createDecipher('aes192', internals.passwordSecret);
-      var d = c.update(passstore, 'hex', 'binary');
+      const c = crypto.createDecipheriv('aes192', internals.passwordSecret);
```
awick, line 131: this can NOT change

viewer/config.js
```diff
@@ -149,22 +149,22 @@ exports.obj2auth = function (obj, c2s, secret) {
...
     e = iv.toString('hex') + '.' + e;
-    let h = crypto.createHmac('sha256', secret).update(e).digest('hex');
+    const h = crypto.createHmac('sha256', secret).update(e).digest('hex');
     return e + '.' + h;
   } else {
     // Old style without IV: E or E.H
     secret = secret || internals.serverSecret;
 
-    let c = crypto.createCipher('aes192', secret);
+    const c = crypto.createCipheriv('aes192', secret);
```
awick, line 163: can NOT change

viewer/config.js
```diff
@@ -214,16 +214,16 @@ exports.auth2obj = function (auth, c2s, secret) {
...
+      const h = crypto.createHmac('sha256', secret).update(parts[0], 'hex').digest();
 
       if (!crypto.timingSafeEqual(signature, h)) {
         throw new Error('Incorrect signature');
       }
     }
 
     try {
-      let c = crypto.createDecipher('aes192', secret);
+      const c = crypto.createDecipheriv('aes192', secret);
```
awick, line 226: Can NOT change

viewer/decryptPcap.js
```diff
@@ -38,30 +38,30 @@ function main () {
...
+    const kek = Config.sectionGet('keks', info.kekId, undefined);
     if (kek === undefined) {
       console.error("ERROR - Couldn't find kek", info.kekId, 'in keks section');
       process.exit();
     }
 
     // Decrypt the dek
-    let kdecipher = crypto.createDecipher('aes-192-cbc', kek);
-    let encKey = Buffer.concat([kdecipher.update(Buffer.from(info.dek, 'hex')), kdecipher.final()]);
+    const kdecipher = crypto.createDecipheriv('aes-192-cbc', kek);
```
awick, line 55: Can't change

viewer/multies.js
```diff
@@ -20,44 +20,45 @@
...
-  var Config = require('./config.js');
-  var express = require('express');
-  var async = require('async');
-  var URL = require('url');
-  var ESC = require('elasticsearch');
-  var http = require('http');
-  var https = require('https');
-  var fs = require('fs');
-  var path = require('path');
+  Config = require('./config.js');
```
awick, line 26: This block seems wrong? What was wrong with var.

wiseService/wiseService.js
```diff
@@ -219,13 +220,13 @@ function store2ha1 (passstore) {
...
       // New style with IV: IV.E
-      let c = crypto.createDecipheriv('aes-256-cbc', internals.passwordSecret256, Buffer.from(parts[0], 'hex'));
+      const c = crypto.createDecipheriv('aes-256-cbc', internals.passwordSecret256, Buffer.from(parts[0], 'hex'));
       let d = c.update(parts[1], 'hex', 'binary');
       d += c.final('binary');
       return d;
     } else {
       // Old style without IV: E
-      const c = crypto.createDecipher('aes192', internals.passwordSecret);
+      const c = crypto.createDecipheriv('aes192', internals.passwordSecret);
```
awick, line 229: Don't change

awick (changes requested):
* need to changed the crypto IV stuff back
* few other things
* maybe next time don't do a fix with other changes :) and fix might be dangerous until we stop supporting old crypto stuff

viewer/multies.js
31453: What's wrong with this block? `var` is not useful in most situations and harmful in many, especially with large code bases where scope is very important as there may be many variables with the same name. 

Here, it kind of makes sense. So I can add an inline eslint exclusion. Going forward, I'd like to require that `var` not be used (unless completely necessary).

wiseService/source.opendns.js
```diff
@@ -32,7 +32,7 @@ class OpenDNSSource extends WISESource {
 
     this.waiting = [];
     this.processing = {};
-    this.statuses = { '-1': 'malicious', '0': 'unknown', '1': 'benign' };
+    this.statuses = { '-1': 'malicious', 0: 'unknown', 1: 'benign' };
```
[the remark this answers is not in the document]
31453: You don't need to quote properties of an object, even if they're a number.

viewer/multies.js
awick: I was "taught" that every variable should have var/let/const in front of it. If you don't have var then is it stuck inside the try block?  btw multies --viewer tests are failing

wiseService/source.opendns.js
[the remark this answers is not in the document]
awick: sweet, just wanted to make sure a['0'] and a[0] were the same thing :)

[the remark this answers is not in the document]
31453: Unless they have special characters, of course, like `-`

viewer/multies.js
31453: It's not stuck inside the try block since they are all initialized by `let` above. 

This is a matter of syntax preference, they both behave the same in the end:
- With `var`, the variable is hoisted to top scope. 
- With `let`, it is initialized in the top scope and defined in the try block. 

I'm happy either way since they are effectively the same solution. The `let` syntax is simply what eslint favors.

However, I do not want to disable requiring `const` and `let` globally. It is my opinion that `var` should be used sparingly, if at all. This is an uncommon case where I think it can be utilized effectively and understandably without introducing bugs. So which of these do you prefer:
1. keep it like this
2. go back to `var` but place `eslint-disable-next-line no-var` before every instance of `var`

Again, I'm happy with either solution.

31453: Working on fixing tests now! 🔨

awick: lets just remove the try and use const. People shouldn't be building themselves anymore anyway plus we've improved things

31453: 🥳 fancy!

viewer/apiSessions.js
```diff
@@ -697,7 +694,7 @@ module.exports = (Config, Db, internals, molochparser, Pcap, version, ViewerUtil
         ViewerUtils.getViewUrl(fields.node, (err, viewUrl, client) => {
           let buffer = Buffer.alloc(Math.min(16200000, fields.totPackets * 20 + fields.totBytes));
           let bufpos = 0;
-          const info = url.parse(viewUrl);
+          const info = new URL(viewUrl);
```
31453, line 697: @awick Do you know if this might cause any issues? Looking at the [documentation](https://nodejs.org/api/url.html#url_constructor_new_url_input_base) it looks like `new URL` behaves the same as `url.parse`. But you might know something I don't

awick: Ya I looked at that and think it will work, but haven't tested it :)

PR #1620, opened 2021-03-09
remove clean webpack plugin

31453:
use webpack native clean output flag for prod and dev builds
put hot reload files into subfolder in dist and replace them instead of adding more every time a client file is saved


awick (approved):
LGTM

PR #1656, opened 2021-03-26
Fix theming of native select elements (currently used in SessionDetail

rnbwdsh:


31453:
Looks good. Nice find. Thank you for your contribution! To be eligible for swag, please [sign-up](https://yahoo.github.io/hacktogether/) for Yahoo! Hack Together, if you have not done so, yet.

PR #1657, opened 2021-03-26
Also collapse graph

rnbwdsh:


31453:
I like this idea 👍 Can you update the toggle toolbars button tooltip to make it obvious that the the graph/map area is going to be collapsed too? Maybe "(un)hide toolbars and visualizations"? @awick thoughts?

Also, I marked this PR and issue with the HackTogether label. If you want swag, please [sign-up](https://yahoo.github.io/hacktogether/) for Yahoo! Hack Together, if you have not done so, yet.

PR #1658, opened 2021-03-26
View name typehead

rnbwdsh:


awick (changes requested):
Looks like there is a linting error

rnbwdsh:
My shorter-code fix actually was bugged (wrong copy from console-testing). Should be working now.

Screenshot btw:
view_typeahead

31453:
This is great! I forgot you can apply views this way ✨

PR #1659, opened 2021-03-26
Use the timezone inside the component instead of passing it 5x

rnbwdsh:


31453:
@rnbwdsh Thanks! This is a good idea. The MolochTime component does indeed have a getter for the user object which includes their timezone setting. The same goes for the MolochSearch component.

However, the MolochTime and MolochSearch components also need to be updated to exclude the `timezone` prop and use `user.settings.timezone` instead.

Also, I'm going to mark this PR and issue with the HackTogether label. If you want swag, please [sign-up](https://yahoo.github.io/hacktogether/) for Yahoo! Hack Together, if you have not done so, yet.

31453 (changes requested):
MolochTime and MolochSearch components also need to be updated to exclude the `timezone` prop and use `user.settings.timezone` instead.

rnbwdsh:
> Good idea, I'll fix that too.
In SessionField, it uses
`this.timezone || this.$store.state.user.settings.timezone,`
Whereas the passed in timezone is from the user.settings anyways, so the whole `this.timezone ||` can be omitted as well as the prop.

In fact I could edit it out in most places without any side-effects.

31453:
That's a great idea!

31453:
Also, it looks like the user object from the store is not available at the time that the `datePickerOptions` is being initialized. So, you'll have to wait to set `datePickerOptions` and initialize the time fields until the user object is available.

31453:
And FYI: it looks like there's an existing bug when changing the user timezone settings. It always sets the timezone to the previously selected setting. I'm created a new issue to reference this: #1667 No need to fix it in this PR unless you want to. I just wanted to make you aware of this issue.

viewer/vueapp/src/components/sessions/StickySessions.vue
```diff
@@ -176,6 +176,9 @@ export default {
       } else {
         return this.sessions;
       }
+    },
+    timezone: function () {
+      return this.$store.state.user.timezone;
```
31453, line 181: needs to be `this.$store.state.user.settings.timezone`

viewer/vueapp/src/components/visualizations/Visualizations.vue
```diff
@@ -370,6 +366,9 @@ export default {
           this.$store.commit('updateSeriesType', newValue);
         }
       }
+    },
+    timezone: function () {
+      return this.$store.state.user.timezone;
```
31453, line 371: needs to be `this.$store.state.user.settings.timezone`

31453 (changes requested):
just a few minor changes with `this.$store.state.user.settings.timezone`

rnbwdsh:
Rebased + Fixed by now, thanks for the great support @31453.

31453 (approved):

31453:
Looks great! Thanks for all the hard work ✨

PR #1677, opened 2021-04-26
Fix sessions col resize

31453:
improve column resize functionality by removing colresize lib
allow table to overflow the window width and scroll horizontally
columns take default width when setting back to default columns
session detail always inherits the width of the table
make fit button smaller and not obscured when navbars are collapsed
fix top margin issues for sticky header (with and without navs)
remove margin from right of table
scroll header and body of table when sticky header is activated
don't display scroll bar on header


awick (approved):
LGTM

PR #1679, opened 2021-04-28
wise ui improvements

31453:
decrease size of inputs
darken inputs in dark mode
fix table dark styles in dark mode
darken tabs on stats page in dark mode
darken 404 page in dark mode
don't show error if no search term in query, just show no results
fix forever loading with weird search terms
put alert message area below search
issue search when changing source dropdown
implement search when forward/back browser button is clicked
fix searching for urls (encode "/")


awick (approved):
👍

PR #1690, opened 2021-05-06
add "csv editor" for wise source configs

31453:
user can use csv editor or textarea
also remove interval for loading stats when leaving the stats page


wiseService/vueapp/src/components/Config.vue
```diff
@@ -154,21 +162,107 @@
...
-          <!-- text area input for non json -->
-          <b-form-textarea
-            v-if="!currJSONFile"
-            v-model="currFile"
-            rows="18"
-          />
+          <!-- text area input for non json/csv -->
+          <template
+            v-if="(!currJSONFile && currConfig[selectedSourceKey].format === 'tagger') || rawCSV">
+            <b-form-textarea
```
31453, line 168: If the format is CSV and the user is editing in the text area, should I clarify the CSV format? Maybe say something like:

> Rows are delimited by newlines. Cells are delimited by commas. Comments are delimited by # and should be an entire row by itself.

wiseService/vueapp/src/components/Config.vue
```diff
@@ -717,6 +848,71 @@ export default {
...
+        this.currCSV.rows.push(['']);
+        return;
+      }
+
+      const rows = this.currFile.split('\n');
+
+      for (const row of rows) { // parse rows
+        if (!row.length) { continue; } // emtpy row
+        if (row.startsWith('#')) { // comment row
+          // TODO disallow adding more fields to comment rows?
```
31453, line 868: @awick should we do this? or let users be stupid?

awick (changes requested):

PR #1696, opened 2021-05-17
Ecs3

awick:
ECS support and other 3.0 stuffs

* Index prefix by default is now "arkime_" but can be changed for multi tenancy
* New sessions indices will be {prefix}sessions3_* and will have some ECS fields (TBD)
* All other indices will now be {prefix}{index}_v30, if prefix wasn't used previously the new indices will still be arkime_ or whatever set
* Now setup field aliases for ECS fields on sessions2_* indices
* All sessions ES queries are moving to new names and using the fields: selector instead of _source https://www.elastic.co/guide/en/elasticsearch/reference/current/search-fields.html#search-fields-param which means everything is flat and in an array, except we unflat everything and move some array stuff into non array for now
* getSessionBySearch option is no longer supported (always true now)

TODO:

- [x] Tons of testing
- [x] Run tests on sessions2 and then upgrade to sessions3 and have all tests pass
- [x] Switch to fields from _source everywhere (_source: false)
- [x] capture test suite updates
- [x] columns not right yet
- [x] multies changes
- [x] SegmentSelect component error 
- [x] Sticky sessions component doesn't render
- [x] Open field in spigraph doesn't work (always shows default - node)
- [x] export unique ip:port doesn't work (does work if you use "srcPort" instead of "source.port")
- [x] connections popups don't show up (can't find field to provide friendlyname so it all breaks)
- [x] bunch of .attr of undefined errors on the connections page (that don't go away when navigating from the page)
- [x] settings page "Sort Sessions By" doesn't have a list of fields and the selected one is not selected in the select input
- [x] Custom SPI View Configurations on settings page do not display list of fields for each config in the table
- [x] default col config doesn't display fields in settings page
- [x] applying a view's col config doesn't work (sometimes)
- [x] adding/removing tags directly in session detail doesn't work (maybe it's just a really long delay? sometimes it does work)
- [x] periodic queries never make a match and log `No sessions to add tags (tag) to` over and over
- [x] scrubbing PCAP doesn't work (but deleting SPI does)
- [x] sessions with certs don't display the cert SPI data (it just says "Certificate" over and over)
- [x] ip.protocol == 0 sessions crash viewer when opening session details (needs to always have a source and destination object)
- [x] postgresql section shows array instead of value
- [x] socks area messed if if values are missing
- [x] can't view packets for this session: socks5-http-302-frag.pcap
- [ ] the string field in the test section displays an object
- [x] fix all/most include_type_name stuf
- [x] db.pl backup/restore of old/new backups with/without types works

Issues on Lab - sessions 2:
- [x] opening a session using the "Link" button in session detail doesn't work. it seems to be appending 2@ to the id.
- [x] exporting pcap pends forever
- [x] downloading src/dst bytes image pends forever FIXED NEEDS TESTING
- [x] tagging/removing tags is slow (if it works at all)
- [x] spiview fields more button doesn't show more field values FIXED NEEDS TESTING
- [x] there's always (0) in the count for spigraph field values when graph type is bytes or packets FIXED NEEDS TESTING
- [x] tooltips on timeline graphs show x out of 0 filtered when graph type is bytes or packets FIXED NEEDS TESTING
- [x] connections page doesn't have src/dst field selectors FIXED NEEDS TESTING
- [x] error when hovering over connections link: "Error in render: "TypeError: Cannot read property 'children' of undefined" looks like the field isn't being passed to SessionField because totDataBytes is missing
- [x] Hunt jobs don't run
Issues on Lab - sessions 3:
- [x] connections does exists query on old field names name srcIP totalBytes
- [x] export unique dst ip with port doesn't return any results (api/unique?exp=ip.dst%3AdstPort&counts=1&date=1)

TESTS from 2 (or main) -> ECS3:
You need both a directory for 2 and 3 checked out and built
<pre>
#!/bin/sh
curl -XDELETE http://localhost:9200/_template/*
curl -XDELETE http://localhost:9200/*
curl -XDELETE http://localhost:9200/_all/_alias/_all
curl http://localhost:9200/_refresh
sleep 1

cd ~/arkime.main/tests
./tests.pl --debug --viewer
curl http://localhost:9200/_refresh
sleep 1

cd ../../arkime.ecs/tests
curl http://localhost:9200/_cat/aliases
../db/db.pl --prefix tests http://localhost:9200 upgradenoprompt
../db/db.pl --prefix tests2 http://localhost:9200 upgradenoprompt
curl http://localhost:9200/_cat/aliases
curl http://localhost:9200/_cat/templates
./tests.pl --debug --viewerstart
</pre>

To go back to 2 (or main)
<pre>
curl -XDELETE http://localhost:9200/_template/*
curl -XDELETE http://localhost:9200/*
curl -XDELETE http://localhost:9200/_all/_alias/_all
./tests.pl --debug --viewer
</pre>


31453 (approved):
👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍

PR #1702, opened 2021-05-26
monitor cert and key files

### LINKED ISSUE #1698: monitor keyfile/certfile and if change reload or restart
For viewer.js/esproxy.js monitor the certfile and keyfile and if they change either reload or restart.  Things to watch out for/test

* viewer does a drop priv, so if it can't monitor it make sure you don't just restart over and over
* not sure its even possible to reload, might have to restart
* if files don't exist don't monitor

31453:
for multies/viewer/esproxy
reload the certs if they have changed and the files still exist

fixes #1698

### DIFF
--- a/viewer/config.js
+++ b/viewer/config.js
@@ -503,38 +503,38 @@ exports.configMap = function (section, dSection, d) {
 function loadCertData () {
   exports.keyFileLocation = exports.get('keyFile');
   exports.certFileLocation = exports.get('certFile');
-
-  // eslint-disable-next-line no-useless-catch
-  try {
-    exports.keyFileData = fs.readFileSync(exports.keyFileLocation);
-    exports.certFileData = fs.readFileSync(exports.certFileLocation);
-  } catch (err) {
-    throw err;
-  }
+  exports.keyFileData = fs.readFileSync(exports.keyFileLocation);
+  exports.certFileData = fs.readFileSync(exports.certFileLocation);
 }
 
+let certDataLoaded = false;
 if (exports.isHTTPS()) {
   try {
     loadCertData();
+    certDataLoaded = true;
   } catch (err) {
-    console.log('ERROR loading cert data:', err.toString());
+    console.log('ERROR loading cert or key files:', err.toString());
   }
 }
 
 exports.setServerToReloadCerts = function (server, cryptoOption) {
   if (!exports.isHTTPS()) { return; } // only used in https mode
 
-  try { // try to get the cert files
-    loadCertData();
-  } catch (err) { // if they don't exist don't try to monitor them
+  if (!certDataLoaded) { // cert or key files weren't loaded in the isHTTPS if
+    // so don't try to watch the files. they are likely not accessible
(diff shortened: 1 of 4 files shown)


viewer/config.js
```diff
@@ -500,10 +500,65 @@ exports.configMap = function (section, dSection, d) {
...
+  } catch (err) {
+    throw err;
+  }
+}
+
 if (exports.isHTTPS()) {
-  exports.keyFileData = fs.readFileSync(exports.get('keyFile'));
-  exports.certFileData = fs.readFileSync(exports.get('certFile'));
+  try {
+    loadCertData();
```
31453, line 518: @awick we don't need this anymore if we can assume that setServerToReloadCerts is called for each server that sets certs. This is duplicated on line 528.

viewer/config.js
```diff
@@ -500,10 +500,65 @@ exports.configMap = function (section, dSection, d) {
...
+function loadCertData () {
+  exports.keyFileLocation = exports.get('keyFile');
+  exports.certFileLocation = exports.get('certFile');
+
+  // eslint-disable-next-line no-useless-catch
+  try {
+    exports.keyFileData = fs.readFileSync(exports.keyFileLocation);
+    exports.certFileData = fs.readFileSync(exports.certFileLocation);
+  } catch (err) {
+    throw err;
```
awick, line 512: why do you need this try/catch if you are just going to throw? Actually we need to test this where viewer can't read the cert file after the drop priv and make sure its not constantly erroring.

viewer/config.js
awick: ok

awick (approved):

viewer/config.js
31453: I was using it to log consistent errors in the catch block, so I left it until we decided whether the we can assume that setServerToReloadCerts is called for each server that sets certs. Since you ok'd that assumption, I am going to remove the try catch and do all console logging in that function.

31453: Isn't any change to the cert files after the config loads testing the drop priv stuff?

awick: so if you do the monitor BEFORE the drop priv it might still fire after a change, but you might not be able to read the new file because it is owned by root and read only by root. In that case it might be better to restart. Not sure. I'm also not sure if file watching still works after drop priv, but you definitely wouldn't be able to open that file

The current test stuff doesn't test the drop priv, you would need to add
dropUser=nobody
dropGroup=nobody
make sure the test cert is own by root, chmod 600, and then start viewer as root. then try changing the file.
And then mac may be different than linux :)

31453: Also, in the end, I didn't like the assumption that setServerToReloadCerts would be called for every https server. I'm going a different route that we can discuss when I push the code.

awick: and to make sure I was clear, if there is no error it would be great to reset the cert without restarting. not everyone will have it so the cert isn't readable by nobody.

Issue #1423, opened 2020-04-14
support shortcuts across multiple clusters


awick:
Right now shortcuts have to be created on each cluster because of how ES works.  For people running multiple clusters it would be nice to have a central location to create/edit/view the shortcuts and have the updated in all the different clusters.  It would be nice to have a method to edit once and sync everywhere.  

Possible solutions, not sure the best.
* Make this a parliament feature.  Parliament keeps a db on short cuts and syncs to all clusters it knows about.
* Make this a wise feature.  Requires a wise UI.  Have viewer query wise occasionally and update elasticsearch.  maybe only cron query nodes.
* Make this a viewer feature, which ever viewer you edit on updates all the rest.

31453:
fixed in #1706

PR #1706, opened 2021-06-03
Cross-cluster Shortcuts!

31453:
If usersElasticsearch is set, save the shortcuts to the usersElasticsearch so they are available across all clusters

initial sync of shortcuts from local to remote db
update all npm script to include multies
add usersES and prefix to [all] in config
use _meta save version of shortcuts to determine whether to update local
add numberOfShortcuts instead of using numberOfDocuments('lookups')
compare local vs remote shortcuts and determine if any were deleted
compare remote vs local shortcuts to determine if any were updated/created
don't do shortcuts sync for multies

fixed in #1423


viewer/db.js
```diff
@@ -118,6 +121,18 @@ exports.initialize = async (info, cb) => {
     setInterval(() => { exports.getAliasesCache('sessions2-*'); }, 2 * 60 * 1000);
   }
 
+  // if there's a user's db, sync shortcuts from user's to local db so they can be
+  // used for sessions search (can't use remote db for searching via shortcuts)
+  internals.localShortcutsIndex = fixIndex('lookups');
+  if (internals.info.usersHost) {
```
awick, line 127: Only do this stuff if its a cron query viewer node AND usersElasticserach is set

viewer/db.js
```diff
@@ -987,31 +1002,208 @@ exports.getHunt = async (id) => {
...
+    if (!localResults.hits.hits.length) { return; } // nothing to sync to remote db
+
+    // add each of the local shortcuts to the remote shortcuts db (remote db = user's es)
+    // but first, compare the local shortcuts to the remote shortcuts to determine
+    // if any shortcuts have the same name in the remote db (we can assume that
+    // none of the local shortcuts will be the same as the remote db shortcuts
+    // since they have not been synced before)
+    for (const localShortcut of localResults.hits.hits) {
+      for (const remoteShortcut of remoteResults.hits.hits) {
+        if (remoteShortcut._source.name === localShortcut._source.name) {
```
awick, line 1044: Would be cool if you could compare the short cut and only do this if they are different values. If the same values keep the remote and delete the local one and wait for remote to be copied back down?

viewer/db.js
```diff
@@ -987,31 +1002,208 @@ exports.getHunt = async (id) => {
...
+  if (internals.multiES) { return; } // don't sync shortcuts for multies
+
+  const msg = `updating local shortcuts index (${internals.localShortcutsIndex}) from remote index (${internals.remoteShortcutsIndex})`;
+
+  try {
+    // fetch the version of the remote shortcuts index to check if the local shortcuts index
+    // is up to date. if not, something has changed in the remote index and we need to sync
+    const version = await getShortcutsVersion();
+
+    if (version !== internals.localShortcutsVersion) {
```
awick, line 1105: make this a === followed by a return and remove extra indent everywhere

awick (changes requested):

viewer/db.js
```diff
@@ -118,6 +121,18 @@ exports.initialize = async (info, cb) => {
     setInterval(() => { exports.getAliasesCache('sessions2-*'); }, 2 * 60 * 1000);
   }
 
+  // if there's a user's db, sync shortcuts from user's to local db so they can be
+  // used for sessions search (can't use remote db for searching via shortcuts)
+  internals.localShortcutsIndex = fixIndex('lookups');
+  if (internals.info.usersHost) {
+    internals.remoteShortcutsIndex = `${internals.usersPrefix}lookups`;
+    initialShortcutsSync(); // determine if shorcuts have been synced
+    updateLocalShortcuts(); // immediately udpate shortcuts
```
awick, line 130: don't these 2 until initial finishes

viewer/db.js
```diff
@@ -987,31 +1003,262 @@ exports.getHunt = async (id) => {
...
+    }
+
+    internals.localShortcutsVersion = version;
+  } catch (err) {
+    console.log(`ERROR - ${msg}:`, err);
+  }
+}
 exports.searchShortcuts = async (query) => {
+  return internals.usersClient7.search({
+    index: internals.remoteShortcutsIndex, body: query, rest_total_hits_as_int: true, version: true
```
awick, line 1226: need size:10000 here, otherwise default is 10 results

viewer/db.js
```diff
@@ -987,31 +1003,262 @@ exports.getHunt = async (id) => {
...
+  });
+};
+exports.searchShortcutsLocal = async (query) => {
   return internals.client7.search({
-    index: fixIndex('lookups'), body: query, rest_total_hits_as_int: true
+    index: internals.localShortcutsIndex, body: query, rest_total_hits_as_int: true, version: true
```
awick, line 1231: need size:10000 here, otherwise default is 10 results

viewer/db.js
31453: I'm not sending this function the query with `size: 10000`. I didn't want to break the client. It sends a size for pagination.

viewer/db.js
31453: I'm not sending this function the query with `size: 10000`. I didn't want to break the client. It sends a size for pagination.

viewer/db.js
```diff
@@ -987,31 +1003,273 @@ exports.getHunt = async (id) => {
...
+    ]);
+
+    if (!localResults.hits.hits.length) { return; } // nothing to sync to remote db
+
+    // add each of the local shortcuts to the remote shortcuts db (remote db = user's es)
+    // but first, compare the local shortcuts to the remote shortcuts to determine
+    // if any shortcuts have the same name in the remote db (we can assume that
+    // none of the local shortcuts will be the same as the remote db shortcuts
+    // since they have not been synced before)
+    for (const localShortcut of localResults.hits.hits) {
```
31453, line 1058: @awick mind checking my logic here? 🧠 💥

viewer/db.js
```diff
@@ -987,31 +1003,273 @@ exports.getHunt = async (id) => {
...
+async function initialShortcutsSyncToRemote () {
+  if (internals.multiES) { return; } // don't sync shortcuts for multies
+
+  const { body: doc } = await internals.usersClient7.indices.getMapping({
+    index: internals.remoteShortcutsIndex
+  });
+
+  let initSync;
+  // eslint-disable-next-line no-unreachable-loop
+  for (const index in doc) {
```
awick, line 1019: The millennial way might be Object.keys(doc)[0]

31453: lol, what are you saying? That I'm OLD!? 👵

viewer/db.js
```diff
@@ -987,31 +1003,273 @@ exports.getHunt = async (id) => {
...
+      for (const remoteShortcut of remoteResults.hits.hits) {
+        if (remoteShortcut._source.name === localShortcut._source.name) {
+          // found a local shortcut with the same name as the remote
+          let diff = false; // compare the shortcuts values
+          if (localShortcut._source.shared !== remoteShortcut._source.shared) { diff = true; }
+          if (localShortcut._source.userId !== remoteShortcut._source.userId) { diff = true; }
+          if (localShortcut._source.description !== remoteShortcut._source.description) { diff = true; }
+          if (!compareArrays(localShortcut._source.ip, remoteShortcut._source.ip)) { diff = true; }
+          if (!compareArrays(localShortcut._source.string, remoteShortcut._source.string)) { diff = true; }
+          if (!compareArrays(localShortcut._source.number, remoteShortcut._source.number)) { diff = true; }
```
awick, line 1069: style and slightly more efficient (stop after first diff) woudl be to do
let diff = A ||
               B ||
               C ||

although lint might get mad.

viewer/db.js
awick: LGTM

viewer/db.js
awick: once you have :kids:

31453: 😭

PR #1715, opened 2021-06-18
update _meta version for shortcut index when using set-shortcut db.pl script

31453:


db/db.pl
```diff
@@ -3508,6 +3508,17 @@ sub parseArgs {
...
+    # increment the _meta version by 1
+    my $mapping = esGet("/${PREFIX}lookups/_mapping");
+    my @indexes = keys %{$mapping};
+    my $index = @indexes[0];
+    my $meta = $mapping->{$index}->{mappings}->{_meta};
+
+    my $version = $meta->{version};
+    $version++;
+
+    esPut("/${PREFIX}lookups/_mapping", "{\"_meta\":{\"version\": $version, \"initSync\":" . ($meta->{initSync} ? "true" : "false") . "}}");
```
awick, line 3520: Try changing to something like
$meta->{version}++
esPut("/${PREFIX}lookups/_mapping", to_json({"_meta" => $meta}));

PR #1720, opened 2021-07-13
add bad status codes for failing to fetch pcap

### LINKED ISSUE #1719: PCAP retrieval endpoint returns errors along with a 200 status code
**Describe the bug**

The PCAP retrieval endpoint (i.e. `/node/pcap/id.pcap`) may respond with 200 status code and the error in the content. The content is typically something like:

```
Error talking to node 'x' using host 'y'
```

The endpoint should probably return a `5xx` status code instead.

**Elasticsearch version**: 7.8.0

**Arkime/Moloch version**: 2.4.2 (afaik, it also concerns the following versions)

**OS name and version**: Linux

**How was Arkime/Moloch built/installed:** Docker

**Provide logs, stack traces and steps to reproduce**: Run the capture on two nodes (say `x` and `y`), stop `x`, and retrieve a file stored in `x` using `y`.

31453:
- adds 404 for failing to get a list of sessions to write pcap
- adds 500 for failing to process session id
- uses a link in the client instead of opening a new window for the download so that an error doesn't redirect to a broken page

**_Limitations:_** 
- This **will not** set a bad status code if there is even one successful session that can be written as pcap. This is because the response is already writing the pcap file to the client, so a successful status code has already been sent.
- This **will not** display the error text sent from the server since a download is not a web page and the page that initiated the download cannot monitor the download response. If there is an error, the browser will display an error on its downloads page (depending on browser implementation, Chrome displays "Server Error"). Check logs for pcap download errors.

fixes #1719

### DIFF
--- a/tests/api-sessions.t
+++ b/tests/api-sessions.t
@@ -1,4 +1,4 @@
-use Test::More tests => 60;
+use Test::More tests => 70;
 use Cwd;
 use URI::Escape;
 use MolochTest;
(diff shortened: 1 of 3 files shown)


killerwhile:
LGTM. Thanks for the quick fix!

PR #1726, opened 2021-08-03
fix uncompress by requesting all packets if uncompress requested

31453:
disable packet length select input if uncompress or images requested
display reason for disabling packet length select
display warning that all packets will be requested when toggling on images or uncompress

### DIFF
--- a/viewer/apiSessions.js
+++ b/viewer/apiSessions.js
@@ -488,7 +488,8 @@ module.exports = (Config, Db, internals, molochparser, Pcap, version, ViewerUtil
     req.query.base = req.query.base || 'ascii';
     req.query.showFrames = req.query.showFrames === 'true' || false;
     // displaying images and uncompressing require all the packets from a session
-    req.query.packets = req.query.needimage || req.query.needgzip ? 10000 : +req.query.packets;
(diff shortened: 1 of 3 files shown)


viewer/apiSessions.js
```diff
@@ -487,6 +487,9 @@ module.exports = (Config, Db, internals, molochparser, Pcap, version, ViewerUtil
     req.query.line = req.query.line || false;
     req.query.base = req.query.base || 'ascii';
     req.query.showFrames = req.query.showFrames === 'true' || false;
+    // displaying images and uncompressing require all the packets from a session
+    // *2 because the packets can be out of order, we truncate again in the reassemble call
+    req.query.packets = req.query.needimage || req.query.needgzip ? 10000 : +req.query.packets * 2;
```
awick, line 492: * req.query.packets doesn't have to be set, so you probably should first set to the default of 200.
* the *2 shouldn't be here since req.query.packets is what it is truncated to. I think you should just do req.query.packets*2 down below so req.query.packets*2, 10);

31453: ah ok, I won't modify req.query.packets 🤭

PR #1730, opened 2021-08-09
UI build version

31453:
I'm not sure if this is the best way to do this, but it seems like the most suggested solution. It is definitely the easiest solution to get the version into the UI. However, it forces us to commit the version with the package(-lock).json files. But it gets the job done: 
Screen Shot 2021-08-09 at 1 54 15 PM

I do not think that this will work with merges though. Note how the version references the PREVIOUS commit. So this would need to be run and committed after a merge has happened, not as part of a merge (like is the case here).


awick:
Ya I would really prefer something that is done at webpack time automatically. Maybe add a precommand to the package.json that writes out a buildversion.js file with the info that we load in, kind of like how version.js.in works.

31453:
went with #1734 instead

PR #1731, opened 2021-08-10
use pre npm script to set build version

31453:
There are a few assumptions for this:
1. `printf` exists
2. `git` exists

If we want this to work without requiring `printf` and `git` we can add an npm git module to the webpack scripts that to fetch the latest tag and use it in the UI (like this one: https://www.npmjs.com/package/simple-git)


31453:
yep. have confirmed. it's magic 🪄

awick:
So version.js is generated from version.js.in when configure is run, so its a configure time variable which is ok for that variable. This variable I want to be the git version, which will be different. Capture print out both variables.

Its ok to require git to be installed, since thats what capture does. What I'm really wondering if there is a way to do this webpack. From https://webpack.js.org/guides/environment-variables/ it looks like we could do something like <code>webpack --env buildversion=\`shell git describe --tags\` --otherargs</code> which will get that into webpack config.js, but wasn't sure if you can get from config.js into app.js, i assume you can using templates or something.

31453:
went with #1734 instead

PR #1750, opened 2021-09-02
Add Modbus TCP parser

mcgillowen:

Classify with the port 502 as possible indicator of Modbus TCP and checks that protocol ID field is 0.
Parse the Modbus TCP classified packets and extract the following:
* Protocol ID (always 0)
* Unit ID
* List of transaction IDs, taken from source packets
* List of function codes, taken from source packets
* List of exception codes, taken from dest packets if there is an exception indicated in the function code 

None

No, no changes to viewer

Yes

## License


awick:
Cool!!  Looks like a great start, couple comments
* It is missing the detail.jade file to show the data in viewer
* It won't deal with tcp frames getting chopped up. If you don't want it to thats fine, but then you should only look at the first packet in each direction. Otherwise you should look at how some of the other parsers have a buf/len that it copies the data received into and then parses it out of that
* you are setting datalen field multiple times, but it only holds 1 value, so it will always be the last, is that ok?
* should there be any differences on source vs dest traffic?

mcgillowen:
> It is missing the detail.jade file to show the data in viewer
I've added it now

>It won't deal with tcp frames getting chopped up. If you don't want it to thats fine, but then you should only look at the first packet in each direction. Otherwise you should look at how some of the other parsers have a buf/len that it copies the data received into and then parses it out of that
Ok, I'll have a look, I am not sure which would be the best

>you are setting datalen field multiple times, but it only holds 1 value, so it will always be the last, is that ok?
Yes, that is my intention, I only want the final/total value and couldn't think of a better way to have it set at the last packet

> should there be any differences on source vs dest traffic?
No, not really IMO, I could possibly improve the transaction ID tracking by only saving them from source traffic

I am not an expert when it comes to Modbus TCP, I was asked to add Modbus classification by a customer and thought that these header values could be useful

Thanks for your comments

PR #1787, opened 2021-12-01
Fix cron queries

TNogisto:
New arkime sessions do not have "timestamp" field.

Seems to work :)

## License


awick:
thanks

PR #1794, opened 2021-12-14
add esquery to history

31453:
only implemented for sessions queries

### DIFF
--- a/db/db.pl
+++ b/db/db.pl
@@ -65,6 +65,7 @@
 # 68 - cron query enhancements
 # 70 - reindex everything, ecs, sessions3
 # 71 - user.roles, user.cont3xt
+# 72 - save es query in history
 
 use HTTP::Request::Common;
 use LWP::UserAgent;
@@ -76,7 +77,7 @@ use IO::Compress::Gzip qw(gzip $GzipError);
 use IO::Uncompress::Gunzip qw(gunzip $GunzipError);
 use strict;
 
-my $VERSION = 71;
+my $VERSION = 72;
 my $verbose = 0;
 my $PREFIX = undef;
 my $OLDPREFIX = "";
@@ -5237,6 +5238,11 @@ sub historyUpdate
       },
       "forcedExpression": {
         "type": "keyword"
+      },
+      "esQuery": {
+        "type": "object",
+        "dynamic": "true",
+        "enabled": "false"
       }
     }
   }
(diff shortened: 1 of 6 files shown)


db/db.pl
```diff
@@ -5237,6 +5238,11 @@ sub historyUpdate
       },
       "forcedExpression": {
         "type": "keyword"
+      },
+      "esQuery": {
```
31453, line 5242: should we stringify the json instead?

db/db.pl
```diff
@@ -7311,7 +7317,7 @@ sub dataNodes
         sessions3Update();
         historyUpdate();
         fieldsUpdate();
-    } elsif ($main::versionNumber <= 71) {
+    } elsif ($main::versionNumber <= 72) {
```
31453, line 7320: need this above (in the < 71 part) too?

viewer/viewer.js
```diff
@@ -17,7 +17,7 @@
  */
 'use strict';
 
-const MIN_DB_VERSION = 70;
+const MIN_DB_VERSION = 72;
```
31453, line 20: we need to upgrade the version or else the history index will complain about the esquery field

viewer/viewerUtils.js
```diff
@@ -273,8 +273,10 @@ module.exports = (Config, Db, molochparser, internals) => {
 
       const indices = await Db.getIndices(reqQuery.startTime, reqQuery.stopTime, reqQuery.bounding, Config.get('rotateIndex', 'daily'));
       if (indices.length > 3000) { // Will url be too long
+        req._molochESQuery = query;
```
31453, line 276: Do you think this is the best place to do this? It's the last step buildSessionQuery. Alternatively, I could do it in the finish callback every time buildSessionQuery is called (~14 times).

viewer/vueapp/src/components/history/History.vue
```diff
@@ -299,6 +299,11 @@
                     </em>
                   </div>
                 </div> <!-- /query params -->
+                <!-- es query -->
+                <div class="mt-3" v-if="item.esQuery">
+                  <h5>Elasticsearch Query</h5>
+                  <pre class="mr-3 ml-3">{{ item.esQuery }}</pre>
```
31453, line 305: Want this formatted in any special way?

db/db.pl
awick: yes i think we should store as a string

db/db.pl
awick: not in this case

viewer/viewer.js
31453: leave it to 70 for now

viewer/vueapp/src/components/history/History.vue
awick: MAKE PRETTY

PR #1832, opened 2022-02-22
Speed up column chooser

onegang:
Speed up the column chooser. Tiny change for your consideration. =)

We have an Arkime instance running, and have added more than one hundred fields due to various in-house enrichments on the data. This makes the column chooser in Arkime viewer very slow and laggy. This commit speeds it up quite a bit. 

**Did you run `npm run lint` from the viewer or parliament directory (whichever you are making changes to) and correct any errors** Yes

**Did you Ensure that all tests still pass by navigating to the `tests` directory and running `./tests.pl --viewer`** Yes

## License

### DIFF
--- a/viewer/vueapp/src/components/sessions/Sessions.vue
+++ b/viewer/vueapp/src/components/sessions/Sessions.vue
@@ -1101,7 +1101,7 @@ export default {
      * @return {number} number  The index of the visible header
      */
     isColVisible: function (id) {
-      return this.isVisible(id, this.tableState.visibleHeaders);
+      return this.tableState.visibleHeaders.indexOf(id);
     },
     /**
      * Toggles the visibility of a column given its id


31453:
@onegang Thank you for this PR! Unfortunately, this fix will not work for instances where Arkime has been upgraded from v2.x to v3.x. This is because the field names have changed quite a bit. We need the `isVisible` function that uses `FieldService.getField` to fetch the appropriate field regardless of the field name. 

For example, if a user has configured their columns in Arkime v2.x, they might have selected the "Dst data bytes" field which in v2.x was called `dstDataBytes` but is now called `server.bytes`. There are many other fields that have this issue. If you'd like to learn more, check out the PR that made all of these field name changes here: #1696.

Feel free to patch this locally for your instance if you started with v3. You'll have to make this change and bundle the client application using `npm run viewer:build` ([here is some more information](https://github.com/arkime/arkime/blob/main/CONTRIBUTING.md) about how to run and test viewer but I'm assuming you've already gotten this far since you were able to contribute this PR 😄 ). **Be warned that this will break your user's custom column configurations if you intend to upgrade from v2.x to v3.x in the future.**

31453:
Yes, that is possible and a good idea. It will speed the dropdown u for instances where there are many fields and most of those fields have the same `dbField` and `dbField2`. However, there will always be a slight lag when opening the dropdown since it is populated when it first opens (so that it doesn't calculate the list when the page loads and slow down the entire Sessions page).

31453:
We could not do a 1:1 replacement for the `getField` function since sometimes searches for a field using aliases. 

@awick What exactly do you mean by reverse map? Have a map where every field's fields point to the field object (dbField, dbField2, fieldECS, rawField, and exp)? Do you think the fields endpoint should create this map or the client?

awick:
I think the client would build the reverse maps. It used to at one point for some fields, but that might have been removed. Right now what to searched is passed in so I wasn't sure if that is required still or not.

As far as 1:1 getField replacement you would still pass ignoreAliases in, or I guess I don't know what you mean.

Something like this, IFF we can ignored the flelds passed in and use some globalFields things. I wasn't sure.
```
getField(search, fields, ignoreAliases)
  if (!search) return undefined
  if (reverseFields[search]) return reverseFields[search];
  if (!ignoreAliases) return reverseAliases[search];
  return undefined

```

31453:
We have different fields maps for different field properties: 
```
internals.fieldsMap[field._id] = source;
internals.dbFieldsMap[source.dbField] = source;
internals.dbFieldsMap[source.dbField2] = source;
if (source.fieldECS !== undefined) {
  internals.dbFieldsMap[source.fieldECS] = source;
  internals.fieldsMap[source.fieldECS] = source;
}
```
(this is in the Config `loadFields` function)

So, maybe we can combine `internals.dbFieldsMap` and `internals.fieldsMap` into one map that contains all the field properties (`_id` === `exp` above)? Then we'd just need to create an aliases map on the client.

31453:
Yeah, this is the server. I was thinking since we are already doing this on the server once on initial startup, it would make sense to do it there. No?

awick:
That means all the memory will be duplicated and will be a bigger response from server. Should be pretty easy to do on client

31453:
Yeah, the response would be much larger. I'm not saying it's harder to do on the client, just trying to understand all the pros/cons to each approach. I think you're right, the larger response is not something we need on initial load of the application (this data is sent in the first `appinfo` call).

31453:
@onegang this should be improved with this PR: #1834

onegang:
Thanks! Will check out the PR #1834 instead.

PR #1851, opened 2022-03-29
add password change to cont3xt settings

31453:
a user can change their own password in cont3xt
an arkime admin can change any users password on the arkime users page
a cont3xt admin can change any users password on the cont3xt users page

### DIFF
--- a/common/arkimeUtil.js
+++ b/common/arkimeUtil.js
@@ -198,6 +198,45 @@ class ArkimeUtil {
       { success: false, text: text || 'Server Error!' }
     );
   }
+
+  // express middleware to set req.settingUser to who to work on, depending if admin or not
+  // This returns fresh from db
+  static getSettingUserDb (req, res, next) {
+    let userId;
+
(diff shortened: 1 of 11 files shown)


common/arkimeUtil.js
```diff
@@ -198,6 +198,45 @@ class ArkimeUtil {
...
     );
   }
+
+  // express middleware to set req.settingUser to who to work on, depending if admin or not
+  // This returns fresh from db
+  static getSettingUserDb (req, res, next) {
+    let userId;
+
+    if (req.query.userId === undefined || req.query.userId === req.user.userId) {
+      // TODO what do you think about doing it this way?
```
31453, line 208: I added `getSettingUserDb` to `ArkimeUtil` so we can use it in `viewer` and `cont3xt`. However, it required both `Config` and `internals` to be present. So instead of using those, I added middleware to `viewer` and `cont3xt` that sets `req.regressionTests` and `req.noPasswordSecret` based on `Config` in `viewer` and `internals` in `cont3xt`.

@awick thoughts?

PR #1869, opened 2022-04-12
add version to all apps

31453:
cont3xt, parliament, wise
put version component in common
update cont3xt help for dates


awick:
could git.js be in common/build/git.js or something?

PR #1875, opened 2022-04-21
new sessionsMaxDaysForAggregations

awick:
Don't do maps/bytes/... aggregations if query is more than this many days. -1 turns off. Default 365. Not sure I like the var name yet.


awick:
Don't like this solution

PR #1883, opened 2022-05-03
Share Hunts With Roles

31453:
- can share hunt with specific roles
- make common roles dropdown
- put role parsing for UI in common

### DIFF
--- a/viewer/apiHunts.js
+++ b/viewer/apiHunts.js
@@ -781,7 +781,8 @@ ${Config.arkimeWebURL()}sessions?expression=huntId==${huntId}&stopTime=${hunt.qu
         stopTime: req.body.query.stopTime,
         view: req.body.query.view
       },
-      description: req.body.description
+      description: req.body.description,
+      roles: req.body.roles
     };
 
     async function doneCb (doneHunt, invalidUsers) {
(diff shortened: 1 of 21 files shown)


31453:
@awick should I update `MIN_DB_VERSION` in viewer?

viewer/apiHunts.js
```diff
@@ -888,12 +889,24 @@ ${Config.arkimeWebURL()}sessions?expression=huntId==${huntId}&stopTime=${hunt.qu
...
-          delete hunt.query;
+          let userHasRole = false;
+          if (hunt.roles.length) {
+            for (const role of hunt.roles) {
+              if (req.user.roles.indexOf(role) > -1) {
+                userHasRole = true;
+                break;
+              }
+            }
+          }
```
awick, line 900: Can this just be if hunt.roles.length && req.user.hasRole(hunt.roles) it already handles array vs single

viewer/apiHunts.js
```diff
@@ -1061,6 +1074,39 @@ ${Config.arkimeWebURL()}sessions?expression=huntId==${huntId}&stopTime=${hunt.qu
...
+   * @returns {boolean} success - Whether the operation was successful.
+   * @returns {string} text - The success/error message to (optionally) display to the user.
+   */
+  huntAPIs.updateHunt = async (req, res) => {
+    try {
+      const { body: { _source: hunt } } = await Db.getHunt(req.params.id);
+
+      // update properties
+      hunt.description = req.body.description;
+      hunt.roles = req.body.roles;
```
awick, line 1092: should we validate these roles exist?

viewer/apiHunts.js
```diff
@@ -781,7 +781,8 @@ ${Config.arkimeWebURL()}sessions?expression=huntId==${huntId}&stopTime=${hunt.qu
         stopTime: req.body.query.stopTime,
         view: req.body.query.view
       },
-      description: req.body.description
+      description: req.body.description,
+      roles: req.body.roles
```
awick, line 785: should we validate roles exist?

31453: Here, I was going to allow roles to be undefined (since old hunts will have undefined roles and the client handles this). Optionally, I could set this to an empty array if there are no roles defined upon creation by the client. I'm not sure which is best

PR #1884, opened 2022-05-06
turn off graph if > configured time range

### LINKED ISSUE #1878: Auto Graph Hiding for large time ranges
* The goal is to not do the aggregation query (facets=1) when a large time range is selected.
* The definition of what a large time range is should be configurable per cluster, with maybe a default of 30 days
* The user should be able to override and still request a graph be shown.
* For now just do in sessions/spiview, not sure if it makes sense for spigraph.

Q:
* should over ride be per query, per browser session, a period of time, or forever - initial thought is per query or per browser session

Possible implementation:
* When graph is showing if the user chooses a date range or custom range is greater than turnOffGraphDays then replace the graph area with a new area that states what is going on. Something like "This cluster is set to auto hide the graph after X days to help with performance blah blah"
* Have a button/pull down on info area that allows them to choose to do the graph query anyway.
* if pull down, could have options around just this once, for next 5 minutes or during the browser session etc.

31453:
allow override by user
fixes #1878


31453:
Things to consider:

- Additional options on how long the override persists?
- Do we like combining the hide/pin visualizations button with the override button? There are already SOOOOoooo many buttons.
- How should this interact with the show/hide visualization button? Currently, if you're time range > configured max days and your visualizations are hidden, you have to (1) click unhide, then (2) click fetch data. Is this right? Or should it override the max days on unhide? I think the two step process kind of makes sense since it forces a user to acknowledge that their search might take a long time. Thoughts?

viewer/viewer.js
```diff
@@ -2033,7 +2033,8 @@ app.use(cspHeader, setCookie, (req, res) => {
     anonymousMode: !!internals.noPasswordSecret && !Config.get('regressionTests', false),
     businesDayStart: Config.get('businessDayStart', false),
     businessDayEnd: Config.get('businessDayEnd', false),
-    businessDays: Config.get('businessDays', '1,2,3,4,5')
+    businessDays: Config.get('businessDays', '1,2,3,4,5'),
+    turnOffGraphDays: Config.get('turnOffGraphDays', 30)
```
31453, line 2037: is 30 days a good default?

31453:
What about an option to completely disable this per cluster? Maybe -1 or something for `turnOffGraphDays`?

Issue #1887, opened 2022-05-18
add inner mac address to the session for encapsulated protocols like: GRE, Geneve, VxLAN


scilicet64:
**Is your feature request related to a problem? Please describe.**
While capturing ERSPAN encapsulated traffic from a virtual environment we needed to filter on encapsulated mac-address to label traffic, this was not possible with the latest version. 

**Describe the solution you'd like**
We would like to discuss our solution before doing a pull-request. 
My working solution can be found in https://github.com/scilicet64/arkime/tree/multimac.
All tests pass and lint gives no error. Expected tests results for GRE, Geneve and VxLAN where modified because it now outputs the multiple mac addresses. 
For showing and debugging purpose it's currently using a debug with LOG function "moloch_field_macoui_add_debug" that needs to be removed when this solution is checked and discussed for performance.
My code is storing the offset to the first found mac address in the ethernet callback (moloch_packet_ether).
I picked a maximum of 10 encapsulated mac's to be stored (MOLOCH_ETHER_MAX_STORE 10), when it would arrive at encapsulated mac 11, it would overwrite the latest most inner mac. Three encapsulated addresses is what I have seen in practice, I would suggest ten is safe. 

**Describe alternatives you've considered**
First we just added the mac and ouifields for only the most inner mac, however we encoutered multilayer GRE/VxLAN and also needed to track those mac-addresses. Also we experimented with a new field name "encapsulated_mac", but the usability of this was not so great. We found adding it to the existing "mac" field feels more natural, just like in Wireshark for example.

**related issue**
issue: https://github.com/arkime/arkime/issues/1676 can possibly be solved with the same method for multiple encapsulated ip's. The issue how it's described is already implemented, it already shows the most inner encapsulated GRE-ip. However it can be improved for multiple ip's

awick:
Can you do a PR maybe in just your own repo so its easy to see the diff if you don't want to do one here.

awick:
My initial reaction is adding 60 bytes to an important 88 byte data structure isn't the right approach. I would think about using offsets instead to the mac addresses and/or double parsing if this feature is turned on.

scilicet64:
> Can you do a PR maybe in just your own repo so its easy to see the diff if you don't want to do one here.

here you are: https://github.com/arkime/arkime/pull/1888

scilicet64:
> 



[...]

Issue #1676, opened 2021-04-26
GRE tunnel endpoint IP-s should be visible in session details


TNogisto:
I think that GRE tunnel endpoint IP-s should be visible in session details. Currently they are not and this can lead to confusion, as the only way how to "identify gre" from Arkime viewer is to check if it has protocol "gre". This is not ideal.

Splitting "gre ip" to "gre src" and "gre dst" could also be a good idea.

**Elasticsearch version**: 7.9.1

**Arkime/Moloch version**:2.7.1

**OS name and version**: Ubuntu 18

**How was Arkime/Moloch built/installed:** deb

**Provide logs, stack traces and steps to reproduce**: open a gre tunnel session, would expect to see "gre tunnel" fields there, but they are not there.

scilicet64:
it seems the gre ip is indeed not visible as "GRE IP" like it should. it's listed in the IP db, but only when gre is the latest endpoint ip.

when a vxlan frame follows the gre ip is not visible like the screenshot. ERSPAN ip is 100.96.1.3* but the inner vxlan ip is shown: (not erspan ip: 10.0.0.*)

![erspan-failure](https://user-images.githubusercontent.com/46298739/169098463-33593178-237e-445e-bca4-2ff05bb743cd.png)

awick:
please provide a sample pcap, that really is the best way we can fix these types of issues

scilicet64:
this one is fixed in https://github.com/arkime/arkime/pull/1891
can I please add the Vxlan ip in the same style of gre ip under this PR/ticket ?

awick:
Let us know if this isn't what you were looking for

PR #1889, opened 2022-05-18
Add encapsulated mac/oui address to session

scilicet64:
**Add inner mac address to the session for encapsulated protocols like: GRE, Geneve, VxLAN**

While capturing ERSPAN encapsulated traffic from a virtual environment we needed to filter on encapsulated mac-address to label traffic, this was not possible with the latest version.
Using offsets to the macaddress in the packets the mac's are added to the session if they are found.

This PR is discussed in https://github.com/arkime/arkime/issues/1887.
Relevant: https://github.com/arkime/arkime/issues/1676, this issue can be solved using the same method. I would fix 1676 after discussing and finishing this PR.

YES, no errors found

yes, all tests pass after changing the GRE,Geneve and VxLAN tests, because they now return multiple mac addresses.

## License


capture/moloch.h
```diff
@@ -571,6 +572,8 @@ typedef struct molochpacket_t
     uint8_t        ipProtocol;     // ip protocol
     uint8_t        mProtocol;      // moloch protocol
     uint8_t        readerPos;      // position for filename/ops
+    uint8_t        ethCount;       // the number of ethernet frames found
+    uint32_t       ethOffset[MOLOCH_ETHER_MAX_STORE];  // offset to last ethernet frame from start
```
awick, line 576: At most this should be a uint16_t, although i wonder if it could be uint8_t

capture/moloch.h
```diff
@@ -571,6 +572,8 @@ typedef struct molochpacket_t
     uint8_t        ipProtocol;     // ip protocol
     uint8_t        mProtocol;      // moloch protocol
     uint8_t        readerPos;      // position for filename/ops
+    uint8_t        ethCount;       // the number of ethernet frames found
```
awick, line 575: this could be a :3 if you change the max to 8. make sure you move to other bit fields

capture/packet.c
```diff
@@ -329,12 +341,13 @@ LOCAL void moloch_packet_process(MolochPacket_t *packet, int thread)
         }
 
         if (pcapFileHeader.dlt == DLT_EN10MB) {
+            LOG("Adding normal mac to session with 0 offzet direction %d", packet->direction );
             if (packet->direction == 1) {
-                moloch_field_macoui_add(session, mac1Field, oui1Field, pcapData+0);
-                moloch_field_macoui_add(session, mac2Field, oui2Field, pcapData+6);
+                moloch_field_macoui_add_debug(session, mac1Field, oui1Field, 0,0,pcapData+0);
```
awick, line 346: when you change these and the following lines back, make sure you are consistent on formatting. i prefer ", " however line consistency is the most important.

capture/packet.c
```diff
@@ -1070,6 +1102,34 @@ LOCAL MolochPacketRC moloch_packet_ether(MolochPacketBatch_t * batch, MolochPack
 #endif
         return MOLOCH_PACKET_CORRUPT;
     }
+
```
awick, line 1105: definitely should have a config switch that controls this feature and maybe even sets a limit on number of macs to save.

capture/packet.c
```diff
@@ -1070,6 +1102,34 @@ LOCAL MolochPacketRC moloch_packet_ether(MolochPacketBatch_t * batch, MolochPack
...
+             data[9],
+             data[10],
+             data[11]);
+    LOG("moloch_packet_ether if lucky MAC B:%d, %s,%d",packet->ethCount,str,packet->ethOffset[packet->ethCount]);
+
+    if (packet->ethCount < MOLOCH_ETHER_MAX_STORE){
+        packet->ethCount++; // don't move the counter beyond the maximum number of ethernet frames we store, just overwrite last frame
+    }
+    else{
+        LOG("Warning: more then %d ethernet frames found, overwriting the most inner",MOLOCH_ETHER_MAX_STORE);
```
awick, line 1131: should have a flag/timer to only print this once or occasionally.

awick (changes requested):
Still not sure if i like this approach vs just setting a flag, or using the vpn/gre offset and reparsing a second time.

capture/moloch.h
scilicet64: it's about +50 per new mac in what I have seen. so when we do 7 we need to store a value like 350 or more. I will change this to uint16_t

scilicet64: +50 bytes further in the packet data pointer.
I needed to store all the offsets like this because I  didnt see another way to reach the session pointer.
If I understand how to reparse instead of storing all mac offsets it in the packet pointer, I can redo this later.
First I will commit all requested changes

capture/packet.c
scilicet64: # Default Arikime captures the inner mac address, increase this number if also the other mac addresses need to be captured, this increases need for memory (max=7)
#maxMacAddressCount=6

something like this?

scilicet64:
> I'm confused by the #1676 reference, which I thought was just asking we display the gre ips, which has been fixed. Is it something else?

I will explain this in the other ticket, so it will not litter the information here. For this context it's indeed not relevant anymore.

scilicet64:
Looking at the bigger picture, also the other tunneled IP's need to be in the viewer. 
In current code the offset to gre ip is named vpnIpOffset. maybe created with the idea to support other vpnIps.
However in the real life sample of [tests/pcap/gre+erspan+vxlan.pcap](https://github.com/arkime/arkime/pull/1891/commits/94b731505b53dcd4746f7b04636abbb940242851#diff-ada8b38da52917c57f1699e0123b3106b38eb5d53d5906fc4484e143a973bec9)  there are multiple vpnIps.
We can add a new named IP variable for each type of vpn (gre/vxlan/geneve/etc.) like it's now or make a new generalized vpnIpsOffset[maxIPs] array like the current the current PR is doing for mac's.

awick:
So I'm still waiting for a description of the need here. What is the use case? What is the problem you are trying to solve? 

It is not a goal of Arkime to be wireshark or network performance monitor.

I could potentially see captureing more than 1 or 2 levels of VPN, but need to see/read the use case. Also need to understand if there are N wrappers, but we are only storing 2 levels, would you want the inner or outer 2. I would guess inner, but could be wrong.

As far as implementation I'm still thinking that if we want to support N levels it should either be done with double parsing OR potentially could combine the two arrays and use the top bits to mark if ether vs ip vs vpn headers.

scilicet64:
> So I'm still waiting for a description of the need here. What is the use case? What is the problem you are trying to solve?
> 
> It is not a goal of Arkime to be wireshark or network performance monitor.
> 
> I could potentially see captureing more than 1 or 2 levels of VPN, but need to see/read the use case. Also need to understand if there are N wrappers, but we are only storing 2 levels, would you want the inner or outer 2. I would guess inner, but could be wrong.
> 
> As far as implementation I'm still thinking that if we want to support N levels it should either be done with double parsing OR potentially could combine the two arrays and use the top bits to mark if ether vs ip vs vpn headers.

Our use-case is analyzing traffic in 3 layers like https://github.com/scilicet64/arkime/blob/94b731505b53dcd4746f7b04636abbb940242851/tests/pcap/gre-erspan-vxlan.pcap
The typical traffic in our virtualized environment has 3 pairs of MAC and IP. For example like in the supplied pcap:
GRE MAC/IP
VXLAN MAC/IP
INNER MAC/IP
To do fine analytics and labeling of the data it's needed to filter/search and to display the middle (VXLAN) and  list MAC and IP's in the detailview.
For the analytic example we discovered a strange IP in the virtual network (with over 8000 Machines) and needed to pinpoint the strange behaviour.
In this specific case we had the need to filter the MAC's of the middle layer, to see all the traffic of this specific VM when it did not have the assigned IP as expected. The GRE tunnel IPs and MACs just represent the info from the physical boxes sending/receiving the actual packets. For our analysis this outer layer is useless. But maybe not for others monitoring bigger ISP links.

Also when we have the MAC's we can assign the VMcenter assets to the middle layer with the Wise plugin (addapted for MAC-matching) to make labeling possible.

The N wrappers is not a usecase for us, we have worked upto 3 layers in several variations like above.
We dont mind if MAC/IP's are grouped in the generic MAC or IP field.

So a workable solution in this case would be to have something like a MiddleLayer Mac/IP instead of N-layers. Looking at the current code, I think adding a VXLAN IP just like the current GRE IP would have least impact. Another option is renaming the GRE to MiddleLayer and use this for all tunnels to store Layer-1 (Middle). What do you support most, Im sure you also have own ideas?

awick:
I agree, if we just need the previous ip and previous ether adding a vpnEtherOffset would be the way to go. The gre field names are wrong since it could be other things that just gre, so we should fix that too.

awick (changes requested):
Looking good

capture/packet.c
```diff
@@ -1504,6 +1559,22 @@ void moloch_packet_init()
         "fieldECS", "destination.mac",
         (char *)NULL);
 
+    vpnmac1Field = moloch_field_define("general", "lotermfield",
+                                    "vpnmac.src", "Src VPNMAC", "source.vpnmac",
+                                    "Source ethernet vpn mac addresses set for session",
+                                    MOLOCH_FIELD_TYPE_STR_HASH,  MOLOCH_FIELD_FLAG_ECS_CNT | MOLOCH_FIELD_FLAG_LINKED_SESSIONS | MOLOCH_FIELD_FLAG_NOSAVE,
```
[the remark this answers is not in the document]
scilicet64: in a first private draft I used encapsulatedMAC for that reason. Do you think this also applies to vpnIP?

capture/db.c
```diff
@@ -738,6 +740,9 @@ void moloch_db_save_session(MolochSession_t *session, int final)
         if (session->fields[mac1Field]) {
             SAVE_FIELD_STR_HASH(mac1Field, MOLOCH_FIELD_FLAG_ECS_CNT);
         }
+        if (session->fields[vpnmac1Field]) {
+            SAVE_FIELD_STR_HASH(vpnmac1Field, MOLOCH_FIELD_FLAG_ECS_CNT);
```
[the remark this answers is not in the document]
scilicet64: just tried this and testing failed... When I remove this saving there, It seems to overwrite all the other fields(as,geo,ip,mac,bytes,packets,port) in source.xxx and destination.xxx from the special saving. Does this imply we can't save it to the general "source" and "destination" fields?  If so, do you want this in another group?

capture/packet.c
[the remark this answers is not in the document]
awick: encapsulated is too long and hard to spell :) Maybe wrap? maybe ECS has a term for it, ill have to look

scilicet64:
this is the result of the viewer with all new fields:
image

PR #1892, opened 2022-05-19
Add required dependencies to Arch package

Didr:
I've added a few missing dependencies for ArchLinux package. This is information for the pacman installer to figure out which other packages it needs to install.
"perl-http-message perl-lwp-protocol-https perl-json" is required for `db/db.pl`.
"libnet" is required for `bin/capture`.

I found these by looking at the build container: https://github.com/arkime/arkime/blob/main/release/buildArch/Dockerfile

**Testing**
I've been unsuccessful in installing screwdriver.cd, so I haven't tested this yet.

## License


awick:
awesome, could you add to the release-arch section also?

Didr:
Good catch! I've added the dependencies there too.

awick:
thanks

PR #1903, opened 2022-05-25
share shortcuts with users/roles

31453:


awick (commented):
db.pl set-shortcut command should at least map the --shared option to setting the role

PR #1907, opened 2022-06-01
share notifiers

31453:
addresses the shortcuts portion of #1901

### DIFF
--- a/db/db.pl
+++ b/db/db.pl
@@ -5515,6 +5515,16 @@ sub notifiersUpdate
 {
   "_source": {"enabled": "true"},
   "dynamic": "true",
+  "dynamic_templates": [
+    {
+      "string_template": {
+        "match_mapping_type": "string",
+        "mapping": {
+          "type": "keyword"
+        }
+      }
+    }
+  ],
   "properties": {
     "name": {
       "type": "keyword"
--- a/db/db.pl
+++ b/db/db.pl
@@ -5517,6 +5517,24 @@ sub notifiersUpdate
   "properties": {
     "name": {
       "type": "keyword"
+    },
+    "users": {
+      "type": "keyword"
+    },
+    "roles": {
+      "type": "keyword"
+    },
+    "user": {
+      "type": "keyword"
+    },
+    "type": {
+      "type": "keyword"
+    },
+    "created": {
+      "type": "date"
+    },
+    "updated": {
+      "type": "date"
     }
   }
 }';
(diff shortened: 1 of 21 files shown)


31453:
Thoughts on unique Notifier names? I am no longer enforcing/requiring Notifier names to be unique (since the API uses ID instead of Name for notifier identification). The only problem I see is that if there are multiple Notifiers with the same name, the users will get confused when trying to apply a Notifier to a hunt/periodic query.

db/db.pl
```diff
@@ -5489,6 +5489,80 @@ sub lookupsUpdate
...
+  notifiersUpdate();
+}
+
+sub notifiersUpdate
+{
+    my $mapping = '
+{
+  "_source": {"enabled": "true"},
+  "dynamic": "true",
+  "properties": {
```
awick, line 5517: unless there is a reason not to, I think we should also add a dynamic string template, it basically makes anything not listed automatically a keyword instead of ending up with text fields. See fields_v30 and others. Still need to definitely list out the date fields.

    my $mapping = '
{
  "_source": {"enabled": "true"},
  "dynamic_templates": [
    {
      "string_template": {
        "match_mapping_type": "string",
        "mapping": {
          "type": "keyword"
        }
      }
    }
...

db/db.pl
```diff
@@ -67,7 +67,7 @@
 # 71 - user.roles, user.cont3xt
 # 72 - save es query in history, hunt description
 # 73 - hunt roles
-# 74 - shortcut sharing with users/roles
+# 74 - shortcut sharing with users/roles, notifiers index
```
awick, line 70: This will need to be 75 because those that already installed 74 (like us) won't get notifiersCreate/Move called.

db/db.pl
```diff
@@ -7441,21 +7525,27 @@ sub dataNodes
...
-    if ($main::versionNumber <= 74) {
+    if ($main::versionNumber < 74) {
         checkForOld7Indices();
         sessions3Update();
         historyUpdate();
         huntsUpdate();
         lookupsUpdate();
+        notifiersCreate();
+        notifiersMove();
+    } elsif ($main::versionNumber <= 74) {
```
awick, line 7536: change about to < 75 and this one to <= 75

db/db.pl
awick: you can leave them there, it is slight more efficient for ES if the fields you know about are listed.

PR #1911, opened 2022-06-10
tests.pl: add --skipgeo option

matoro:
This allows capture tests to be run offline.  It assumes that the usual
Geo files are already present in the directory.  Needed for packaging
that doesn't allow network access at build/test time.

## License

### DIFF
--- a/tests/tests.pl
+++ b/tests/tests.pl
@@ -23,20 +23,22 @@ my $INSECURE = "";
 
 ################################################################################
 sub doGeo {
-    if (! -f "ipv4-address-space.csv") {
-        system("wget https://s3.amazonaws.com/files.molo.ch/testing/ipv4-address-space.csv");
-    }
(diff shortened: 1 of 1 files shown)


tests/tests.pl
```diff
@@ -468,6 +474,7 @@ sub doViewer {
     print "  --elasticsearch <url> Set elasticsearch URL\n";
     print "  --debug               Turn on debuggin\n";
     print "  --valgrind            Use valgrind on capture\n";
+    print "  --skipgeo             Skip downloading Geo files, assume already present (useful for offline)\n";
```
awick, line 477: I'm confused why this change is needed, since if the files already exist it won't try and download them anyway, and the help says to assume they are present.

awick (changes requested):

matoro:
Oops, you're right - not sure why I didn't catch that this check already exists!  My bad.

Issue #1910, opened 2022-06-08
The viewer cannot view the pcap data when gzip and localindex are both enabled.


lns0303:
**Describe the bug**
Hi andy, I found a bug where the **viewer cannot view the pcap data** when **gzip and localindex are both enabled**.

![image-20220608131641074](https://user-images.githubusercontent.com/37865384/172553274-ca6b987d-3998-4bde-b0a8-a94be35fe676.png)

But when I turn on only localindex or only gzip, it works fine. Then I found the cause of the problem: **The localindex encoding cannot store the packetPos of gzip**.

When gzip is turned on, `packetPos` is calculated by (#728)

``` c
packetPos = (offsetOfBlockStart << uncompressedBits) + (offsetOfPacketStartInUncompressedData)
```

`uncompressedBits` calculated by 

``` c
simpleGzipBlockSize = moloch_config_int(NULL, "simpleGzipBlockSize", 32000, 0, 0xfffff);
uncompressedBits = ceil(log2(simpleGzipBlockSize));		// default 15, range 15 ~ 20
```

In my environment `maxFileSizeG = 12`, so `offsetOfBlockStart` is a number of **2^30 bytes** level. After calculating `packetPos` will reach **2^45** that `localindex` can't hold.

``` c
    localPcapIndex = moloch_config_boolean(NULL, "localPcapIndex", FALSE);
    if (localPcapIndex) {
        if (config.pcapDir[1]) {
            DPS_LOG_WARN("WARNING - Will always use first pcap directory for local index");
        }

        // localindex max storage length is 0x07ffffffffL(2^35), so it can't store gzip packetPos(2^45)
        config.maxFileSizeB = MIN(config.maxFileSizeB, 0x07ffffffffL);
        config.gapPacketPos = FALSE;
        moloch_writer_index = writer_simple_index;
    }
```

I wonder if it is possible to extend the localindex to **56bits**? This way `uint64_t` is just right for storage(56bits for data, 8bits for flag) . So that in the extreme case (uncompressedBits=20), the maximum support can be up to a file **64gb** (2^56 == 2^6 * 2^30 * 2^20).

I made the following changes and the problem no longer occurs.

``` c
diff --git a/capture/writer-simple.c b/capture/writer-simple.c
index 137f0aa..bd43ecd 100644
--- a/capture/writer-simple.c
+++ b/capture/writer-simple.c
@@ -767,8 +767,26 @@ void writer_simple_index (MolochSession_t * session)
             }
[...]

PR #1912, opened 2022-06-12
Extend localindex to fix issue #1910

lns0303:
Viewer cannot view or download the pcap data when gzip and localindex are both enabled. That's because when gzip is turned on, the **packetPos** value exceeds the range stored by the localindex (0~2^35-1). To fix this, I expanded the localindex to 56bits and limited the maximum file size to `0xffffffffffffffUL >> (uncompressedBits + 1)`. For more information, please see #1910 
After my testing, the problem was solved.

#1910 

yes

yes

## License


capture/writer-simple.c
```diff
@@ -865,7 +883,12 @@ void writer_simple_init(char *name)
             LOG("WARNING - Will always use first pcap directory for local index");
         }
 
-        config.maxFileSizeB = MIN(config.maxFileSizeB, 0x07ffffffffL);
```
awick, line 868: The else isn't needed since the max of maxFileSizeB is 1024^4 or 0x10000000000

The goal of the previous maxFileSizeB was to stop folks from making it too big where variable length encoded ints might start using more space than fixed size. Because of the gzip issue we have to allow it, but I still wonder if we should allow folks to create files so big.

lns0303: ok, that's right.

PR #1919, opened 2022-06-22
export table data with filter

TobySalusky:
* IntegrationTable syncs filteredData with IntegrationValue

* selectedFields also change filteredData


31453 (approved):
👍 looks great!

PR #1922, opened 2022-06-27
Table search highlighting

TobySalusky:
* Added HighlightableText component that uses an array of spans to highlight text (this element wraps any text rendered at the base layer that can be held in an IntegrationTable cell)
* IntegrationTables calculate span highlighting information for each cell while filtering data
* IntegrationArrays use an array of span highlighting arrays to highlight each of their sub-elements individually
* Updated URL test for IntegrationValue (necessary due to extra `<span></span>` wrapper around textual link)


cont3xt/vueapp/src/components/integrations/IntegrationTable.vue
```diff
@@ -250,37 +254,88 @@ export default {
...
-      this.filteredData = this.data.filter((row) => {
-        let match = false;
-        const query = newSearchTerm.toLowerCase();
 
-        for (const field of this.fields) {
+      /**
+       * @param arr the array to be simultaneously filtered/mapped over
+       * @param filterMapFunc (arrElement) -> [boolean, outputElement]
+       */
+      const filterMap = (arr, filterMapFunc) => {
```
awick, line 267: I've seen folks recommend using reduce() as a solution for filter+map, did you look at that?

cont3xt/vueapp/src/components/integrations/IntegrationTable.vue
```diff
@@ -250,37 +254,88 @@ export default {
...
+          // create span information for highlighting
+          return [{ start: matchStart, end: matchStart + matchObj[0].length }];
+        };
+        const matchAndCellHighlightDataFromValue = (fullValue, matchQuery) => {
+          let match = false;
+          let cellHighlightData = null;
+          if (Array.isArray(fullValue)) {
+            cellHighlightData = [];
+            for (const [index, value] of fullValue.entries()) {
+              const matchObj = value.toString().toLowerCase().match(matchQuery);
```
awick, line 292: if we are truly doing RegExp then you should convert matchQuery to once before calling match, otherwise match will do for every call. If not supporting regexp should this be a indexOf?

cont3xt/vueapp/src/utils/HighlightableText.vue
```diff
@@ -0,0 +1,83 @@
...
+      }
+
+      const spanList = [];
+      const expandedHighlights = [{ start: -1, end: 0 }, ...this.highlights];
+
+      const correctHighlightStart = (start) => {
+        return Math.min(start, this.text.length - 3);
+      };
+
+      for (const [i, highlightedSpan] of expandedHighlights.entries()) {
```
awick, line 53: Personally I hate entries :) Just curious, is there an advantage over a normal for loop?

cont3xt/vueapp/src/utils/displayValues.js
```diff
@@ -0,0 +1,37 @@
+import dr from 'defang-refang';
```
awick, line 1: Not sure findDisplayValue is the right name. Should it be format/display/...? Also the file name and function name should probably match if this is the only thing? Does it need its own file?

cont3xt/vueapp/src/utils/HighlightableText.vue
awick: nope, just wondering

cont3xt/vueapp/src/components/integrations/IntegrationTable.vue
TobySalusky: Yea, I noticed that (and meant to refactor it to do so, but I think I got side-tracked somewhere along the way)—I switched to using reduce in my latest commit.

cont3xt/vueapp/src/components/integrations/IntegrationTable.vue
TobySalusky: Good idea! I made the change to create a RegExp beforehand, since regex functionality is desired.

cont3xt/vueapp/src/utils/displayValues.js
TobySalusky: I changed the function (and filename) to formatValue and added some more comments. I figured it would be most appropriate to have it in its own file. I suppose it could also be put in filters.js (although I didn't put it here since findDisplayValue/formatValue has more to do with how the data is actually stored by integrations).

PR #1924, opened 2022-06-29
addition of view and submit query params for cont3xt

TobySalusky:
* view and submit query parameters that are used after views/integrations have been loaded following page load (signaled by immediateSubmissionReady on the store)
* view is synced with the users currently selected view
* view is a case-insensitive string that selects a view by name (with special cases to ensure 'none' and 'all' work as intended)
* submit=y is tacked onto shareable links generated using Shift+L to make shared links automatically issue a search (one time only)

### DIFF
--- a/cont3xt/vueapp/src/components/pages/Cont3xt.vue
+++ b/cont3xt/vueapp/src/components/pages/Cont3xt.vue
@@ -509,7 +509,6 @@ export default {
 
       // apply 'view' query param
       if (this.$route.query.view !== undefined) {
-        console.log(this.$route.query);
         this.setViewByQueryParam(this.$route.query.view);
       }
 
--- a/cont3xt/vueapp/src/components/pages/Cont3xt.vue
+++ b/cont3xt/vueapp/src/components/pages/Cont3xt.vue
@@ -417,7 +417,8 @@ export default {
       'getIntegrationsError', 'getLinkGroupsError', 'getLinkGroups',
       'getSidebarKeepOpen', 'getShiftKeyHold', 'getFocusSearch',
       'getIssueSearch', 'getFocusStartDate', 'getFocusLinkSearch',
-      'getToggleCache', 'getDownloadReport', 'getCopyShareLink'
+      'getToggleCache', 'getDownloadReport', 'getCopyShareLink',
+      'getViews', 'getImmediateSubmissionReady', 'getDoableIntegrations'
     ]),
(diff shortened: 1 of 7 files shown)


cont3xt/vueapp/src/components/pages/Cont3xt.vue
```diff
@@ -742,6 +772,39 @@ export default {
...
+      }
+      const lowercaseViewName = viewName.toLowerCase();
+      for (const view of this.getViews) {
+        if (view.name.toLowerCase() === lowercaseViewName) {
+          this.$store.commit('SET_SELECTED_VIEW', view.name);
+          this.$store.commit('SET_SELECTED_INTEGRATIONS', view.integrations);
+          return;
+        }
+      }
+      removeViewParam();
```
awick, line 794: why remove the view if it is a good view? If we left it, it would be easier to copy paste the url WITH view.

awick (commented):
Thoughts about adding a All/None to the view menu, maybe after a sep so its clear there are system ones?

cont3xt/vueapp/src/components/pages/Cont3xt.vue
```diff
@@ -742,6 +772,39 @@ export default {
...
+
+      if (lowercaseViewName === 'none') {
```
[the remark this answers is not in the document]
WilliamSalusky: Good idea! `all` and `none` should become system defaults. I'd like to see system defaults at the top of the drop down with a separator before user created views.

cont3xt/vueapp/src/components/pages/Cont3xt.vue
WilliamSalusky: My initial thought was to drop it after initial load of shared link click through, but now I think it *would* be better to retain it in the query string.

cont3xt/vueapp/src/components/pages/Cont3xt.vue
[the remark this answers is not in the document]
TobySalusky: This has now been added in the latest commits :)

cont3xt/vueapp/src/components/pages/Cont3xt.vue
TobySalusky: The view query parameter is now only dropped when it is invalid or is not found among the users views (system defaults (all/none) are also safe).

PR #1931, opened 2022-07-12
Audit log history

TobySalusky:
* added audit log history page and ES/LMDB backend support
* added view selector shortcut (shift+V)
* default itype card backing in case of no data (no integrations)
* other minor fixes and refactors (ThreatStream tags, weirdness with modifying searchTerm after search, failing to load integrations panel without persisted data)


cont3xt/vueapp/src/components/pages/AuditHistory.vue
```diff
@@ -0,0 +1,273 @@
...
+  },
+  mounted () {
+    this.loadAuditsFromSearch();
+  }
+};
+</script>
+
+<style scoped>
+.indicator-limit-width {
+  max-width: 30rem;
```
awick, line 270: github puts this in red, something wrong with it?

cont3xt/vueapp/src/components/pages/AuditHistory.vue
```diff
@@ -0,0 +1,273 @@
...
+        >
+      </template>
+      <!--   /customize column sizes   -->
+
+      <!--   Button Column   -->
+      <template #cell(buttons)="data">
+        <b-button v-if="getUser && getUser.removeEnabled"
+            @click="deleteLog(data.item._id)"
+            class="btn btn-xs btn-warning"
+            v-b-tooltip.hover="'Delete log'">
```
31453, line 61: Use the `noninteractive` feature of tooltips so that hovering over the tooltip doesn't keep it activated. This will help when a user wants to click a button above a button that has a tooltip activated.

https://bootstrap-vue.org/docs/components/tooltip#accessibility

cont3xt/vueapp/src/components/pages/AuditHistory.vue
```diff
@@ -0,0 +1,273 @@
...
+            :disabled="!filter"
+            title="Remove the search text">
+          <span class="fa fa-close" />
+        </b-button>
+      </template>
+    </b-input-group>
+    <!--  search bar form  -->
+
+    <!-- time range inputs -->
+    <time-range-input class="mt-1"
```
31453, line 30: I think the time range input could be put next to the search input to utilize the space more. I don't imagine history searches are going to be too long.

Also, I think there should be a little more margin/padding between the search area and the table (I'm sure @awick  disagrees though 😄)

cont3xt/vueapp/src/components/pages/AuditHistory.vue
TobySalusky: It looks like github highlights set-sizings (i.e. pixel or rem measurements)—but in this case that is what I meant since I want to avoid massive strings from blowing up the table. I would have preferred relative measurements like percentages, but I can't find a good way for them to play nice with set-sized features like the buttons and timestamps.

cont3xt/vueapp/src/components/services/AuditService.js
```diff
@@ -0,0 +1,72 @@
...
+   * @returns {Promise} - The promise that either resolves to an `Audit[]` or rejects in error
+   */
+  getAudits (query) {
+    return new Promise((resolve, reject) => {
+      const options = {
+        method: 'GET',
+        params: query,
+        url: 'api/audits'
+      };
+      Vue.axios(options).then((response) => {
```
31453, line 40: Can we use fetch here instead of axios. I think that cont3xt has been able to avoid axios altogether (minimizes the bundle size and increases initial load times https://bundlephobia.com/package/axios@0.27.2).

cont3xt/db.js
```diff
@@ -222,6 +253,51 @@ class DbESImplementation {
...
+      // If already exists ignore error
+      if (err.meta.body?.error?.type !== 'resource_already_exists_exception') {
+        console.log(err);
+        process.exit(0);
+      }
+    }
+
+    await this.client.indices.putMapping({
+      index: 'cont3xt_history',
+      body: {
```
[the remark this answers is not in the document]
TobySalusky: I think I fixed this with my latest commit, does it look right now?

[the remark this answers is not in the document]
awick: yep

31453 (approved):
👍 LGTM! Great work! ✨

PR #1938, opened 2022-07-15
Toggle all history cont3xt

TobySalusky:
* button visible to cont3xtAdmin that toggles whether they see all history or just their own
* userId column only shown to cont3xtAdmin
* query parameter seeAll=true on history page


31453 (approved):
👍 ✨

PR #1942, opened 2022-07-19
fixed integration button mismatch

TobySalusky:
When dealing with multiple indicators of one itype (e.g. DNS IPs), previously, the integration-buttons on each only reflect the top indicator's data (resulting in incorrect counts and/or missing data).

They are now matched correctly.


31453 (approved):
Nice find! 👀

Issue #1952, opened 2022-08-02
export PCAP fails if there are any "sessions" without packets


mmguero:
**Describe the bug**

If you attempt to do an **Export PCAP** from the sessions view (using the dropdown to the right of the search bar) for visible sessions when some of the sessions which will be included are not backed by PCAP (in my case they are session documents written via logstash but otherwise matching the same schema as the native arkime session documents), an error is raised and no PCAP is exported. See log at the bottom of this bug report.

I'm running v3.4.2, [here's the line number at `apiSessions.js:1147`](https://github.com/arkime/arkime/blob/v3.4.2/viewer/apiSessions.js#L1147) it's complaining about.

I think a check needs to be placed in here somewhere to not attempt to retrieve `packetPos`/packet location/payload if no packet is present.

**Elasticsearch version**:

OpenSearch v2.1.0

**Arkime/Moloch version**:

Arkime v3.4.2

**OS name and version**:

Debian 11 (x86_64 Linux)

**How was Arkime/Moloch built/installed:** (rpm, deb, easybutton, ...)

[easybutton build in Docker](https://github.com/mmguero-dev/Malcolm/blob/11ab87ebcde33657fe4928e18ba27381fe658848/Dockerfiles/arkime.Dockerfile#L83)

**Provide logs, stack traces and steps to reproduce**:

Logs:

```
Trace: ES Search Error - query: {
  "index": "arkime_sessions3-220802",
  "body": {
    "query": {
      "ids": {
        "values": [
          "220802-aH83zjkVmIf3-4352CFPsw"
        ]
      }
    },
    "_source": false,
    "fields": [
      "node",
      "network.packets",
      "packetPos",
      "source.ip",
      "source.port",
      "destination.ip",
      "destination.port",
      "ipProtocol",
      "packetLen"
    ],
    "profile": false
  },
  "rest_total_hits_as_int": true
} err: TypeError: Cannot read properties of undefined (reading 'length')
    at /opt/arkime/viewer/apiSessions.js:1147:43
    at fixPacketPos (/opt/arkime/viewer/db.js:408:14)
    at /opt/arkime/viewer/db.js:532:12
    at Object.exports.search (/opt/arkime/viewer/db.js:570:17)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at Object.exports.search (/opt/arkime/viewer/db.js:572:13)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
[...]

PR #1953, opened 2022-08-02
address arkime/arkime#1952, in which export PCAP fails if there are any sessions without packets

mmguero:
See #1952. There needs to be a check to make sure `fields.packetPos` is not null when getting the list for PCAP export, in `sessionAPIs.processSessionId` in `apiSessions.js.

arkime/arkime#1952

## License

### DIFF
--- a/viewer/apiSessions.js
+++ b/viewer/apiSessions.js
@@ -1140,7 +1140,11 @@ module.exports = (Config, Db, internals, ViewerUtils) => {
 
       const fields = session.fields;
 
-      if (maxPackets && fields.packetPos && fields.packetPos.length > maxPackets) {
+      if (!fields.packetPos) {
+        return endCb(null);
+      }
+
+      if (maxPackets && fields.packetPos.length > maxPackets) {
         fields.packetPos.length = maxPackets;
       }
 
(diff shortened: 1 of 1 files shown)


mmguero:
I tested this by getting a sessions query with 490 entries (total), 159 of which are true arkime sessions with PCAP. I did an export PCAP of the entire matching 490 sessions, and ended up with a valid PCAP file with 52,113 packets. I then filtered to just arkime sessions in the sessions view, and exported that result set and ended up with a PCAP of the exact same size and number of packets.

awick:
I wonder if the following would work. Do you have time to try?
<pre>
    Db.getSession(id, options, (err, session) => {
      if (err || !session.found) {
        console.log('ERROR - session get error in processSessionId', util.inspect(err, false, 50), session);
        return endCb('Session not found', null);
      }
      if (!fields.packetPos) {
        return endCb(null);
      }
</pre>

mmguero:
Sure I'll give it a shot.

mmguero:
Hmmm, fields would have to be assigned first, I'll move the check you added a few lines down and try that.

mmguero:
Yes that worked, I will update my PR.

mmguero:
Okay, I just pushed a second commit, only change now is just the addition of:

```
      if (!fields.packetPos) {
        return endCb(null);
      }
```

right after the local fields variable is assigned. I did the same test as I described earlier, got the same results in the exported PCAPs I tried.

awick:
thanks!

PR #1961, opened 2022-08-09
test history and lg deletes

awick:
* still log even if not tags in search
* test some history stuff
* fix delete LG/history failing if not in ES


TobySalusky (approved):
It all looks good to me 🙂

PR #1973, opened 2022-08-17
Send CSP-headers for cyberchef.html

p4fg:
The file viewer/viewer.js has a defintion for serving the static file cyberchef.html.
The code does not work like intended as no CSP-headers are sent out with the response:

```
HTTP/1.1 200 OK
X-Frame-Options: DENY
X-XSS-Protection: 1; mode=block
Accept-Ranges: bytes
Cache-Control: public, max-age=86400
Last-Modified: Wed, 17 Aug 2022 17:07:47 GMT
ETag: W/"fdec-182acc682fb"
Content-Type: text/html; charset=UTF-8
Vary: Accept-Encoding
Date: Wed, 17 Aug 2022 18:01:03 GMT
Connection: close
Content-Length: 65004
```

This PR fixes the problem and results in the defined CSP-header being sent:

```
HTTP/1.1 200 OK
X-Frame-Options: DENY
X-XSS-Protection: 1; mode=block
Content-Security-Policy: default-src 'self'; style-src 'self' 'unsafe-inline'; script-src 'self' 'unsafe-eval' 'unsafe-inline'; object-src 'self' data:
Accept-Ranges: bytes
Cache-Control: public, max-age=86400
Last-Modified: Wed, 17 Aug 2022 17:07:47 GMT
ETag: W/"fdec-182acc682fb"
Content-Type: text/html; charset=UTF-8
Vary: Accept-Encoding
Date: Wed, 17 Aug 2022 18:01:12 GMT
Connection: close
Content-Length: 65004
```

## Checks
* npm run lint: YES
* ./tests.pl --viewer: YES

## License

### DIFF
--- a/viewer/viewer.js
+++ b/viewer/viewer.js
@@ -1968,10 +1968,10 @@ app.get(
 );
 
 // cyberchef apis -------------------------------------------------------------
-app.get('/cyberchef.html', express.static( // cyberchef client file endpoint
+app.get('/cyberchef.html', [ArkimeUtil.missingResource, cyberchefCspHeader], express.static( // cyberchef client file endpoint
   path.join(__dirname, '/public'),
   { maxAge: dayMs, fallthrough: false }
-), ArkimeUtil.missingResource, cyberchefCspHeader);
+));
 
 app.get( // cyberchef endpoint
   '/cyberchef/:nodeName/session/:id',
--- a/viewer/viewer.js
+++ b/viewer/viewer.js
@@ -1968,10 +1968,10 @@ app.get(
 );
 
 // cyberchef apis -------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


viewer/viewer.js
```diff
@@ -1968,10 +1968,10 @@ app.get(
 );
 
 // cyberchef apis -------------------------------------------------------------
-app.get('/cyberchef.html', express.static( // cyberchef client file endpoint
+app.get('/cyberchef.html', [ArkimeUtil.missingResource, cyberchefCspHeader], express.static( // cyberchef client file endpoint
```
awick, line 1971: Thanks! This doesn't seem right either since missingResource returns an error.
Maybe ```app.get('/cyberchef.html', [cyberchefCspHeader], express.static(``` and the missingResource still stays on the end?

PR #1979, opened 2022-08-19
Fix for spelling-mistake when using message from exception in parliament

p4fg:
Quick fix for a minor issue for logging update-errors. The attribute for message was mis-spelled.

## License


awick:
thanks!

PR #1983, opened 2022-08-19
Updated tags endpoints to use `checkCookieToken`

holmesec:
The paths `/api/sessions/addtags` and `/api/sessions/removetags` uses the `checkHeaderToken` middelware instead of `checkCookieToken` which most state changing endpoints uses. `checkCookieToken` requires that the `x-arkime-cookie` header is present, which makes it secure from CSRF attacks, but `checkHeaderToken` does not.

This PR removes the `checkHeaderToken` function and updates the tags endpoints to use `checkCookieToken` instead.

**Note:** This will break the the API from "being used from places other than just the UI" as specified in the in-line comment as the reason for this functions existence. I'm not sure if there's still a use-case for this, but if this still needs to be supported, we need to come up with another soltuion.

## License

### DIFF
--- a/viewer/viewer.js
+++ b/viewer/viewer.js
@@ -493,15 +493,6 @@ function checkCookieToken (req, res, next) {
   return next();
 }
 
-// use for APIs that can be used from places other than just the UI
-function checkHeaderToken (req, res, next) {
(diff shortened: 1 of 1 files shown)


awick:
Works as designed

PR #1986, opened 2022-08-19
Add Content-Type: application/json to /api/fields endpoint

renniepak:
Add Content-Type: application/json to /api/fields endpoint

The /api/fields endpoint responded with a Content-Type: text/html. This would enable potential XSS attacks. By setting the correct content type for the presented data (application/json), this issue is mitigated.

## License


awick:
Would work but went with ArkimeUtil.noCacheJson like most other locations.

PR #1991, opened 2022-08-22
improve roles dialog/display, forceExpression/QTL inherited

awick:
Hide fields not needed for roles.
expressions from roles are combined
Use the lowest QTL from ourselves and roles


TobySalusky:
Looks good overall. My only comment would be that it might be worth adding a method to turn the `!data.item.userId.startsWith('role:')`s on Users.vue into something like `isUser(data.item)` or `!isRole(data.item)`

PR #1993, opened 2022-08-23
Add proper encoding to prevent HTML injection

renniepak:
Add proper encoding to prevent HTML injection

## License


awick:
Fixed in #1978

PR #2102, opened 2022-12-17
Type and range of bucket

Azyka:
An integer underflow was triggered when the arkime capture module was processing a pcap file. The crafted pcap file manipulates the value of  `long int lastPacketSecs`, and causes the underflow of `int bucket` in the following processing. Also, the range of `bucket` ought to be limitted to avoid the out of bound reference.

### DIFF
--- a/capture/pq.c
+++ b/capture/pq.c
@@ -122,7 +122,7 @@ void moloch_pq_upsert(MolochPQ_t *pq, MolochSession_t *session, int timeout, voi
     MolochPQItem_t *item;
     HASH_FIND(pqh_, (pq->keys[session->thread]), session->sessionId, item);
     if (item) {
-        long int bucket = item->expire - pq->bucket0[session->thread];
+        int bucket = item->expire - pq->bucket0[session->thread];
         if (bucket < 0) bucket = 0;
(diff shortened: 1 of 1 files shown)


awick:
Hi can you attach the pcap or slack it to me.

Azyka:
> Hi can you attach the pcap or slack it to me.
Sure, I will slack the samples to you.

awick:
Thanks for the pcaps, looks like I still get sanitize errors with the current PR, so I'll look some more. I think the general issue is that arkime has Y2038 issues that I need to fix. :(

Azyka:
It really weird, as I haven't catch the entire structrue of arkime. I also wonder how this change takes effects.

awick:
it looks like libpcap has a Y2038 issue (or design choice) that I'm overriding in PR #2103, once i merge that we should retest what is needed next.

Azyka:
get it

awick:
So I took a different approach with #2104 although i don't know it will work if the timestamp is maxtime.

Azyka:
I tested different timestamp including the maxtime based on my previous pcap, and it worked fine without any error.
Impressive change :)

PR #2108, opened 2022-12-21
Bugfix for SLL layers

JSCU-CNI:
Other layers than IPv4 can be within the encapsulation of SLL. This can crash the front-end viewer of Arkime. The commit in this merge request fixes this by making sure more encapsulation is correctly interpreted.

## License


awick:
thanks

Issue #2096, opened 2022-12-14
cont3xt - allow raw edit on link group create page


awick:
Allow a raw mode on the initial create link group in cont3xt. Use case is when I want to copy a link group from 1 system to another I have to first create a fake entry before i can do a raw edit.

PR #2144, opened 2023-01-13
raw create link groups

31453:
put raw edit in linkgroupform so the code is reused

fixes [#2096](https://github.com/arkime/arkime/issues/2096)


awick:
lgtm

PR #2146, opened 2023-01-17
Divided ssh.version into ssh.versionDst and ssh.versionSrc

JSCU-CNI:
It would be helpful for aggregations to also split the source and destination SSH versions. `Hassh` and `HasshServer` are also divided within the session model. 

## License


awick:
Thanks! We should continue to set the old fields for now so we don't have a breaking change. I also don't think the capture or viewer tests will pass and need to be updated.

awick:
Are you going to be able to fix this or should I close?

PR #2156, opened 2023-01-24
Use passportjs for all auth in the right nonhacky way

awick:
Move all the auth stuff to passport strategies so we can add other strategies in the future.


31453 (approved):

31453:
Works for me ✨

PR #2157, opened 2023-01-25
OIDC passport auth

awick:
Implement oidc support for arkime/wise/viewer. Parliament isn't supported yet.

Here is what a keycloak config would look like. For a different system make sure you change the authUserIdField and what the template uses to get the user name.
```
userNameHeader=oidc
authDiscoverURL=<DISCOVER or ISSUER or WELLKNOWN URL>
authClientId=<CLIENTID>
authClientSecret=<CLIENTSECRET>
authUserIdField=preferred_username
authRedirectURIs=http://localhost:8123/arkime/auth/login/callback
userAutoCreateTmpl={"userId": "${this.preferred_username}", "userName": "${this.name}", "enabled": true, "webEnabled": true, "headerAuthEnabled": true, "emailSearch": true, "createEnabled": false, "removeEnabled": false, "packetSearch": true }
```


viewer/config.js
```diff
@@ -530,5 +532,13 @@ Auth.initialize({
   userAutoCreateTmpl: exports.get('userAutoCreateTmpl'),
   userAuthIps: exports.get('userAuthIps'),
   s2s: true,
-  s2sRegressionTests: !!exports.get('s2sRegressionTests')
+  s2sRegressionTests: !!exports.get('s2sRegressionTests'),
+  authConfig: {
+    httpRealm: exports.get('httpRealm', 'Moloch'),
```
31453, line 537: Arkime instead here? And in WISE and Cont3xt?

31453:
What is it going to take to also support Parliament? First we need to use Arkime users, right?

viewer/config.js
awick: The value can't change because it will break all the encrypted passwords. :( Let me make sure I didn't break wise/cont3xt. :)

awick:
Ya I was waiting on parliament because it seems to be in a state of flux around how users work and I didn't want to break anything ;)

viewer/config.js
awick: So for 5.0 I think maybe we should make a goal of Moloch->Arkime most places. Maybe for this we change the default to Arkime and if you want backwards compat you need to set to Moloch manually. I'll start a 5.0 list

31453:
Yeah, Parliament doesn't use Arkime users yet. It's on the list of things to do. There's a warning in the UI that the password stuff that is in there currently will be deprecated "soon" 🙂

awick:
I think I did hack parliament while you were out to use some users auth stuff, but not everything. :)

PR #2176, opened 2023-02-11
CA file support for OIDC auth

Didr:
The current OIDC auth flow (#2157) doesn't use `caCertFile` from the config.
This PR copies the result from `getCaTrustCerts` (https://github.com/arkime/arkime/blob/main/viewer/config.js#L199) into the OIDC auth flow and if defined, will use the provided CA to override the default CAs (Mozilla) used in nodejs.

Relevant section from the OIDC library: https://github.com/panva/node-openid-client/blob/main/docs/README.md#customizing-http-requests
Default behavior of "ca" option: https://nodejs.org/api/tls.html#tlscreatesecurecontextoptions

Yes.

Sadly, I couldn't get this to work. I've manually tested without any auth and with the OIDC auth. I've not tested the header auth.

## License


Didr:
I realize now that this'll break things if someone is using separate CAs for their OIDC and elastic/opensearch setup.

awick:
This seems fine, other than this is specific to viewer. Maybe should move this out of viewerUtils into ArkimeUtils?

Didr:
Still need to test cont3xt and WISE. I'll leave a message when that's done.

awick:
thanks!

PR #2182, opened 2023-02-15
Extract TTL from DNS A and AAAA answers

killerwhile:
Extract TTL from DNS A and AAAA answers.
Only A and AAAA answers are extracted here, because the way ips and hostnames are stored in the SPI today (in a hash, i.e. dedup'ed) makes hard the TTL storage to match the other DNS answers.
Heuristic is that IPs are unique in the DNS record answer, and ttl offsets matches ips offsets.

## License


killerwhile:
@awick I'm not fully convinced about the way I'm storing TTL for responses right now. Could you please chime in and provide your insights?
The main problem is that hosts and ips are de-dup'ed (via string_hash structures), and I don't see how to express the mapping between the response and the TTL.

capture/parsers/dns.c
```diff
@@ -614,6 +622,12 @@ void moloch_parser_init()
         MOLOCH_FIELD_TYPE_STR_HASH,  MOLOCH_FIELD_FLAG_CNT,
         (char *)NULL);
 
+     ttlField = moloch_field_define("dns", "integer",
+         "dns.ttl", "Query TTL", "dns.ttl",
```
awick, line 626: Is it the Query or Answer TTL?

killerwhile: Answer, thanks!

awick:
I would want to understand the use case. Currently with few exceptions (certs) we don't group multiple transactions inside of a single session which is what it seems like you want here. That would take some arkime design changes that we should talk about. We do parallel arrays in some places, but that would require the other fields to need to change.

awick:
closing for now, let me know when you wanna chat

killerwhile:
Yes sure! As you have heard, we're working on another DNS parser implementation which is superseding this MR. Thanks!

PR #2195, opened 2023-02-25
Fix incorrect WISE auth method for OIDC

Didr:
wiseService.js had an error in it's `userNameHeader` check which made OIDC authentication impossible. "digest" was referenced twice.

## License


awick:
thanks

PR #2209, opened 2023-03-14
Added signal handler for flushing sessions when needed

JSCU-CNI:
When having an Arkime capture process running for a long time it would be helpful to flush all sessions when needed. 

By adding this signal handler it can be done without restarting the capture process.

## License

### DIFF
--- a/capture/main.c
+++ b/capture/main.c
@@ -342,6 +342,11 @@ void reload(int UNUSED(sig))
     moloch_plugins_reload();
 }
 /******************************************************************************/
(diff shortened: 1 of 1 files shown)


awick:
Why is this needed? I don't think it compiles right now.

JSCU-CNI:
A new commit has been forced pushed over the old commit. This resolves the typo.

We are currently using this small feature because of the way we receive and process pcap data.

awick:
Can you provide more details? Flushing the sessions can cause a big pause and other issues.

PR #2213, opened 2023-03-15
Allow role forced expression without user forced expression

### LINKED ISSUE #2212: Forced Expression fails via Role
Assigning a Forced Expression via role and not directly to the user results in an error.

Steps to reproduce:

1. Create role with Forced Expression:  node == capture1
2. Assign role to a user that does not have Forced Expression set

Error in viewer:

ERROR - Forced expression ( && (node == capture1)) doesn't compile - Error: Parse error on line 1:
&& (node == capture1

**Elasticsearch version**: 7.10.2

**Arkime/Moloch version**:  v4.2.0-12-gbf7761cf @ 2023/03/15 14:38:42

**OS name and version**: Ubuntu 22.04.1 arm64

**How was Arkime/Moloch built/installed:** easybutton

divinehawk:
Assigning a Forced Expression via role results in an error when there is no Forced Expression already set on the user.

The current code begins the first Forced Expression with "&&".

Fixes #2212 

## License

### DIFF
--- a/common/user.js
+++ b/common/user.js
@@ -980,8 +980,11 @@ class User {
       allRoles.add(r);
 
       if (role.expression && role.expression.trim().length > 0) {
-        if (!this.#allExpression) { this.#allExpression = ''; }
-        this.#allExpression += ' && (' + role.expression.trim() + ')';
+        if (!this.#allExpression) {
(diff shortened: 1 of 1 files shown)


awick:
thanks

PR #2216, opened 2023-03-17
Do not allow usersAdmin to delete superAdmin

dhakalananda:

The application does not allow a usersAdmin to create and update a superAdmin but is allowing to delete them. This means an escalated privilege for usersAdmin. Only a superAdmin should be allowed to delete another superAdmin as per the permission hierarchy.

HackerOne report: 1909898

## License


awick:
I merged this after changes, not sure why it didn't close this

Issue #867, opened 2018-06-15
gh0st protocol detection


florin2528:
**Elasticsearch version**:
 "version" : {
    "number" : "5.6.8",
    "build_hash" : "688ecce",
    "build_date" : "2018-02-16T16:46:30.010Z",
    "build_snapshot" : false,
    "lucene_version" : "6.6.1"
  },

**Moloch version**:
moloch-capture 0.50.0/v0.50.0 
glib2: 2.52.3
libpcap: libpcap version 1.8.1
curl: libcurl/7.55.0 OpenSSL/1.0.2k zlib/1.2.8
pcre: 8.31 2012-07-06
yara: 3.6.3

**OS name and version**:
Linux 4.9.0-5-amd64 #1 SMP Debian 4.9.65-3+deb9u2 (2018-01-04) x86_64 GNU/Linux

**How was Moloch built/installed:** (rpm, deb, singlehost built, ...)
deb

**Provide logs, stack traces and steps to reproduce**:
It has been notice that Moloch detects gh0st protocol, but with an investigation regarding this we found out that actually it's matching this protocol on DCE/RPC packets based on the HEX below:
payload8.dst.hex == 05000c0710000000

Can I kindly ask you to confirm this. 

Kind Regards,
Florin

awick:
I'm confused what you want to confirm?  Sounds like its doing what you said.

florin2528:
But the traffic which Moloch is capturing as gh0st protocol actually it's not.

awick:
Many of the protocol labels are just best moloch can determine

florin2528:
thank you for clarifying this issue.

PR #2225, opened 2023-03-30
Update misc.c Gh0st parser identification to reduce false positives

ObsidianKnife:
Gh0st parser has false positives by not looking at both bytes of the zlib header x789c at offsets 13 and 14

Problem:
gh0st RAT traffic is currently not identifying the whole zlib header. This has resulted in hundreds of false identifications of NTLM  sessions as gh0st sessions.  later in this same code block (line 53) an alternate detection is made that includes the full zlib header. This update is to make sure both methods check data[14].

Issue #867 was closed dismissing a similar issue with false identification of DCE/RPC traffic as gh0st
Sources for understanding gh0st use of zlib header:
    https://chopshop.readthedocs.io/en/latest/modules/gh0st_decode.html

Proposed fix is not tested because I can not set that up in my current environment. 

## License

### DIFF
--- a/capture/parsers/misc.c
+++ b/capture/parsers/misc.c
@@ -44,7 +44,7 @@ LOCAL void gh0st_classify(MolochSession_t *session, const unsigned char *data, i
     if (len < 15)
         return;
 
-    if (data[13] == 0x78 &&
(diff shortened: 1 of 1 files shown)


awick:
thanks

PR #2278, opened 2023-06-02
Update config.c  - Fix buffer overflow

baptistapedro:
#### Description
Fix integer overflow on config prefix.

#### Bug Analysis
```
 config.prefix           = moloch_config_str(keyfile, "prefix", "arkime_");
    int len = strlen(config.prefix);
    if (len > 0 && config.prefix[len - 1] != '_') {
        char *tmp  = malloc(len + 2);       // <-------------- [if len  is INT_MAX + 2 it will wrap to a negative value]
        memcpy(tmp, config.prefix, len);   //   <-------------- [buffer overflow]
        tmp[len] = '_';
        tmp[len+1] = 0;
        g_free(config.prefix);
        config.prefix = tmp;
    }
```

### DIFF
--- a/capture/config.c
+++ b/capture/config.c
@@ -23,6 +23,7 @@
 #include <fcntl.h>
 #include <inttypes.h>
 #include <errno.h>
+#include <limits.h>
 
 extern MolochConfig_t        config;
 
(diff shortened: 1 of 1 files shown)


awick:
Thanks, although I find this kind of silly to make sure the prefix is less than 2G. Since the max length of an elasticsearch/opensearch index is 255, a better fix would be to just check for that and LOGEXIT if too big.

baptistapedro:
Hello @awick, I am not that familiar with elasticsearch and my setup is not working properly.
But if you run arkime command line with such big config file/config prefix it wont crash?

awick:
I'm sure it will crash, but no one will make a prefix that is 2G since it won't work if its greater than ~250. I'll add a useful check.

baptistapedro:
@awick,
I've submitted this PR as part of the bounty program on hackerone. If such config file is loaded by any user it may lead to remote code execution. As you said it is proven to be a vulnerability though unlikely that a user will use such a big prefix by himself.

The submission is still open in hackerone.

awick:
I'm not sure how it will be a RCE since its a local config file. It will most likely be considered out of scope, although I don't determine that. See first item on https://arkime.com/security

PR #2279, opened 2023-06-06
fix confirm password text

TobySalusky:
* changed placeholder and label for new password confirmation input


31453 (approved):
Good catch 👀

PR #2284, opened 2023-06-12
Bruteforce resistant key derivation scheme

ahmed-tabib:
The cookie authentication mechanism uses a weak key derivation scheme from the password. It simply hashes the password with SHA-256. This would allow any user with an auth token to launch an offline bruteforce attack on the password. I tested the attack on my personal laptop and achieved speeds of ~110 MH/s.
The solution is simple, I used the crypto.pbkdf2Sync() function instead of SHA with 1,000,000 iterations. This adds a delay of ~100-250ms on startup, but once the key is generated you wouldn't need to do it again.
Also, I added a block that checks when passstore decryption fails, and tries the old key derivation scheme. This probably isn't the right way, I tried to figure out a way to re-encrypt the passstore but I honestly have no idea where to start.

## License

### DIFF
--- a/common/auth.js
+++ b/common/auth.js
@@ -97,9 +97,9 @@ class Auth {
     Auth.#userNameHeader = options.userNameHeader;
     Auth.#passwordSecretSection = options.passwordSecretSection ?? 'default';
     Auth.passwordSecret = options.passwordSecret ?? 'password';
-    Auth.passwordSecret256 = crypto.createHash('sha256').update(Auth.passwordSecret).digest();
+    Auth.passwordSecret256 = crypto.pbkdf2Sync(Auth.passwordSecret, 'salt', 1000000, 32, 'sha256');
     if (options.serverSecret) {
-      Auth.#serverSecret256 = crypto.createHash('sha256').update(options.serverSecret).digest();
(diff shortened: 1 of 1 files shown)


awick:
Thanks for the PR. 

Can you describe the attack in a little more detail? Since you mentioned the cookie, that is created by Auth.obj2auth, and I thought the iv would protect from the style you describe.

ahmed-tabib:
Yeah absolutely. The way I tested it I used the hmac, since it uses the same key. The signature is calculated as s = hmac(sha256(password), iv+encrypted_content)). Since Hmac is computationally equivalent to 2 passes of sha-256, this allows us to try many many passwords very quickly, like I stated above, having a message and it's hmac, you can try around 110 million passwords per second. I created my own hashcat module for this. I can share it with you if you'd like to run it yourself.

awick:
Sure would you mind email it to arkime.security@yahooinc.com or DM on slack.

ahmed-tabib:
I sent an email with some instructions on building the plugin and running it.

awick:
Thanks, I'm going to be making the changes over on https://github.com/arkime/arkime/pull/2287/files if you can take a look

awick:
thank you for your help

PR #2299, opened 2023-06-21
added authMode and basic auth

awick:
* new setting "authMode" is used everywhere instead of userName Header
* authMode should default to digest everywhere now
* added basic auth support


31453 (approved):
LGTM. 5.0 already has lots of breaking changes! 🙀

Issue #2281, opened 2023-06-08
FR: Allow for export of the User's table as CSV from the Users page


amaddalena:
Please setup an icon/button to all an export of the Users page to CSV (same table as displayed on the Users page) 
- It would be useful when going through to remove old accounts, get a list of disabled accounts, audit stuffs, etc. TIA!

🦉 🦉 🦉 🦉 🦉 🦉 🦉 🦉

awick:
Will be part of 5.0

PR #2300, opened 2023-06-21
backend of /api/users.csv with tests #2281

awick:
Still needs @31453 UI magic to surface

### DIFF
--- a/common/arkimeUtil.js
+++ b/common/arkimeUtil.js
@@ -72,6 +72,15 @@ class ArkimeUtil {
   }
 
   // ----------------------------------------------------------------------------
+  static noCache (req, res, ct) {
+    res.header('Cache-Control', 'no-cache, private, no-store, must-revalidate, max-stale=0, post-check=0, pre-check=0');
+    if (ct) {
+      res.setHeader('Content-Type', ct);
+      res.header('X-Content-Type-Options', 'nosniff');
+    }
+  };
+
+  // ----------------------------------------------------------------------------
   /**
    * Is there any intersection between the two arrays
    */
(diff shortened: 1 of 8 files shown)


common/arkimeUtil.js
```diff
@@ -71,6 +71,15 @@ class ArkimeUtil {
     return next();
   }
 
+  // ----------------------------------------------------------------------------
+  static noCache (req, res, ct) {
```
31453, line 75: Isn't this already in ViewerUtils? We should remove it from there and use this everywhere

31453:
The UI for this would just be a button on the Users page to download the csv (passing in the query params)?

common/user.js
```diff
@@ -510,6 +510,69 @@ class User {
...
+    if (Array.isArray(req.body.start) || Array.isArray(req.body.length)) {
+      return res.send('Invalid');
+    }
+
+    const query = {
+      from: parseInt(req.body.start) || 0,
+      size: parseInt(req.body.length) || 10000
+    };
+
+    if (ArkimeUtil.isString(req.body.filter)) {
```
31453, line 532: We do this in `apiGetUsers` too. Should we abstract this to a new function?

common/arkimeUtil.js
awick: yep, was going to do another PR for that

awick:
> The UI for this would just be a button on the Users page to download the csv (passing in the query params)?

yep, whenever you have time

common/user.js
awick: ok, will do

awick:
LGTM

Issue #2306, opened 2023-06-26
arkimecapture.service fails to start on Rocky Linux 8 due to a dependency on elasticsearch instead of elasticsearch.service


alfonsojon:
**Describe the bug**
arkimecapture.service has been failing to start on our Rocky Linux 8 instance after a reboot. I checked logs and found the following error:
`Failed to add dependency on elasticsearch, ignoring: Invalid argument`. I was able to manually start `arkimecapture.service` once elasticsearch was running, but it would fail on reboot or a clean boot. This error combined with the behavior lead me to believe that arkimecapture was trying to start before elasticsearch. Going based off the error provided, it told me that elasticsearch was not a valid dependency so I checked & found it should actually be `elasticsearch.service`. I edited the file `/etc/systemd/system/arkimecapture.service` to rely on `elasticsearch.service` instead of `elasticsearch` and this fixed the issue for me. I would like to see if this can be corrected in the source code. Thank you!

**Elasticsearch version**:
8

**Arkime/Moloch version**:
4.1.0

**OS name and version**:
Rocky Linux 8.8

**How was Arkime/Moloch built/installed:** (rpm, deb, easybutton, ...)
installed using instructions for RHEL 8

**Provide logs, stack traces and steps to reproduce**:
details listed above

awick:
merged

PR #2307, opened 2023-06-26
Fix issue with arkimecapture.service starting before elasticsearch.service

alfonsojon:
correct issue with failing to start after elasticsearch on Rocky Linux 8

arkimecapture.service has been failing to start on our Rocky Linux 8 instance after a reboot. I checked logs and found the following error:
`Failed to add dependency on elasticsearch, ignoring: Invalid argument`. I was able to manually start `arkimecapture.service` once elasticsearch was running, but it would fail on reboot or a clean boot. This error combined with the behavior lead me to believe that arkimecapture was trying to start before elasticsearch. Going based off the error provided, it told me that elasticsearch was not a valid dependency so I checked & found it should actually be `elasticsearch.service`. I updated the source of arkimecapture.systemd.service to correct this issue upstream.

https://github.com/arkime/arkime/issues/2306

I did

Yes

## License

### DIFF
--- a/release/arkimecapture.systemd.service
+++ b/release/arkimecapture.systemd.service
@@ -1,6 +1,6 @@
 [Unit]
 Description=Arkime Capture
-After=network.target opensearch elasticsearch
+After=network.target opensearch elasticsearch.service
 
 [Service]
 Type=simple
--- a/release/arkimecapture.systemd.service
+++ b/release/arkimecapture.systemd.service
@@ -1,6 +1,6 @@
 [Unit]
 Description=Arkime Capture
-After=network.target opensearch elasticsearch.service
+After=network.target opensearch.service elasticsearch.service
 
 [Service]
 Type=simple


alfonsojon:
the same may need to be done with opensearch, but we do not use opensearch in our environment & I do not have the ability to test

philhagen:
In looking at this further, it looks like CentOS7 might _require_ the dependency to be listed as `elasticsearch.service`.  I can fix this locally for my requirements, and to be honest, I'm not even sure it's required since I have other installations that seem to operate as expected with just `elasticsearch`.

Also this is not likely something that requires handling within the codebase, but I wanted to make note of it for others who may encounter the problem.

awick:
ugh, looks like i fixed this for capture but not viewer in newer 4.x. Is viewer causing the issue?

philhagen:
yes, that's what my (admittedly incomplete) testing showed - just viewer was affected.

PR #2317, opened 2023-07-05
prevent user history view in demo mode

dhakalananda:
Demo mode prevents history viewing but the specific API is leaking history. This is a security issue.


awick:
Thanks for pointing this out, but I'm not sure its a security issue in demo mode.

dhakalananda:
Please check out #2320.

PR #2319, opened 2023-07-05
fix access control in remote hunt endpoint

dhakalananda:
User should have packetSearch permission in order to do anything related to hunts but `/api/hunt/:nodeName/:huntId/remote/:sessionId` does not have any protection.

This is a fix for that security issue.

### DIFF
--- a/viewer/viewer.js
+++ b/viewer/viewer.js
@@ -1786,7 +1786,7 @@ app.delete( // remove users from hunt endpoint
 
 app.get( // remote hunt endpoint
   ['/api/hunt/:nodeName/:huntId/remote/:sessionId', '/:nodeName/hunt/:huntId/remote/:sessionId'],
-  [ArkimeUtil.noCacheJson],
+  [ArkimeUtil.noCacheJson], User.checkPermissions(['packetSearch'],
   HuntAPIs.remoteHunt
 );
 


awick:
Thanks for pointing this out, we are going to fix it in 5.0 using a different method. Since you have to know both a huntId and sessionId we consider this out of scope for security.

dhakalananda:
Hi,

Getting sessionId is not a big deal for a user in the organization. A user can get huntId if they have previous access to hunt or through some other means.

If it is going to be fixed, this should be considered a valid one. That is how it works :)

awick:
All I meant was there is a difference between a security issue, a bug, and an improvement. Could you describe how you would use this to be a SECURITY issue?

dhakalananda:
Again, as I said, a user having "Disable Hunt" permission should not be able to view anything related to the hunt. Just being able to make an API call makes it vulnerable to Improper Access Control. While this might not be a high-severity issue, this IS a low-severity issue because it is affecting the confidentiality of the org and we are able to perform actions against the design of the application.

awick:
Sorry, in this case I'm really looking for the scenario since we consider anything with a sessionId or huntId out of scope. (if you already know the sessionId that means you saw the session, and if you know the huntId you know the hunt.) The only data "viewed" is it the session matched the hunt.

PR #2320, opened 2023-07-05
fix history/cron in demo mode

dhakalananda:
A sinlge PR for #2317 and #2318

### DIFF
--- a/viewer/viewer.js
+++ b/viewer/viewer.js
@@ -1094,11 +1094,11 @@ function expireCheckAll () {
 // APIs disabled in demoMode, needs to be before real callbacks
 if (Config.get('demoMode', false)) {
   console.log('WARNING - Starting in demo mode, some APIs disabled');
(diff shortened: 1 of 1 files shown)


dhakalananda:
The reason this is a security issue is because: A demo website is used by multiple users so having the user's history public is an information leak as it contains requests to all the endpoints and all. Having CRUD access to periodic queries should also be forbidden to avoid creating thousands of cron jobs in the app.

awick:
It is a demo instance and we state that everyone can view everyone's data.

history - I do not think is a security issue, agree we should fix
cron - I do see that as potential DOS issue

Unfortunately this PR has a number of issues, I'll be making a new one, thanks for pointing out these issues

dhakalananda:
I have submitted a report on HackerOne #2051313. Please triage it on the platform as well.

awick:
Ok thanks, I don't have access but the BB folks can reply.

PR #2323, opened 2023-07-06
fix prototype pollution

dhakalananda:
There is a prototype pollution vulnerability that will DOS users if a request is sent to `POST /api/user/settings`.

Blacklisting` __proto__` will fix the issue.

HackerOne report: #2053341

### DIFF
--- a/cont3xt/cont3xt.js
+++ b/cont3xt/cont3xt.js
@@ -315,7 +315,11 @@ function apiPutSettings (req, res, next) {
     }
 
     if (user.cont3xt === undefined) { user.cont3xt = {}; }
-
+    
+    if (req.body.hasOwnProperty('__proto__')) {
+      return res.send({ success: false, text: '__proto__ not allowed' });
+    }
+    
     if (req.body?.settings) {
       user.cont3xt.settings = req.body.settings;
       save = true;


awick:
The PR looks like its for cont3xt but the URL provided is for viewer, so not sure which this is for? Are you putting more details of how to reproduce in the H1 report?

If this is a general issue with prototype and json parser we will probably want to look at a middleware fix.

dhakalananda:
Oops, sorry. The commit is correct and the issue is on cont3xt. The endpoint is `PUT /api/settings` in 3218 port. I have provided all the information in the HackerOne report.

> If this is a general issue with prototype and json parser we will probably want to look at a middleware fix.

I'd definitely recommend it because more codebase changes might lead to more of these issues so having a site-wide fix would be better.

awick:
Ok, I don't have access to the H1 reports so will have to wait. I also think we've already switched to a json parser that doesn't have this issue anymore with proto but once I have a test case I can confirm.

awick:
Ok was able to reproduce this one, basically there wasn't any actually pollution happening, but the elasticsearch library won't process json that has certain keywords.

Anyway I'm going to close this PR since it won't actually fix the issue and opened #2324. In the future issues might be better than PRs for tracking these issues. (But feel free to do a PR on that issue.)

PR #2325, opened 2023-07-07
Update json body parser to make sure secure-json-parse is happy with it

### LINKED ISSUE #2324: The nodejs elasticsearch library won't load json with certain keywords
The elasticsearch js library is using secure-json-parse to parse JSON with prevents it from load json with certain keywords. We should make sure we don't save anything with those keywords, might need to switch to using secure-json-parse with the json-parser library.

awick:
Fixes #2324 by not accepting any post/put json payloads that doesn't pass secure-json-parse parsing. Unfortunately this causes a double parse since we don't actually use the results of secure-json-parse :(

We also create the jsonParser in one place so all tools share how it is created

## License


TobySalusky (approved):
Looks good 👍

PR #2331, opened 2023-07-17
Improve usersAdmin permission checks

awick:
Now a usersAdmin user can only change passwords on users it has the same Admin roles as.

userA is usersAdmin/cont3xtAdmin
userB is cont3xtAdmin
userC is cont3xtAdmin/arkimeAdmin

userA can change userB password but not userC because it doesn't have arkimeAdmin

## License


31453 (approved):
LGTM

PR #2334, opened 2023-07-19
create overview index before defaults

TobySalusky:
* await Db initialize to avoid race condition where default overviews are created before overview index

### DIFF
--- a/cont3xt/db.js
+++ b/cont3xt/db.js
@@ -364,7 +364,14 @@ class DbESImplementation {
           iType: { type: 'keyword' },
           viewRoles: { type: 'keyword' },
           editRoles: { type: 'keyword' },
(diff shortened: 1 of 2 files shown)


cont3xt/db.js
```diff
@@ -355,6 +357,15 @@ class DbESImplementation {
     await this.client.indices.putMapping({
       index: 'cont3xt_overviews',
       body: {
+        properties: {
+          creator: { type: 'keyword' },
```
TobySalusky, line 361: I figured I would put mappings for the fields we know of explicitly—maybe that's better? If we're planning on adding 'type', should we be adding `type: { type: 'keyword' }` now, or once that's implemented, perhaps?

awick: sure

PR #2338, opened 2023-07-21
Custom overview fields

TobySalusky:
* custom fields can be created by selecting Custom at the bottom of the options for field in an Overview config.
* the custom object (which is the same format as index.js card fields) can be edited in the local raw edit box by expanding or in the full json raw edit page
* type:linked and type:custom to differentiate field types
* verify custom object on backend and store as a json string
* updated cont3xt overview tests
* improve warnings on OverviewCard for certain bad field configs (linked: typo in from/field name; custom: missing path)
* normalizeCardField shared between frontend/backend

### DIFF
--- a/cont3xt/overview.js
+++ b/cont3xt/overview.js
@@ -76,12 +76,87 @@ class Overview {
         Db.putOverview(iType, createDefaultForIType(iType)).then(() => {
 
         }).catch((err) => {
-          console.log(`ERROR - failed to create default overview for ${iType} iType... skipping`, err);
+          console.log(`WARNING - failed to create default overview for ${iType} iType... skipping`, err);
         });
       }).catch((err) => {
-        console.log(`ERROR - failed to check existence of ${iType} iType default overview... skipping:`, err);
+        console.log(`WARNING - failed to check existence of ${iType} iType default overview... skipping:`, err);
       });
     }
+  }
+
+  // Verify the given overview, returns error { msg: string } on failure, { custom } otherwise
+  static verifyOverviewCustomField (custom) {
+    // non-empty strings are always valid (they're applied as both label and field)
+    if (ArkimeUtil.isString(custom)) { return { custom }; }
+
+    if (custom == null || typeof custom !== 'object') {
+      return { msg: 'Custom field must be a string or object' };
+    }
+    custom = (
+      ({ // only allow these properties in custom fields
+        // eslint-disable-next-line no-shadow
+        label, field, type, path, fields, defang, pivot, join, fieldRoot, fieldRootPath, filterEmpty, defaultSortField, defaultSortDirection, altText, noSearch, postProcess
+      }) => ({
+        label, field, type, path, fields, defang, pivot, join, fieldRoot, fieldRootPath, filterEmpty, defaultSortField, defaultSortDirection, altText, noSearch, postProcess
+      })
+    )(custom);
+
+    if (custom.label != null && !ArkimeUtil.isString(custom.label)) {
+      return { msg: 'Custom label must be a string when present' };
+    }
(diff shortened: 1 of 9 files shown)


cont3xt/overview.js
```diff
@@ -76,14 +76,89 @@ class Overview {
...
         });
       }).catch((err) => {
-        console.log(`ERROR - failed to check existence of ${iType} iType default overview... skipping:`, err);
+        console.log(`WARNING - failed to check existence of ${iType} iType default overview... skipping:`, err);
       });
     }
   }
 
+  // Verify the given overview, returns error { msg: string } on failure, { custom } otherwise
+  static verifyOverviewCustomField (custom) {
```
TobySalusky, line 88: this and similar verifiers seem pretty boiler-platey to me. Out of curiousity, have we ever considered schema-based validation (like zod, yup, etc)? We're pretty much writing our own z.object.safeParse (returns a clean validated object or an error object) from zod here, so I wonder if that might be preferred?—though I suppose that would mean a new outside dependency, which we maybe want to avoid?

31453: I have not considered this. I agree that this could be "outsourced" to another tool, however I'd have to investigate whether this is worth the time/effort. I think this is something to add to our list of things to improve upon, but not something we want to do right now.

cont3xt/overview.js
```diff
@@ -246,6 +346,20 @@ class Overview {
...
+        // take best guess at type in case it is missing
+        field.type ??= field.custom == null ? 'linked' : 'custom';
+
+        // custom is stored as a json string, so we parse it before sending back
+        if (field.type === 'custom') {
+          try {
+            field.custom = JSON.parse(field.custom);
+          } catch {
+            // we'd rather not crash the service if the custom field is somehow invalid
+            field.custom = '_cont3xt_INVALID_CUSTOM_FIELD';
```
31453, line 359: Do we want to kill the user's custom field json or send an error for them to correct their json?

TobySalusky: hmm, I'm not quite sure what the best choice is.
I'm pretty sure this should already be technically impossible since our verifier should be ensuring that our json is valid (and thus parseable). 
I just thought I would add this in case something changes or goes terribly wrong down the road.
We could just fully remove the field I guess. Alternatively an error for the user is also an option, but I think that might be complicating the response and it's handling even more for a result that theoretically shouldn't occur in the first place.
The reason I like this approach is that it still lets you know that something is wrong if you dig into it, without us needing to add some sort of type:error or consider the case where custom doesn't exist (because it really always should).
Maybe none of this is necessary at all! I just find JSON parsing on the server a little scary

31453: I agree, JSON parsing is always a little scary 😱 Since we previously ensure the JSON is valid (I missed that), then this should be fine! I like that you're catching this just in case (even though it is unlikely to get in this catch).

Thank you for helping me understand!

cont3xt/overview.js
TobySalusky: Yea, I just thought it might be a good consideration for the future—seeing as we've been doing more of this sort of validation lately. I personally like the more declarative style of zod, especially the way you can mark fields optional versus required and handle properties that might be different types—plus we wouldn't need to use the funny immediately-invoked lambda anymore 😆.

31453: Agreed. I'll put it on my list of things to look in to! I think it could be beneficial.

31453 (approved):

cont3xt/overview.js
awick: I had looked at some before but it seemed like you were just programming them in another DSL, but probably would be a good idea and less mistakes. :)

PR #2341, opened 2023-07-24
a user cannot be added as a role

31453:


common/user.js
```diff
@@ -617,6 +617,10 @@ class User {
 
     req.body.roles ??= [];
 
+    if (!User.hasValidRoles(req.body.roles)) {
```
31453, line 620: Should we error out like this or should we just silently remove the "bad" roles and still update/create the user?

TobySalusky: I like the explicit approach of erroring, since it would allow the requester to realize their mistake instead of it going unnoticed. And I imagine this isn't a problem someone would run into on the UI (since we use RoleDropdown that only give valid roles as options), so it shouldn't hurt usability 🙂

TobySalusky:
Overall looks good—though I wonder if we might want to change the name `hasValidRoles` to something like `validateRoles`, since `hasValidRoles` sounds like it would be an instance method of the User class, but it is not.

common/user.js
awick: agree

awick (approved):
LGTM, although the cool kids would use roles.every :)

PR #2344, opened 2023-07-25
remove dasboard only mode from parliament

31453:
any Arkime user can view the Parliament dashboard
this is effectively dasboardOnly mode


awick (changes requested):
Don't we still need dashboardOnly for 4.x if you don't want to setup auth? Or does it work in anonymous mode? I might still support the dashboardOnly option and not error out so it isn't a breaking change. Maybe it should set anonymous mode?

If this is for 5.x then it should be on the dev5 branch

31453:
Going to implement this on dev5 when implementing commonAuth requirement.

PR #2354, opened 2023-08-01
UI reshuffle cont3xt

TobySalusky:
* reformulated cont3xt UI for bulk
* fixed CSS/behavior of integration panel
* improved page flows (scrolling) and navbar CSS
* red-highlight for active indicator, darken/lighten hovered indicators in tree
* reduced padding in indicator tree nodes
* link groups is a panel collapse-able in the UI or with shift-'>'


cont3xt/vueapp/src/components/integrations/IntegrationTidbit.vue
```diff
@@ -6,12 +6,12 @@
 
     </b-tooltip><!--  /tooltip  -->
 
-    <label v-if="labeled" :for="id" class="text-warning">
```
31453, line 9: Long values overflow the itype container, but not the panel. It would be nice if they got cut off by their immediate parent instead of the panel.

PR #2360, opened 2023-08-07
add dns rr https parsing

awick:
Support parsing DNS RR_HTTPS records

## License


awick:
@TNogisto approved

PR #2362, opened 2023-08-09
folding and bulk navigation

TobySalusky:
* hjkl to navigate indicator result tree
* scroll node to middle when navigating
* click '+ {n} hidden' to expand children
* shift - and shift + to collapse/expand all root nodes
* updated shortcut reminder tab
* updated shortcuts to use e.code instead of e.keyCode

### DIFF
--- a/cont3xt/vueapp/src/App.vue
+++ b/cont3xt/vueapp/src/App.vue
@@ -37,6 +37,18 @@
           <br>
           <code>'>'</code> - toggle the link group panel
           <br>
+          <code>'-'</code> - collapse all top-level indicator result tree nodes
+          <br>
+          <code>'+'</code> - expand all top-level indicator result tree nodes
+          <br>
+          <code>'h'</code> - collapse active indicator result tree node, or navigate left
+          <br>
+          <code>'j'</code> - navigate down in indicator result tree
+          <br>
+          <code>'k'</code> - navigate up in indicator result tree
+          <br>
+          <code>'l'</code> - expand active indicator result tree node, or navigate right
+          <br>
           <code>'shift + enter'</code> - issue search/refresh
           <br>
           <code>'esc'</code> - remove focus from any input and close this dialog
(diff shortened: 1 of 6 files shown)


cont3xt/vueapp/src/App.vue
```diff
@@ -37,6 +37,18 @@
           <br>
           <code>'>'</code> - toggle the link group panel
           <br>
+          <code>'-'</code> - collapse all top-level indicator result tree nodes
```
awick, line 40: this should be shift-'-' right? since an actually won't work :)

cont3xt/vueapp/src/App.vue
```diff
@@ -113,81 +125,114 @@ export default {
...
         break;
-      case 72: // h
+      case 'KeyH':
         // open help page if not on help page
         if (this.$route.name !== 'Help') {
           this.routeTo('/help');
         }
         break;
-      case 190: // . (seen as `>`, since shift is required)
+      case 'Period': // (seen as `>`, since shift is required)
```
awick, line 220: Thoughts about having < open the view panel to match?

cont3xt/vueapp/src/App.vue
TobySalusky: yes, so 'shift -', then? do we want 'shift +' to match that, or keep just '+' since you need shift for that anyways?

cont3xt/vueapp/src/App.vue
TobySalusky: Sure! is adding that to this PR fine?

cont3xt/vueapp/src/App.vue
awick: sure you can make it shift + so it looks pretty next to it :)

cont3xt/vueapp/src/App.vue
awick: yep

awick (approved):
LGTM when test pass

PR #2364, opened 2023-08-10
overview shortcut

TobySalusky:
* add overview shortcut as shift+O
* add padding to hotkey orange V for view dropdown so that the button doesn't shift/change width


awick (approved):

TobySalusky:
Is this good to merge? I don't remember main failing arch.
Though I wouldn't expect these changes to break it, so maybe it was something that changed with the Screwdriver job itself?

awick:
yes, gtg, arch is just sad and i haven't spent time on it. maybe someone who loves arch can look :)

PR #2366, opened 2023-08-11
overviews help page

TobySalusky:
* help page for overviews that gives basic info on what an overview is, how to change default, and how to configure fields
* incorporates descriptions.txt Card Description info
* examples from DNS records and some other existing integration fields


31453 (approved):
This is beautiful ✨ it puts all the other help pages to SHAME!

PR #2377, opened 2023-08-15
use common config for parliament

31453:
add parliament ini (file/port/cert/key needed here instead of command line args)
use upgrade to write ini based on commonAuth settings remove commonAuth from UI
remove commonAuth update API
update parliament help


CHANGELOG
```diff
@@ -27,6 +27,7 @@ Node Versions:
 
 NOTICE: Restart wiseService before capture when upgrading
 NOTICE: Cross-cluster Shortcuts require you to not restart all your viewers at once after upgrading
+NOTICE: Create a parliament config file before upgrading (see arkime.com/settings#parliament)
```
31453, line 30: @awick I'm going to great a parliament section on the settings page. Do you think this is a good NOTICE or is the BREAKING below enough?

parliament/parliament.js
```diff
@@ -2061,46 +1913,102 @@ app.use((req, res, next) => {
...
+      console.log('Debug Level', ArkimeConfig.debug);
+    }
+    internals.webBasePath = getConfig('parliament', 'webBasePath', '/');
+  } catch (err) {
+    console.log(err);
+    process.exit();
+  }
+
+  // ERROR OUT if there's no parliament config
+  if (!ArkimeConfig.getSection('parliament')) {
```
31453, line 1932: Right now I'm erroring out if there's not parliament config created. I _could, maybe, possibly, stupidly_ use the supplied command line args (port/key/cert/config) to populate the INI. But I feel like that's overcomplicating things? But it would be nice. I'm undecided... 🤷 💭

CHANGELOG
awick: I think its fine, the FAQ entry will probably be more important

parliament/parliament.js
awick: I think its fine to require a parliament section

Issue #2121, opened 2023-01-09
Bulk Cont3xt Search


31453:
Bulk search for multiple indicators. Indicators can be of different types.

- [x] Display multiple cards/tabs that are collapsible. 
- [x] Links get displayed based on the "active" indicator currently selected by the user.
- [x] There may be MANY indicators, so there should be a way to push items to the top/bottom by determining relevance/severity.
- [x] Put them into logical OR arrays and assign to variables like {arrayIps}, ${arrayEmail} for use in Cont3xt link groups for queries to Splunk or Arkime, etc.. #2380
- [x] Include information on the indicator results that there is something bad or that it has been whitelisted

31453:
Possible limit to the number of indicators (100)?
Search for all at once, progressive loading is already implemented but we may need to improve queuing.
Maybe some warning/flag to have the user check their enabled integrations.
Allow comma or newline separated list of indicators.
New endpoint or change endpoint that can separate the indicators.

**UI notes**:
Condensed/summary view that turns into the normal view when clicking on a specific indicator result.
You can navigate back to summary view, then to another indicator result.
Ability to scroll through all indicators.
Some top level tidbits for indicators.

31453:
@WilliamSalusky Question for the logical OR arrays: Would you want these to be only the indicators that you searched or any results expanded and returned? 

For example: If you searched for `10.0.0.1, example.com, 10.0.0.2` and example.com expanded  to include an IP of 10.0.0.3. Would you want the `${arrayIp}` to be `10.0.0.1, 10.0.0.2` or `10.0.0.1, 10.0.0.2, 10.0.0.3`?

31453:
Here's a possible solution that implements the latter option described above ☝️ #2380

WilliamSalusky:
I initially submitted this thinking that only the top level indicators as queried would be used to construct the per iType arrays, but if you are saying that an all inclusive list that includes children is easier, I'm happy to see child indicators included. I call that a bonus! 

And if that is the case, a future option already discussed that would disable generating child indicators would resolve that if it caused an issue for any user.


Taking this a bit further, the logical OR array for splunk could be extended to benefit other queryable destinations [i.e. Elastic/OpenSearch, Greylog, etc] through the support of multiple array formats.

```
${double_quoted_OR_arrayIp} = "10.0.0.1" OR "10.0.0.2" OR "10.0.0.3" ;# literal 'OR' separators
[...]

PR #2380, opened 2023-08-17
add vars for arrays of indicators in links

31453:


cont3xt/vueapp/src/components/links/LinkGroupCard.vue
```diff
@@ -420,6 +420,25 @@ export default {
...
+      // replaces ${arrayType} with a comma-separated list of all indicators of that type
+      // for example ${arrayIp} gets replaced with all the IPs in the results
+      for (const type of types) {
+        const matchStr = `\\$\\{array${type.charAt(0).toUpperCase() + type.slice(1)}\\}`;
+        if (url.match(new RegExp(matchStr, 'g'))) {
+          const array = [];
+          for (const key in this.$store.state.indicatorGraph) {
+            const indicator = this.$store.state.indicatorGraph[key];
+            if (indicator.indicator.itype === type) {
+              array.push(indicator.indicator.query);
```
awick, line 433: does this need dr.refang here?

cont3xt/vueapp/src/components/links/LinkGroupCard.vue
```diff
@@ -420,6 +420,25 @@ export default {
       return normalizedLinkGroup;
     },
     getUrl (url) {
+      const types = ['domain', 'ip', 'url', 'email', 'hash', 'phone', 'text'];
+      // replaces ${arrayType} with a comma-separated list of all indicators of that type
+      // for example ${arrayIp} gets replaced with all the IPs in the results
+      for (const type of types) {
+        const matchStr = `\\$\\{array${type.charAt(0).toUpperCase() + type.slice(1)}\\}`;
```
awick, line 427: you didn't ask, but why not just make the array ['arrayDomain', ...] :)

awick (approved):

cont3xt/vueapp/src/components/links/LinkGroupCard.vue
31453: because I use the type twice in the loop. to make the regex and then to determine if it's the correct type to add to the array

31453:
After talking with @awick, we could implement 3 varieties of this:
1. An array including ALL of each type
2. An array including TOP LEVEL of each type
3. An array including the CHILDREN of the selected indicator of each type.

Per my example in #2121 (if you searched for 10.0.0.1, example.com, 10.0.0.2 and example.com expanded to include an IP of 10.0.0.3.):
1. `10.0.0.1, 10.0.0.2, 10.0.0.3`
2. `10.0.0.1, 10.0.0.2`
3. If you selected example.com `10.0.0.3`

cont3xt/vueapp/src/components/links/LinkGroupCard.vue
31453: nope. the indicatorGraph object does not have defanged queries.

PR #2387, opened 2023-08-30
merge parliament/viewer notifiers

31453:
put parliament notifiers in users es
add alerts and on props to viewer notifiers
add common notifier component to share
use fetch instead of axios


common/notifier.js
```diff
@@ -12,6 +12,7 @@ class Notifier {
   static #debug;
   static #esclient;
   static #notifiersIndex;
+  static #defaultAlerts = { esRed: false, esDown: false, esDropped: false, outOfDate: false, noPackets: false };
```
31453, line 15: is there a better place to put this?

common/notifier.js
```diff
@@ -373,10 +387,12 @@ class Notifier {
         return res.serverError(404, 'Notifier not found');
       }
 
-      // only name, fields, roles, users can change
+      // only on, name, fields, roles, users, alerts can change
       notifier.name = req.body.name;
       notifier.roles = req.body.roles;
       notifier.fields = req.body.fields;
+      notifier.on = req.body.on || false;
```
awick, line 394: should this be !!req.body.on

common/vueapp/Notifiers.vue
```diff
@@ -238,17 +243,40 @@
             @selected-roles-updated="updateNotifierRoles"
             :display-text="notifier.roles && notifier.roles.length ? undefined : 'Share with roles'"
           /> <!-- /notifier sharing -->
+          <template v-if="parentApp === 'parliament' && notifier.alerts">
```
awick, line 246: should we still show this in viewer in case you edit a parliament notifier in viewer?

common/notifier.js
```diff
@@ -373,10 +387,12 @@ class Notifier {
...
+      notifier.alerts = req.body.alerts || Notifier.#defaultAlerts;
```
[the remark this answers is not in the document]
31453: I always forget about it 🧠

common/vueapp/Notifiers.vue
31453: My thought is that you can use notifiers for parliament or viewer. But when you're in parliament you want to configure which types of alerts you get from your parliament. And when you're in viewer you just want to add them to crons/hunts. I think it would be confusing to have the alerts show up for viewer since they don't pertain to viewer.

PR #2390, opened 2023-09-01
TLS Cert validity with negative values

mcgillowen:

Currently if TLS certificate expiry date is before the start validity date the `cert.validDays` and `cert.remainingDays` are not being saved in the SPI. This information though can be very useful in identifying forged TLS certs, so this PR removes the check for `cert.validDays` and `cert.remainingDays`.

We also added a greater granularity for these values with a granularity of seconds, this allows to identify certs that are generated with extremely short validity periods.

The `cert.remaining*` values are no longer calculated based on the current time of parsing, but rather the `firstPacket` timestamp, which IMO is a better choice since it is more consistent when potentially re-analysing a captured PCAP. If the value would be negative we set a value of `0`. This entailed also modifying the tests `--fix` script.

N/A

N/A

Yes

## License


awick:
Since this is changing what an existing field means I'm tempted to move this to 5.0 and the dev5 branch instead of main.

mcgillowen:
Ok, good point, I'll rebase my PR on the dev5 branch.

PR #2404, opened 2023-09-19
add editRoles to shortcuts, views, and queries

31453:
use middleware to check for resource access
added tooltip to dropdown for when roles are selected but there are 2 dropdowns
fix bad logic in hasRole function on client (how did that ever work!?)
remove roles columns from settings tables
update docs
fixed typos

### DIFF
--- a/common/vueapp/UserService.js
+++ b/common/vueapp/UserService.js
@@ -147,7 +147,7 @@ export default {
   /**
    * Determines whether a user has role to perform a specific task
    * @param {Object} user The user to check roles for
-   * @param {String} role The role in question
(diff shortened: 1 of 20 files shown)


common/vueapp/UserService.js
```diff
@@ -159,12 +159,14 @@ export default {
         reverse = true;
         r = r.substring(1);
       }
-      if ((!reverse && !user.roles.includes(r)) ||
```
awick, line 162: please write a test for this that fails with the old code and works with this new code. this hurts my head.

awick:
looks good, would be nice to have some editRoles tests :)

PR #2407, opened 2023-09-20
Transfer Resource Ownership

31453:


viewer/apiViews.js
```diff
@@ -229,8 +229,28 @@ class ViewAPIs {
...
+        let user = ArkimeUtil.commaOrNewlineStringToArray(view.user);
+        user = await User.validateUserIds(user);
+
+        if (user.invalidUsers?.length) {
+          return res.serverError(403, `Invalid user: ${user.invalidUsers[0]}`);
+        }
+
+        if (user.validUsers?.length) {
+          view.user = user.validUsers[0];
+        }
```
awick, line 248: the else should return an error, otherwise view.user ends up not getting set right?

31453: Technically there should never be an else case here. There are always `invalidUsers` and/or `validUsers` returned from the `validateUserIds` function. There is not a case where both are empty. But maybe I should still catch that case if we change the `validateUserIds` function?

common/auth.js
```diff
@@ -283,6 +283,61 @@ class Auth {
...
+    return async (req, res, next) => {
+      if (req.user.hasRole(Auth.appAdminRole)) { // an admin can do anything
+        return next();
+      } else {
+        try {
+          if (req.params.key !== undefined) {
+            req.body.key = req.params.key;
+          }
+
+          const id = req.params.id || req.body.key;
```
awick, line 308: ?? ?

awick: or maybe dont need anymore

PR #2420, opened 2023-10-03
initial saml support

awick:
## License


0xNIEI:
Does this mean we are not going to see SAML Support in 5.0.0?

awick:
correct, we need help testing and couldn't find anyone interested.

PR #2423, opened 2023-10-05
Add SPDX-License to all of capture

awick:
Switch capture/db to use SPDX-License-Identifier: Apache-2.0 instead of license text
## License


31453 (approved):
✨

PR #2473, opened 2023-10-25
Update LICENSE

awick:
Maybe this will make openssf happy
## License


31453 (approved):
I still think you need `Apache-2.0` but we'll seeeeeeeeee

31453:
I was WRONG 😭

awick:
we got an internet point!!!

PR #2482, opened 2023-10-27
uploadRoles and new authModes

awick:
* Can now set uploadRoles to control who can upload, defaults to arkimeUser, still need uploadCommand to be set also

* pass the logout url to the vueapp instead of authMode

* several new auth modes: headerOnly, header+digest, header+basic

## License

### DIFF
--- a/common/vueapp/Logout.vue
+++ b/common/vueapp/Logout.vue
@@ -5,7 +5,7 @@ SPDX-License-Identifier: Apache-2.0
 <template>
   <b-button
     :size="size"
-    v-if="authMode && authMode === 'form'"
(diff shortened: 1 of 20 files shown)


common/vueapp/Logout.vue
```diff
@@ -5,7 +5,7 @@ SPDX-License-Identifier: Apache-2.0
 <template>
   <b-button
     :size="size"
-    v-if="authMode && authMode === 'form'"
+    v-if="logoutUrl"
```
31453, line 8: 😲

31453 (approved):

PR #2565, opened 2023-12-21
decode punycode indicators

### LINKED ISSUE #2155: Cont3xt - Identify punycode encoded international domain names in search input
At indicator extraction time check for punycode encoded domain names that start with plain ascii `xn--`, and classify the iType as DOMAIN for API integration processing. Please display the utf-8 decoded result in parens next to the punycode domain in the indicator result tree.

example:
```
DOMAIN: xn--yho-ela6g.com  (yáhoó.com)
```

Integrations should be processed with the original punycode encoded domain name when gathering results.

31453:
fixes #2155 

## License

### DIFF
--- a/cont3xt/integration.js
+++ b/cont3xt/integration.js
@@ -16,6 +16,8 @@ const ipaddr = require('ipaddr.js');
 const Audit = require('./audit');
 const RE2 = require('re2');
 const normalizeCardField = require('./normalizeCardField');
+// Note: need the trailing slash to get the userland module instead of node core punycode module (deprecated)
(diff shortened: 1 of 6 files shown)


awick:
looks good, do you think it would be better to do on the client?

31453:
Yeah, I wasn't sure whether to do this on the client or server. I thought server since I could wrap it into the classify function. But I'd be happy to discuss the pros and cons. Yes, a test would be nice :) I can do that

awick (approved):

PR #2571, opened 2024-01-03
Reader schemes

awick:
Introduce the concept of reader schemes to process pcap files from multiple local and report locations. Support for new schemes can be built into capture/viewer or added with a plugin. The schemes run in a single thread, which will make file processing faster then previous version. Switching to "scheme" mode vs libpcap-file mode is done either by using -r with any scheme OR the --scheme option. It is all or nothing, once in scheme mode non scheme supported items will not be used.

Currently we support:
file
s3 - normal s3 url
s3http(s) - s3http://host/bucket/key using the s3 protocol, supports minio or s3 obj urls
http(s) - http://host/path


31453 (dismissed):
🚀  I think scheme makes sense. I don't love it, but I don't have any better suggestions.

31453 (dismissed):

31453 (approved):

Issue #2594, opened 2024-01-12
Allow intercepting TCP headers for the JA4Plus plugin


noeltimothy:
**Is your feature request related to a problem? Please describe.**
This feature request has been submitted to allow Arkime to send TCP packets that do not have any payload to TCP callbacks.
The current code does not send TCP packets without a payload to callbacks registered by plugins.
In case of the new Ja4Plus plugin, we want to be able to access TCP packets to record and create fingerprints based on the below criteria and fields.

- SYN packet timestamp
- SYN-ACK packet timestamps along with re-trasmitted SYN-ACKs
- ACK pacet timestamp
- TCP options
- TCP window size
- IP time to live
 
**Describe the solution you'd like**
Instead of adding the above fields to the Arkime session, we thought it would be cleaner to just send the entire ArkimePacket to plugins that have registered for TCP. The length of the packet will be zero in this case. It is upto the plugin to peek into the packet to decipher the TCP headers, IP headers or any other data that it requires.

**Describe alternatives you've considered**
Add timestamps into the TCP packet code and record different timestamps in the Arkime session.

**Additional context**
Details of JA4L and JA4T are @ https://github.com/FoxIO-LLC/ja4.

noeltimothy:
Submitted a PR for this at https://github.com/arkime/arkime/pull/2595

awick:
Used a named func instead #2605

PR #2595, opened 2024-01-12
fix for issue #2594, allow plugins to receive zero length TCP packets

noeltimothy:
This feature request has been submitted to allow Arkime to send TCP packets that do not have any payload to TCP callbacks.
The current code does not send TCP packets without a payload to callbacks registered by plugins.
In case of the new Ja4Plus plugin, we want to be able to access TCP packets to record and create fingerprints based on the below criteria and fields.

- SYN packet timestamp
- SYN-ACK packet timestamps along with re-trasmitted SYN-ACKs
- ACK pacet timestamp
- TCP options
- TCP window size
- IP time to live

Instead of adding the above fields to the Arkime session, we thought it would be cleaner to just send the entire ArkimePacket to plugins that have registered for TCP. The length of the packet will be zero in this case. It is upto the plugin to peek into the packet to decipher the TCP headers, IP headers or any other data that it requires.

2594

No changes were made to either viewer or parliament

Yes, successfully ran tests. Observed failures but not related to these changes.
[arkime-test-out.txt](https://github.com/arkime/arkime/files/13979153/arkime-test-out.txt)

## License

### DIFF
--- a/capture/parsers/tcp.c
+++ b/capture/parsers/tcp.c
@@ -182,6 +182,9 @@ int tcp_packet_process(ArkimeSession_t *const session, ArkimePacket_t *const pac
             session->tcpSeq[packet->direction] = seq + 1;
             session->synSet |= (1 << packet->direction);
         }
+	// Send only the header up with a length 0, useful for tools like JA4Plus
+        if (pluginsCbs & ARKIME_PLUGIN_TCP)
+            arkime_plugins_cb_tcp(session, packet, 0, packet->direction);
         return 1;
     }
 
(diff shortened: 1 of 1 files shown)


capture/parsers/tcp.c
```diff
@@ -213,6 +216,9 @@ int tcp_packet_process(ArkimeSession_t *const session, ArkimePacket_t *const pac
             session->ackTime = (packet->ts.tv_sec - session->firstPacket.tv_sec) * 1000000 +
                                (packet->ts.tv_usec - session->firstPacket.tv_usec) + 1;
         }
+	// Send only the header up with a length 0, useful for tools like JA4Plus
+        if (pluginsCbs & ARKIME_PLUGIN_TCP)
+            arkime_plugins_cb_tcp(session, packet, 0, packet->direction);
```
awick, line 221: since there is no return here, why is this one needed?

noeltimothy: Corrected this, I moved this to where the return is when the length <= 0.

awick:
used a named func instead

Issue #2664, opened 2024-02-21
[proposal] New generic field type with support for more complex structs


mcgillowen:
**Is your feature request related to a problem? Please describe.**
Creating fields that are more complex than a unitary/array value requires large changes to core Arkime source files like `arkime.h`, `db.c`, `fields.c`, etc, this makes it near impossible to write more complex parser outputs that need to store for example JSON objects.

An example of such a field is the `ARKIME_FIELD_TYPE_CERTSINFO` field type.

**Describe the solution you'd like**

There should be a generic field type that can work with the more complex/nested fields, the suggestion would be an `ARKIME_FIELD_TYPE_OBJECT` that supports generic objects as fields.

## ARKIME_FIELD_TYPE_OBJECT proposal

⚠️ This is still a WIP and some or most of the parts below may change while trying to implement such a field ⚠️ 

### Functions that would need to be implemented by a `ARKIME_FIELD_TYPE_OBJECT` field

* save: creates the object in JSON in a BSB, takes in a pointer to a BSB as argument as well as the object and session in question
* free: takes a pointer to the object
* hash: calculates a hash (uint32) for the object
* cmp: compares 2 objects

### Functions that are needed to manipulate ARKIME_FIELD_TYPE_OBJECT objects

* register: takes pointers for the functions save, alloc, free, hash and cmp along with the name of the object and returns the field position
* add: adds the specific struct containing the data/information, takes the field position, pointer to the session in question and a pointer to the object and the JSON length/size of it

#### ARKIME_FIELD_TYPE_OBJECT implementations

```
void certinfo_save(BSB *jbsb, ArkimeFieldObject_t *object, ArkimeSession_t *session);
void certinfo_free(ArkimeFieldObject_t *object);
uint32_t certinfo_hash(const void *key);
int certinfo_cmp(const void *keyv, const void *elementv);
```

#### ARKIME_FIELD_TYPE_OBJECT functions

```
int arkime_field_object_register(const char *name, ArkimeFieldObjectSaveFunc save, ArkimeFieldObjectFreeFunc free, ArkimeFieldObjectHashFunc hash, ArkimeFieldObjectCmpFunc cmp);
gboolean arkime_field_object_add(int pos, ArkimeSession_t *session, ArkimeFieldObject_t *object, int len);
```

#### ARKIME_FIELDS_TYPE_OBJECT struct

```
typedef struct arkime_field_object {
    struct arkime_field_object      *o_next, *o_prev;
[...]

PR #2671, opened 2024-02-22
Generic object field type

mcgillowen:

Their currently isn't an easy way to define a generic object field, something akin to the TLS certs field which isn't generic at all.

The proposed solution is discussed in issue #2664.

#2664 

No, since viewer or parliament aren't touched.

Yes

## License

### DIFF
--- a/capture/db.c
+++ b/capture/db.c
@@ -1337,7 +1337,7 @@ void arkime_db_save_session(ArkimeSession_t *session, int final)
             ArkimeFieldObject_t *object;
 
             HASH_FORALL_POP_HEAD2(o_, *ohash, object) {
-                saveCB(&jbsb, object);
+                saveCB(&jbsb, object, session);
                 freeCB(object);
                 BSB_EXPORT_u08(jbsb, ',');
             }
--- a/capture/db.c
+++ b/capture/db.c
@@ -1323,6 +1323,25 @@ void arkime_db_save_session(ArkimeSession_t *session, int final)
 
             BSB_EXPORT_rewind(jbsb, 1); // Remove last comma
             BSB_EXPORT_cstr(jbsb, "],");
+
+            break;
+        }
+        case ARKIME_FIELD_TYPE_OBJECT: {
+            ArkimeFieldObjectHashStd_t *ohash = session->fields[pos]->ohash;
+            ArkimeFieldObjectSaveFunc saveCB = config.fields[pos]->object_save;
+            ArkimeFieldObjectFreeFunc freeCB = config.fields[pos]->object_free;
+
+            BSB_EXPORT_sprintf(jbsb, "\"%sCnt\":%d,", config.fields[pos]->dbField, HASH_COUNT(o_, *ohash));
+            BSB_EXPORT_sprintf(jbsb, "\"%s\":[", config.fields[pos]->dbField);
+
+            ArkimeFieldObject_t *object;
+
+            HASH_FORALL_POP_HEAD2(o_, *ohash, object) {
+                saveCB(&jbsb, object);
+                freeCB(object);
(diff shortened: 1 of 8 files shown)


capture/field.c
```diff
@@ -1444,6 +1454,78 @@ void arkime_field_certsinfo_free (ArkimeCertsInfo_t *certs)
...
     ARKIME_TYPE_FREE(ArkimeCertsInfo_t, certs);
 }
+/******************************************************************************/
+int arkime_field_object_register(const char *name, ArkimeFieldObjectSaveFunc save, ArkimeFieldObjectFreeFunc free, ArkimeFieldObjectHashFunc hash, ArkimeFieldObjectCmpFunc cmp) {
+    int object_pos;
+    ArkimeFieldInfo_t *object_info;
+
+    object_pos = arkime_field_define(name, "notreal",
+                                     name, name, name,
+                                     "Object Info", /*TODO: Improve the help string to contain object name*/
```
awick, line 1464: or maybe just pass in?

capture/db.c
```diff
@@ -1323,6 +1323,28 @@ void arkime_db_save_session(ArkimeSession_t *session, int final)
...
+            ArkimeFieldObjectFreeFunc freeCB = config.fields[pos]->object_free;
+
+            BSB_EXPORT_sprintf(jbsb, "\"%sCnt\":%d,", config.fields[pos]->dbField, HASH_COUNT(o_, *ohash));
+            BSB_EXPORT_sprintf(jbsb, "\"%s\":[", config.fields[pos]->dbField);
+
+            ArkimeFieldObject_t *object;
+
+            HASH_FORALL_POP_HEAD2(o_, *ohash, object) {
+                saveCB(&jbsb, object, session);
+                freeCB(object);
```
awick, line 1341: I know certs doesn't do it, but lets add if(freeField) here and the ohash free below

awick (commented):
Looking good, 2 minor things

capture/db.c
mcgillowen: Sure, sounds like a good idea, hadn't really noticed it in the other types.

mcgillowen:
I'm working on migrating TLS certs to this generic field type, I already have something close to working, the ordering seems to be different from the tests, still checking that out.

And I also had to remove the `void arkime_field_certsinfo_update_extra (ArkimeCertsInfo_t *certs, char *key, char *value);` function since in this current implementation of generic fields there is nothin similar, but as far as I could see that function was never used.

awick:
It is used in the ja4plus plugin https://github.com/arkime/ja4

I check it out and then do a soft link of ja4plus.c into the plugins directory.

You probably will needs to make a certs_internal.h file that the plugin can use that has the new ArkimeCertsInfo_t in it and move arkime_field_certsinfo_update_extra into the parsers dir.

awick:
or i guess for now you could leave the types in arkime.h, and we can do the move after its working, up to you

capture/field.c
```diff
@@ -1382,6 +1384,14 @@ void arkime_field_free(ArkimeSession_t *session)
             }
             ARKIME_TYPE_FREE(ArkimeCertsInfoHashStd_t, cihash);
             break;
+        case ARKIME_FIELD_TYPE_OBJECT:
+            ArkimeFieldObjectFreeFunc freeCB = config.fields[pos]->object_free;
```
awick, line 1388: this is failing on old compilers. you need to have { } after the case and before the break, or move the freeCB variable declaration above the switch. I do both. :)

mcgillowen: OK, did not know that, I'll fix it when I get in front of my computer
Thanks

awick (approved):

PR #2683, opened 2024-02-28
lock integrations

### LINKED ISSUE #2128: Lock Integration Settings
Admin should be able to lock a user's integration settings (the user can't change their integration settings).

* Per integration allow a locked=true in config file. (Integration.register)
* If set, the user config for that integration is totally ignored, even if previously set (Integration.getUserConfig?)
* in the UI it would be nice to disable editing the integration config, other than turning it off per user

31453:
previous integration setting changes are still there, just overridden by global setting
can still update integration settings via the API but won't do anything as the global setting will be used
show locked indicator and explanation in UI
fixes #2128

## License


awick (approved):

awick (dismissed):
again again

awick (approved):
again again again

Issue #2560, opened 2023-12-20
Split DNS requests from CNAME responses


problematiq:
Please separate queries and responses for DNS into their own fields.
This is useful when looking for beacons via DNS without having to sift through the CNAME responses.

```
Is there a way to separate out a dns query from a DNS response?
so, I ask for blank[.]com and it returns cdn.blank[.]com CNAME, 7.7.7.7 A, bob[.]blanl[.]com CNAME
The host.dns field now has the query and the cnames.
```

31453:
Looking at redoing DNS Parser.

awick:
Let me know if #2694 doesn't address

PR #2694, opened 2024-03-04
New DNS parser

mcgillowen:

The current DNS parser mixes the hostnames and IPs, this new parser should fix that and uses the OCSF DNS Activity as output format.

None

Yes

The tests are currently not passing, the new output seems to be non-deterministic, still trying to figure out why.

## License


awick:
Looking good, few comments so far

* is "ocsfdns" the right root in the db schema? (do you have some links to share about the schema, maybe add to top of file?)
* filenames probably should be ocsfdns to match
* please add the /*...*/ separator between functions
* curious why rdata is merging fields that have been parsed out instead of just having separate fields

mcgillowen:
> is "ocsfdns" the right root in the db schema? (do you have some links to share about the schema, maybe add to top of file?)

From what I could find OCSF stores different activities independent of each other, multiple documents for the same "session" , but I chose "ocsfdns" since it's a ocsf representation of the DNS activity, I'm open to other suggestions

I'll add a link and comment the different fields, the schema defines a lot of optional, recommended fields and some required ones, I chose what I deemed most sensible

> I wonder if the expression should be "ocsf.dns...."

That is actually a good idea, if more ocsf schemas get added, I'll see how it works

> filenames probably should be ocsfdns to match

Makes sense, I'll change that

> please add the /.../ separator between functions

Sure, no problem

> curious why rdata is merging fields that have been parsed out instead of just having separate fields

It's mainly because initially I had a single field "rdata" which contained the "canonical" string representation for each RR type and the save function would simply copy that string into the JSON but then decided I preferred having the "canonical" string representation defined in the save function and the parsing only extract the data without having to then create the correct string representation.
I'm not attached to either way, if you prefer it the other way around, I can change it. 

QQ what do you use to format/lint the C code?

awick:
> From what I could find OCSF stores different activities independent of each other, multiple documents for the same "session" , but I chose "ocsfdns" since it's a ocsf representation of the DNS activity, I'm open to other suggestions

So they all live at the root level ?

> I'll add a link and comment the different fields, the schema defines a lot of optional, recommended fields and some required ones, I chose what I deemed most sensible

ok, I found some stuff, but was hoping there was some combined easy to read one that you were using :)

> > curious why rdata is merging fields that have been parsed out instead of just having separate fields
> 
> It's mainly because initially I had a single field "rdata" which contained the "canonical" string representation for each RR type and the save function would simply copy that string into the JSON but then decided I preferred having the "canonical" string representation defined in the save function and the parsing only extract the data without having to then create the correct string representation. I'm not attached to either way, if you prefer it the other way around, I can change it.

It seems like if we are parsing stuff we should just make them new fields, unless ocsf doesn't do that. For example with mx priority the first thing is someone will say they want to be able to do integer searches. lol

> QQ what do you use to format/lint the C code?
https://astyle.sourceforge.net/ -  I couldn't get clang-tidy to work like I wanted :) I'll add some comment somewhere. You can just type "make format" in capture dir

mcgillowen:
> So they all live at the root level ?

Not really, each protocol activity/event would be a separate document in the index, so if there were 3 DNS queries there would be 3 separate documents in the ES index.

As far as I could find events don't get grouped together, they are all independent of each other, so there is no idea of a root level

I based everything off of https://schema.ocsf.io/1.1.0/classes/dns_activity?extensions=

> It seems like if we are parsing stuff we should just make them new fields, unless ocsf doesn't do that. For example with mx priority the first thing is someone will say they want to be able to do integer searches. lol

OCSF only defines a single "rdata" field in it's DNS answer object type, https://schema.ocsf.io/1.1.0/objects/dns_answer?extensions=, and it's format depends on the type and class of the RR

awick:
Let me ask another way. So lets say i had dns and http events, I get that there would be a document for each, would it be
1) a dns index and a http index
2) a dns wrapper and a http wrapper inside 1 index
3) no wrappers and just all the fields are at the root level?


> I based everything off of https://schema.ocsf.io/1.1.0/classes/dns_activity?extensions=

thx, that makes it look like its 3 above

> OCSF only defines a single "rdata" field in it's DNS answer object type, https://schema.ocsf.io/1.1.0/objects/dns_answer?extensions=, and it's format depends on the type and class of the RR

so same problems as ECS :(

mcgillowen:
As far as I could find it is supposed to be option 3, and yes it is similar to ecs in that regard.

Maybe OCSF isn't the best fit for Arkime, we could come up with our own for DNS, call it "dnsv2" or something similar and even have a configuration option to choose between different output formats, the challenging part would then be having multiple pugjs files for each output format

mcgillowen:
The main issue with the current implementation is that it's unusable if one wants to match returned the IP/NS/CNAME/others with the queried hostnames, etc.. Everything is lumped together over the potentially multiple DNS query/responses in a session, without being able to figure out what goes together.

awick:
Ok I think I have a crazy solution that will actually work and not require any mapping changes.
Right now we do something like:
```
dns: {
  host: ["query1", "query2"],
  ip: ["ip1", "ip2"]
}
```

if we instead output it as
```
dns: [
  {
    host: "query1",
    ip: "ip1"
  },
  {
    host: "query2",
    ip: "ip2"
  }
]
```

then what is actually generated inside ES for searches will be exactly the same, the only difference will be the source document. Then the dns jade file would just need to check if its an array (new) or object (old) data when displaying a session. I think there is also some magic in viewer that would have to change too.

Now counts would be different, and we would need to add a new dns.count, but this might be the best way forward.

mcgillowen:
There you are assuming you only get IPs as answer, which is not the case, you can also have CNAME/NS/MX/others. It also doesn't keep the CNAME chaining intact either which is something we are interested in seeing, also check the issue #2560 .

There is also the type, classe, TTL that should be kept with the answer, and the name in the RR which can differ from the queried hostname.

awick:
sorry I wasn't clear, in my example the point wasn't the fields, the points are
* no new mapping or fields needed
* switching from object with arrays to arrays with objects
* you'll be able to use almost all the same expressions we have now that will go against old data and new data. (so no dns.host == X OR ocsf.dns.host == X for historical and recent changes)

We can definitely add new fields to this, and it will probably make more sense with this format

mcgillowen:
Ok, that could work.

So the only thing that would need to be changed in this PR would be the save function, removing the OCSF specific parts, and renaming the expressions, fields? Or am I missing something?

And since there can be multiple answer records per query, these should also be matched up preferably.

Here's an example of the viewer with the implementation in this PR.

image

awick:
so need to make sure you are using the same field names for things that are the same. we will also need to clean up that jade, probably using tables or something like certs.

mcgillowen:
OK, I'll take the current output of the dns-wireshark.pcap file and adapt it to something like your proposal.

I agree that the jade needs to be cleaned up, it's a first draft to display the changes and was then planning on getting Elyse's input to improve it once the format was more or less defined.

31453:
Looks good! My only thought is that this should all be in one DNS section, with subsections, [like the HTTP section](https://github.com/arkime/arkime/blob/main/capture/parsers/http.detail.jade).

mcgillowen:
I've taken the output of the dns-wiresharkrepo.pcap and created a file that contains both the current implementation and a proposed new output.

Here's an excerpt:

```json
{
  "answers": [
    {
      "rrc": "IN",
      "flags": ["RECURSION DESIRED","RECURSION AVAILABLE"],
      "packet_uid": 63343,
      "mailserverHost": "(40)smtp4.google.com",
      "rr_name": "google.com",
      "rr_type": "Answer",
      "ttl": 552,
      "rrt": "MX"
    },
    {
      "rrc": "IN",
      "flags": ["RECURSION DESIRED","RECURSION AVAILABLE"],
      "packet_uid": 63343,
      "mailserverHost": "(10)smtp5.google.com",
      "rr_name": "google.com",
      "rr_type": "Answer",
      "ttl": 552,
      "rrt": "MX"
    },
    {
      "rrc": "IN",
      "flags": ["RECURSION DESIRED","RECURSION AVAILABLE"],
      "packet_uid": 63343,
      "mailserverHost": "(10)smtp6.google.com",
      "rr_name": "google.com",
      "rr_type": "Answer",
      "ttl": 552,
      "rrt": "MX"
    },
    {
      "rrc": "IN",
      "flags": ["RECURSION DESIRED","RECURSION AVAILABLE"],
      "packet_uid": 63343,
      "mailserverHost": "(10)smtp1.google.com",
      "rr_name": "google.com",
      "rr_type": "Answer",
      "ttl": 552,
      "rrt": "MX"
    },
    {
      "rrc": "IN",
      "flags": ["RECURSION DESIRED","RECURSION AVAILABLE"],
      "packet_uid": 63343,
      "mailserverHost": "(10)smtp2.google.com",
      "rr_name": "google.com",
      "rr_type": "Answer",
      "ttl": 552,
      "rrt": "MX"
    },
    {
      "rrc": "IN",
      "flags": ["RECURSION DESIRED","RECURSION AVAILABLE"],
      "packet_uid": 63343,
      "mailserverHost": "(40)smtp3.google.com",
      "rr_name": "google.com",
      "rr_type": "Answer",
      "ttl": 552,
      "rrt": "MX"
    },
    {
      "rrc": "IN",
      "flags": ["RECURSION DESIRED","RECURSION AVAILABLE"],
      "packet_uid": 63343,
      "mailserverIp": "216.239.37.26",
      "rr_name": "smtp4.google.com",
      "rr_type": "Additional",
      "ttl": 600,
      "rrt": "A"
    },
    {
      "rrc": "IN",
      "flags": ["RECURSION DESIRED","RECURSION AVAILABLE"],
      "packet_uid": 63343,
      "mailserverIp": "64.233.167.25",
      "rr_name": "smtp5.google.com",
      "rr_type": "Additional",
      "ttl": 600,
      "rrt": "A"
    },
    {
      "rrc": "IN",
      "flags": ["RECURSION DESIRED","RECURSION AVAILABLE"],
      "packet_uid": 63343,
      "mailserverIp": "66.102.9.25",
      "rr_name": "smtp6.google.com",
      "rr_type": "Additional",
      "ttl": 600,
      "rrt": "A"
    },
    {
      "rrc": "IN",
      "flags": ["RECURSION DESIRED","RECURSION AVAILABLE"],
      "packet_uid": 63343,
      "mailserverIp": "216.239.57.25",
      "rr_name": "smtp1.google.com",
      "rr_type": "Additional",
      "ttl": 600,
      "rrt": "A"
    },
    {
      "rrc": "IN",
      "flags": ["RECURSION DESIRED","RECURSION AVAILABLE"],
      "packet_uid": 63343,
      "mailserverIp": "216.239.37.25",
      "rr_name": "smtp2.google.com",
      "rr_type": "Additional",
      "ttl": 600,
      "rrt": "A"
    },
    {
      "rrc": "IN",
      "flags": ["RECURSION DESIRED","RECURSION AVAILABLE"],
      "packet_uid": 63343,
      "mailserverIp": "216.239.57.26",
      "rr_name": "smtp3.google.com",
      "rr_type": "Additional",
      "ttl": 600,
      "rrt": "A"
    }
  ],
  "answersCnt": 12,
  "query": {
    "packet_uid": 63343
  },
  "qc": "IN",
  "host": "google.com",
  "opcode": "QUERY",
  "qt": "MX",
  "status": "NOERROR"
}
```

In the answers I changed `qc` and `qt` to `rrc` and `rrt` respectively, but I guess these could be aliased to be queried by `qc` and `qt`. There's a couple of fields in the implementation that I'm not sure how to change them, especially `rr_name`.

What do you think?

Full file of the suggestion
[dnsv2.json](https://github.com/arkime/arkime/files/14499287/dnsv2.json)

awick:
hmm will have to think about more, the issue is if it isn't exactly the same path then viewer will have to do conversion if we wanted to REPLACE the dns parser which was my hope.

awick:
I think you are right that we should have the current format PLUS the answers and a few other NEW fields.

In general I think we should use the DNS terminology for things (https://en.wikipedia.org/wiki/Domain_Name_System) or the ECS terminology (https://www.elastic.co/guide/en/ecs/current/ecs-dns.html), so I don't love the rrc stuff from above.

I would
1) add dns.answers[].(class|data|name|ttl|type|host|ip)
2) if we need the mailhost preference we should add that and not make it part of host
3) for flags i like the 2 letter abbreviation that both DNS/ECS seem to use
4) still emit the current fields

so something like this EXAMPLE, the fields aren't quite right. I also wouldn't add the id's unless you think those are useful.
```{
  "answers": [
    {
      "host": "smtp4.google.com",
      "preference": 40
      "name": "google.com",
      "type": "Answer",
      "ttl": 552,
      "type": "MX"
    },
    ]
    header_flags: ["RD","RA"],
    host: ["all", "the", "hosts"],
    ip: [all, the, ips],
    ...
```

mcgillowen:
That looks like a very good format, there is maybe one bit missing, a field that contains the queried hostname that isn't grouped with all the other domain names.

Replacing dns.c shouldn't be a problem, and I don't think the id field is necessary.

I'll make the requisite changes.

awick:
I dont know why this is hurting my brain so much, but ya this isn't complete. We either need to do
```
dns: {
  queries: [
    { // query 1
      host: "smtp.google.com",
      answers: [...] 1
      ....
    },
    {
    // query 2
      host: "example.com"
      answers: [...] 2
   }],
    host: ["all", "the", "hosts"],
    ip: [all, the, ips],
    ...
}
```

OR
```
dns: [{
   //query 1
  host: "smtp.google.com",
  answers: [...] 1
  ip: [all, the, ips, query, 1],
},
{
   //query 2
  host: "example.com",
  answers: [...] 2
  ip: [all, the, ips, query, 2],
}]
```

mcgillowen:
I was leaning towards the 2nd option.
I'd add  the mailserverHost, nameserverHost, mailserverIp, nameserverIp at the root of each object as well and probably a field `additionalHosts` a list of hostnames that come from the resource records which together with the `host` field would recreate the same information as the current implementation.

```json
dns: [{
   //query 1
  host: "smtp.google.com",
  additionalHosts: [all, the, domain, names, in, RRs],
  nameserverHost: [all, the, NS, domain, names, in, RRs],
  mailserverHost: [all, the, MX, domain, names, in, RRs],
  answers: [...],
  ip: [all, the, ips, query, 1],
  nameserverIp: [all, IPs, who's, RR name, matches, a, NS],
  mailserverIp: [all, IPs, who's, RR name, matches, a, MX],
}, ...]
```

mcgillowen:
I have something close to a working implementation, but I think that the *Cnt fields might be a problem and not sure how to solve it yet.

awick:
i'm fine with breaking most of the Cnt fields :)

mcgillowen:
Ok, I've pushed what I have so far, still have to fix some issues with sorting and double-check that the format is still compatible.

I've so far only run the unit tests of the pcaps (tests/tests.pl) still have to run the remaining tests and correct any issues.

awick:
I like where this is going! 
* Can you remove the ocsfdns items
* Can you remove dns.cold

mcgillowen:
Sure, I still have some cleaning up to do, especially in the parser

mcgillowen:
This is now considered feature complete, but the tagger and wise plugins don't work at the moment.

The viewer tests still need to be adapted to run without error.

mcgillowen:
@awick thanks for those improvements, I was playing around with also including the domain names in the additionalHosts hash into the dns_getcb_host returned hash

mcgillowen:
@31453 could you take a look at the jade template, I did the same as the HTTP card but since there's an extra level the answers section has the same width as the query which groups them, is there another class I should use or could you make them slightly less width or a different color to differentiate them? Thanks

31453 (approved):

PR #2718, opened 2024-03-25
pixel perf and green theme improvements

31453:
just trust me :) 

## License


awick (approved):
where is the zindex change?

PR #2755, opened 2024-04-16
Add routing allocation exclusion in db.pl

killerwhile:

We're using an ES cluster where we dedicate/isolate nodes for storing sessions. Our rational doing that is that metadata indices (especially files) should not be lost even if nodes storing the sessions are overloaded: loosing 1 file means loosing all the sessions stored in the file).
ES provides a mechanism called index routing allocation, and in our case, we tag _metadata_ nodes as such and exclude sessions indices from these metadata nodes.

ES template contains something like: `index.routing.allocation.exclude.arkimeNodeType = metadata`, `arkimeNodeType = metadata` being the tag of the _metadata_ nodes.

This PR provides a way to initialise the sessions template with a specific `index.routing.allocation.exclude` value using the parameter `--sessionsroutingallocationexcludekey key1=value1`

n/a

## License

### DIFF
--- a/db/db.pl
+++ b/db/db.pl
@@ -108,6 +108,7 @@ my $SHARDSPERNODE = 1;
 my $ESTIMEOUT=60;
 my $UPGRADEALLSESSIONS = 1;
 my $DOHOTWARM = 0;
+my $SESSIONSROUTINGALLOCATIONEXCLUDEKEY = 0;
 my $DOILM = 0;
 my $DOISM = 0;
 my $WARMAFTER = -1;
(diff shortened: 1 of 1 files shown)


awick:
Thanks, but I'm wondering if you could do without a change to db.pl. Just make 2 new index templates (with different names than we use :)) One that matches arkime_sessions* and one the meta ones, and set the exclude yourself in that new template. I guess the issue is this won't change existing ones, but not sure the current db.pl handles that either.

If that won't work, please add help to showHelp, and I think the leading "index." isn't needed since

killerwhile:
Thanks for the comment. I'll check the 2nd template and come back

awick:
Let me know if you can't get the multiple templates working

PR #2773, opened 2024-05-03
auto save users

31453:
remove save and cancel buttons
remove all the silly logic that came alone with waiting to save

## License


awick:
Would it be possible to not have it auto close the extra area? So if I change the force expression when it saves it closes the area right now :(

awick (approved):

PR #2791, opened 2024-05-16
Fix incorrect UDP data length when UDP packet shorter than minimum ethernet packet length

mcgillowen:

The UDP parser parses the databytes length incorrectly when the UDP packet is very short, less than what is needed to reach the minimum ethernet packet size which is 46 bytes. When a UDP packet is shorter than 14 bytes, it will have padding added after the UDP data section.

The solution is to read the length field of the UDP header rather than subtract from the total payload length.

Yes

## License


capture/parsers/udp.c
```diff
@@ -61,15 +61,15 @@ int udp_pre_process(ArkimeSession_t *session, ArkimePacket_t *const packet, int
...
+    session->databytes[packet->direction] += ntohs(udphdr->uh_ulen) - 8;
 
     return 0;
 }
 /******************************************************************************/
 int udp_process(ArkimeSession_t *session, ArkimePacket_t *const packet)
 {
     const uint8_t *data = packet->pkt + packet->payloadOffset + 8;
-    int            len = packet->payloadLen - 8;
+    uint16_t        len = (packet->pkt[packet->payloadOffset + 4] << 8 | packet->pkt[packet->payloadOffset + 5]) - 8;
```
awick, line 72: So this will need some kind of MIN call otherwise could cause a buffer overflow read if len is greater then packet size.

capture/parsers/udp.c
```diff
@@ -61,15 +61,15 @@ int udp_pre_process(ArkimeSession_t *session, ArkimePacket_t *const packet, int
     packet->direction = (dir &&
                          session->port1 == ntohs(udphdr->uh_sport) &&
                          session->port2 == ntohs(udphdr->uh_dport)) ? 0 : 1;
-    session->databytes[packet->direction] += (packet->pktlen - packet->payloadOffset - 8);
+    session->databytes[packet->direction] += ntohs(udphdr->uh_ulen) - 8;
```
awick, line 64: I'm not sure why this one should change, since payloadOffset - 8 should be the actual size of the data on the network, or maybe I don't understand.

awick (changes requested):
thanks for adding a pcap!!! Added some question/comments.

capture/parsers/udp.c
mcgillowen: It is correct as long as there isn't any padding after the UDP part, unless I've misunderstood what `session->databytes` represents, I always thought it was the number of bytes that aren't part of the protocol headers.

With the provided PCAP, databytes was 18 instead of 1 before this change.

capture/parsers/udp.c
mcgillowen: Yes, good catch, I'll add it

capture/parsers/udp.c
awick: ah ok, so you are saying when there is padding we shouldn't count that as databytes but do could that in totalBytes. That makes sense. Should this also have the same MIN as adding?

mcgillowen: Yes, exactly.

I'm not sure if it needs the MIN, but could be a good idea just to be safe.

awick (approved):

PR #2794, opened 2024-05-20
allowTemplateLiterals so we can remove backslash single quotes

awick:
A grep was failing for an error with a single quote and took me a while to figure out why. Let this never happen again. Still broken in vue code. ;)

## License


31453 (approved):
Let this never happen again! 💀

PR #2806, opened 2024-05-24
Update easybutton-build.sh

vpiserchia:
Added missing sudo to the apt-get command

apt-get requires root.

## License


awick:
thanks!

PR #2814, opened 2024-05-28
Systemd service identifiers

vpiserchia:
**The Problem**
Systemd do not report the correct service identified.
By using the keyword "SyslogIdentifier" this is possible 

## License

### DIFF
--- a/release/arkimecapture.systemd.service
+++ b/release/arkimecapture.systemd.service
@@ -13,6 +13,7 @@ ExecStart=/bin/sh -c 'BUILD_ARKIME_INSTALL_DIR/bin/capture -c BUILD_ARKIME_INSTA
 WorkingDirectory=BUILD_ARKIME_INSTALL_DIR
 LimitCORE=infinity
 LimitMEMLOCK=infinity
+SyslogIdentifier=arkimecapture
 
 [Install]
 WantedBy=multi-user.target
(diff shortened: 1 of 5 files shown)


awick:
How do I test this?

vpiserchia:
when you install/enable/start the service files then you'll see something like in the journal log:

`
May 29 00:33:42 arkime5 arkimeviewer[420205]: This node will process Periodic Queries (CRON) & Hunts, delayed by 85 seconds
`

instead of

`
May 28 13:36:23 arkime5 node[208362]: WARNING - No cronQueries=true found in /data/arkime/viewer/config.ini, one and only one node MUST have cronQueries=true set for cron/hunts to work
`

awick:
How are you getting the logs into your journal log? I don't see any change in wise.log for example.

vpiserchia:
(with shame) Thank you Andy, I missed to commit the latest code 

In order to use systemd and journald I removed the bash script invocation and fully use the systemd directive `StandardOutput` and `StandardError`.

By commenting out these two lines one can decide to have logs in journald or the normal log files.

awick:
Unfortunately these keywords are not available on RHEL 7, so this will need to wait until we drop support, another month or so. (This is why we didn't use them.)

vpiserchia:
understood, sorry for wasting your time; I didn't realize it

vpiserchia:
Closing since RHEL 7 does not support  used directives

awick:
We should do this in the Arkime 5.3 or 5.4 timeframe

PR #2848, opened 2024-06-19
move some db to async/await

awick:
Move some db.js to async/await. Have some of the caches use promises since that will naturally block multiple waiters.
## License


31453 (approved):
🚀

PR #2871, opened 2024-07-09
Keep at least 10 PCAP files

vpiserchia:
In the file "expire" loop, the number of remaining files is not considered.

## License


awick:
I think we could just change the from: to 10 and remove both of the checks.

awick:
#2887 seems to work

PR #2887, opened 2024-07-19
expire improvements

awick:
* Handle 'G' at the end of freeSpaceG
* improve expire logging
* Actually enforce that we keep 10 files around

## License

### DIFF
--- a/viewer/viewer.js
+++ b/viewer/viewer.js
@@ -971,7 +971,7 @@ function expireDevice (nodes, dirs, minFreeSpaceG, nextCb) {
     }
 
     if (data.hits.total <= 10) {
-      console.log(`EXPIRE WARNING - not deleting any files since ${data.hits.total} <= 10 minimum files per node. Your disk may fill!!! See https://arkime.com/faq#pcap-deletion`);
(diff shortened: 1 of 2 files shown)


viewer/viewer.js
```diff
@@ -927,7 +927,7 @@ function expireDevice (nodes, dirs, minFreeSpaceG, nextCb) {
   }
   const query = {
     _source: ['num', 'name', 'first', 'size', 'node', 'indexFilename'],
-    from: '0',
```
vpiserchia, line 930: +1

viewer/viewer.js
```diff
@@ -971,9 +971,7 @@ function expireDevice (nodes, dirs, minFreeSpaceG, nextCb) {
     }
 
     if (data.hits.total <= 10) {
```
vpiserchia, line 973: I think that this check is wrong now, it should be:
```
if (data.hits.total == 0) {
```

vpiserchia: I checked again and everything works fine, sorry for the noise.

awick: np - hits.total is the number of matching documents ignoring pagination (from)

PR #2892, opened 2024-07-23
try backing off health reqs

### LINKED ISSUE #2801: header length bug : +opportunity to add custom error handler pages :D
I suspect due to the health check, when a client loses connectivity to it's Arkime/cont3xt server for a period of time, the cookie header value continues to grow until the server can no longer process the header. 

That condition results in (the not very pretty): 
```
<!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN">
<html><head>
<title>400 Bad Request</title>
</head><body>
<h1>Bad Request</h1>
<p>Your browser sent a request that this server could not understand.<br />
Size of a request header field exceeds server limit.</p>
</body></html>
```

I guess this is also specific to the web server chosen to front the app. But in the case of Apache, can the project recommend modifications to server error handler pages? Even errors should be Arkime/cont3xt themed!

I wonder if you would be able to modify the cases where header field length exceptions occur, would it even be possible to "suggest" clearing the now broken cookie to the browser specific 'dump' option?

Not sure if this would be a client side violation, but it would be nice to make a clickable reference to help a user clear the site specific cache:

* For Chrome:
```
chrome://settings/content/siteDetails?site=https%3A%2F%2F${local_server_instance_name}%2F&search=cookie
```

How about a custom 400 handler for:
```
/error/HTTP_BAD_REQUEST.html.var
```

Ultimately, there is a bug that only surfaces during periods of client inability to connect to the server, and until site cache is cleared, the user is broken. I've found users who gave up on the app because they didn't know how to resolve this problem manually.

31453:
fixes #2801 

## License

### DIFF
--- a/cont3xt/vueapp/src/utils/Navbar.vue
+++ b/cont3xt/vueapp/src/utils/Navbar.vue
@@ -161,6 +161,8 @@ import Logout from '@/../../../common/vueapp/Logout';
 import Version from '@/../../../common/vueapp/Version';
 
 let interval;
+const minTimeToWait = 10000;
+let timeToWait = minTimeToWait;
 
 export default {
   name: 'Cont3xtNavbar',
@@ -200,13 +202,8 @@ export default {
     } else {
       this.theme = this.getTheme; // initialize theme setting side-effects
     }
-    interval = setInterval(() => {
-      axios.get('api/health').then((response) => {
-        this.healthError = '';
-      }).catch((error) => {
-        this.healthError = error.text || error;
-      });
-    }, 10000);
+
+    this.getHealth();
   },
   methods: {
     /* page functions ------------------------------------------------------ */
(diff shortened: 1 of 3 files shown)


cont3xt/vueapp/src/utils/Navbar.vue
```diff
@@ -217,6 +214,23 @@ export default {
...
       window.location.reload();
+    },
+    /* helper functions ---------------------------------------------------- */
+    getHealth () {
+      interval = setInterval(() => {
+        axios.get('api/health').then((response) => {
+          this.healthError = '';
+          if (timeToWait !== minTimeToWait) {
+            timeToWait = minTimeToWait;
+            this.getHealth();
```
awick, line 225: I think you need a clear above this? Maybe just have a clear at the top of getHealth if interval is set?

awick (approved):

PR #2893, opened 2024-07-24
Force removal of problematic pcap when configured

vpiserchia:
When PCAP file deletion is chosen together with errors ignore, effectively remove files.

**the problem and solution**
As of now Arkime does not respect the configuration parameter "deletePCAP" and "ignoreErrors" when it comes to problematic PCAP.

## License


awick (changes requested):
Looks good other then the first 2 NULL changes don't seem to be needed, can you change back.

awick (approved):

PR #2909, opened 2024-08-13
Init cont3xt Db before Auth

awick:
Move auth init until after Db init since it can take a while and mess up timing.
## License


31453 (approved):
IT WORKS ✨

PR #3022, opened 2024-11-19
Cont3xt d3

31453:
just testing and diffing. going to delete

TODO
- [ ] do this on the dev6 branch
- [ ] put d3 top level instead of in viewer AND cont3xt
- [ ] put cont3xt assets in top level assets?
- [ ] more stuff I'm sure


31453:
postponing this for more important tasks

PR #3118, opened 2025-02-07
update twilio integration

31453:
use v2 api
include new fields
update twilio card
update tidbits
update default card

## License


31453:
This is not complete because I can't seem to update the default card. This needs some discussion and debugging @awick

PR #3134, opened 2025-02-17
switch to node 22

awick:
Use node 22 for builds
Require nodejs >= 20.9 and < 23
Later versions 5.x of Arkime did require 20.0, but we didn't update everything, oops.
## License


31453 (approved):
it seems to work 🎉

PR #3280, opened 2025-07-22
implemented dnssec records parser

dontepanlin:
### implemented dnssec records parser

* This commit only changes dns parser and not viewer

## License

### DIFF
--- a/capture/parsers/dns.c
+++ b/capture/parsers/dns.c
@@ -33,7 +33,12 @@ typedef enum dns_type {
     DNS_RR_TXT        =  16,
     DNS_RR_AAAA       =  28,
     DNS_RR_HTTPS      =  65,
-    DNS_RR_CAA        = 257
+    DNS_RR_CAA        = 257,
+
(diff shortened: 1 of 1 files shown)


awick:
Thanks for the great start! It needs a few changes before we can merge

- [ ] Need a test pcap(s) for the new fields
- [ ] The signature and bitmap need to use BSB_IMPORT_byte instead of BSB_IMPORT_ptr
- [ ] Looks like the signature length is wrong, and maybe should just be the length of the rest of the section, at least that is wireshark seems to do on the one file we currently have in tests.
- [ ] Do you really want to be able to search on all these fields? I think some could probably be eliminated, for example signature and bitmap
- [ ] Need to add arkime_field_define and maybe arkime_field_by_exp_add_internal for the new fields
- [ ] currently extra dns fields aren't saved in the db under a new container (see mx, txt, etc)
- [ ] Need to update dns.detail.jade to display new fields

awick:
reimplemented in #3620

PR #3314, opened 2025-08-20
Fixes for SQS/S3 scheme with MinIO

mcgillowen:

`sqshttp` scheme would not connect over HTTP, only HTTPS, the scheme identification wasn't working.

Scheme S3 requests were getting denied because of incorrect HTTP headers.

No changes to viewer

Yes

## License

### DIFF
--- a/capture/reader-scheme-s3.c
+++ b/capture/reader-scheme-s3.c
@@ -218,7 +218,6 @@ LOCAL void scheme_s3_request(void *server, const ArkimeCredentials_t *creds, con
     } else {
         headers[1] = "Content-Type:";
     }
-    headers[1] = NULL;
     headers[2] = NULL;
     headers[3] = NULL;
 


capture/reader-scheme-s3.c
```diff
@@ -213,7 +213,12 @@ LOCAL void scheme_s3_request(void *server, const ArkimeCredentials_t *creds, con
 
     char *headers[8];
     headers[0] = "Expect:";
-    headers[1] = "Content-Type:";
+    if (pathStyle) {
+        headers[1] = NULL;
+    } else {
+        headers[1] = "Content-Type:";
+    }
+    headers[1] = NULL;
```
awick: This looks like it will clear headers[1] even after setting it.

Can you help me understand why stopping content type from being set is bad? (That is what a empty header with curl does in this case). writer-s3.c has this same code, although it doesn't use the curl aws sig stuff. Did you test this change with real s3?

awick (changes requested):

capture/reader-scheme-s3.c
mcgillowen: Yes, that's a mistake, thanks for spotting it.

The reason is that the aws sig stuff expects the content-type apparently, I was getting `AccessDenied`

```
> GET /test/test.pcap HTTP/1.1
Host: minio.test
Authorization: AWS4-HMAC-SHA256 Credential=test/20250820/us-east-1/s3/aws4_request, SignedHeaders=content-type;host;x-amz-date, Signature=<sig>
X-Amz-Date: 20250820T171517Z
User-Agent: arkime
Accept: */*
Accept-Encoding: deflate, gzip, br, zstd

* Mark bundle as not supporting multiuse
< HTTP/1.1 400 Bad Request
< Accept-Ranges: bytes
< Content-Length: 428
< Content-Type: application/xml
< Server: MinIO
< Strict-Transport-Security: max-age=31536000; includeSubDomains
< Vary: Origin
Aug 20 17:15:17 reader-scheme-s3.c:100 scheme_s3_parse_region(): ERROR - <?xml version="1.0" encoding="UTF-8"?>
<Error><Code>AccessDenied</Code><Message>There were headers present in the request which were not signed</Message><Key>test.pcap</Key><BucketName>test</BucketName><Resource>/test/test.pcap</Resource><RequestId>asdf</RequestId><HostId>asdf</HostId></Error>
< Vary: Accept-Encoding
< X-Amz-Id-2: asdf
< X-Amz-Request-Id: asdf
< X-Content-Type-Options: nosniff
< X-Ratelimit-Limit: 94740
< X-Ratelimit-Remaining: 94740
< X-Xss-Protection: 1; mode=block
< Date: Wed, 20 Aug 2025 17:15:17 GMT
< 
* Failure writing output to destination
* Closing connection 0
```

As soon as I changed that it started working.

I haven't tested it out with AWS but I can give it a try.

awick: Ya that makes sense that it wouldn't work since "SignedHeaders=content-type..." is set. weird. I wonder if a curl bug or something else is wrong.

If you have time to try with AWS that would be great

mcgillowen: I've tried to get it working with and without these changes, but I can't get Arkime to authenticate correctly with AWS sqs.

I get:
```
Aug 21 08:15:32 reader-scheme-sqs.c:83 sqs_done(): Receive failed 403 {"Error":{"Code":"SignatureDoesNotMatch","Message":"The request signature we calculated does not match the signature you provided. Check your AWS Secret Access Key and signing method. Consult the service documentation for details.\n\nThe Canonical String for this request should have been\n'POST\n/772405016768/arkime-test\nAction=ReceiveMessage&MaxNumberOfMessages=1&Version=2012-11-05&WaitTimeSeconds=10\ncontent-type:application/x-www-form-urlencoded\nhost:sqs.eu-central-1.amazonaws.com\nx-amz-date:20250821T081532Z\n\ncontent-type;host;x-amz-date\ne3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855'\n\nThe String-to-Sign should have been\n'AWS4-HMAC-SHA256\n20250821T081532Z\n20250821/eu-central-1/sqs/aws4_request\n48a411ae990afc8005f8eb3fd57bd54e40b660b9f3438f061d1e1f09b3d3ead3'\n","Type":"Sender"},"RequestId":"8f80ff9d-02b3-5d15-8337-9ebdd557b976"}
```

But if I use curl on it's own with the same request and access/secret key it works, it seems that there might be something wrong with the signature calculation but I can't figure out what.

I've triple checked the keys, regenerated new ones but no change.

mcgillowen: The only difference between the 2 requests I've found is that the CLI curl has `accept` in the SignedHeaders and the curl request from within Arkime does not have `accept` in the SignedHeaders.

If I don't specify the `Accept: application/json` header in the CLI curl, it still works but I just get XML back instead of JSON, and SignedHeaders does not contain `accept`

mcgillowen: After testing against AWS S3&SQS, this change does not affect compatibility since the current implementation of S3&SQS scheme reader does not function with path access style buckets (buckets with periods in their names)

awick: Ok, so you think at this point should just merge it?

mcgillowen: Yes, if you don't mind

awick (approved):

PR #3318, opened 2025-08-27
add link to github release

31453:
## License

### DIFF
--- a/common/vueapp/Version.vue
+++ b/common/vueapp/Version.vue
@@ -4,11 +4,12 @@ SPDX-License-Identifier: Apache-2.0
 -->
 <template>
   <span>
-    <span
+    <a
       id="version"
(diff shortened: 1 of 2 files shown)


common/vueapp/Version.vue
```diff
@@ -40,6 +41,11 @@ export default {
       const dateMs = new Date(this.buildDate).getTime();
       const date = timezoneDateString(dateMs, this.timezone);
       return `${this.buildVersion}${!isNaN(dateMs) ? ' @ ' + date : ''}`;
+    },
+    versionLink () {
+      if (!this.version) return '';
+      if (!this.version.includes('-')) return this.version;
```
awick: I think this should maybe look for -GIT since I would like to support -rc1 etc if we do that for dev6. Thoughts?

PR #3406, opened 2025-10-21
Cont3xt - add DomainTools Whois integration

31453:
## License


cont3xt/integrations/domaintools/index.js
```diff
@@ -0,0 +1,178 @@
...
+        field: 'registration.registrar',
+        tooltip: 'registrar',
+        purpose: 'registrar',
+        precedence: 2,
+        order: 110
+      }
+    ]
+  };
+
+  cacheTimeout = 24 * 60 * 60 * 1000;
```
31453, line 134: @awick what do you think this should be?

awick: I think that is great

PR #3431, opened 2025-10-29
misc v6 bugs

31453:
- Copy values doesn't work over http, show alert and allow manual copy
- Fix spiview section toggle icons (- to +)
- Fix sessions value/label alignment by forcing min-height
- Fix PCAP download filename format
- Fix table column resize affecting sort order
- Remove unnecessary map=true parameters
- Hunt search expression error disables create hunt button
- Fix session packets error message for $t

## License

### DIFF
--- a/viewer/vueapp/src/components/utils/utils.js
+++ b/viewer/vueapp/src/components/utils/utils.js
@@ -134,32 +134,19 @@ export default {
     store.commit('setDisabledAggregations', false);
   },
 
+  /**
+   * Sets the map query based on the query time range and whether the user wants to force aggregations
+   * NOTE: this MUST be called after setFacetsQuery
+   * NOTE: mutates the query object and sets the store values
+   * @param {object} query The query parameters for the search to be passed to the server
+   */
   setMapQuery (query) {
-    // Default to no map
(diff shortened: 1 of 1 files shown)


31453:
@copilot review this pull request

viewer/vueapp/src/components/utils/utils.js
```diff
@@ -134,30 +134,33 @@ export default {
...
-    if (!page || page !== 'sessions') {
-      return;
-    }
+  setMapQuery (query) {
+    let outOfRange = false;
 
     // hide the map if the time range is out of bounds
     if (query.date === '-1') {
       query.map = false;
+      outOfRange = true;
```
awick: instead of outOfRange can you just do return?

awick: oh nm, ya I agree with copilot, this is confusing. :)

awick: i think just at the top of the function check the force and map area showing, set to true, and return, then you won't need out of range.

Right now this PR still sets map to true if force is true but the map area isn't shown.

PR #3441, opened 2025-11-05
Fix delete shard for unassigned shards

31453:
Use cancel command for unassigned shards instead of allocate_empty_primary which only works when shard is truly unassigned.
make stats search input wider

## License


awick:
I think for primary shards it really needs to do allocate empty shard, but for replicas the cancel makes sense. We should do more testing.

PR #3467, opened 2025-11-11
Provide a generic way to share different types of data, fixes #3466

### LINKED ISSUE #3466: shareables
For v6 lets create a new index called shareables that we use for all shareable things that don't have backend requirements. Such as summary, info fields, columns, etc. Might be nice to move views someday.

Fields:
name: keyword
type: keyword
creator: keyword
created: date
updated: date
viewRoles: keyword
viewUsers: keyword
editRoles: keyword
editUsers: keyword
data: object

APIs
* {success} create (everything)
* {success} update (_id, everything)
* {success} delete (_id) - only if creator or superUser
* {everything, canDelete} get (_id)- only if edit
* [{_id, type, name, data, canEdit, canDelete}] list (type, viewOnly=true) - returns all items with type, by default viewRoles/viewUsers only otherwise view or edit

awick:
## License

### DIFF
--- a/db/db.pl
+++ b/db/db.pl
@@ -5956,6 +5956,9 @@ sub shareablesUpdate
     "name": {
       "type": "keyword"
     },
+    "description": {
+      "type": "keyword"
+    },
     "type": {
       "type": "keyword"
     },
--- a/tests/api-shareables.t
+++ b/tests/api-shareables.t
@@ -1,4 +1,4 @@
-use Test::More tests => 60;
+use Test::More tests => 64;
 use ArkimeTest;
 use JSON;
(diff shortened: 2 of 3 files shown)


db/db.pl
```diff
@@ -5917,6 +5918,78 @@ sub configsCreate
...
+      "string_template": {
+        "match_mapping_type": "string",
+        "mapping": {
+          "type": "keyword"
+        }
+      }
+    }
+  ],
+  "properties": {
+    "name": {
```
31453, line 5956: add description and maybe a flag to turn off the sharing?

PR #3520, opened 2025-12-03
Fix PCAP/CSV export for visible sessions and download filename

31453:
- Fix "Download Entire Pcap" showing "undefined" filename by using pug attribute instead of Vue binding for download attribute
- Fix "Visible" export returning all results by respecting client's start/length parameters in sessionsListFromQueryChunky
- Fix export error toast by removing invalid this.$t() call in service

## License

### DIFF
--- a/tests/api-sessions.t
+++ b/tests/api-sessions.t
@@ -1,4 +1,4 @@
-use Test::More tests => 102;
+use Test::More tests => 106;
 use Cwd;
 use URI::Escape;
 use ArkimeTest;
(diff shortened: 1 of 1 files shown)


viewer/apiSessions.js
```diff
@@ -39,7 +39,7 @@ class SessionAPIs {
   // INTERNAL HELPERS
   // --------------------------------------------------------------------------
   static async #sessionsListFromQueryChunky (req, res, fields, startCb, chunkCb, endCb) {
-    if (req.query.length === undefined || parseInt(req.query.length) < 1000000) {
+    if (req.query.length === undefined) {
```
31453, line 42: @awick not sure why this was always defaulting to 1000000 🤷

awick: Because I wanted the min number to be huge, but I can go thru and make sure it is always being called correctly. I found at one point for example that not all tags were being set.

31453: ok, but this breaks the export csv and pcap stuff. it's overriding the length parameter when we only want the x num of visible sessions exported. Is this a good compromise?

awick: ok, looks good to me

PR #3742, opened 2026-02-23
Fix ArkimeTable not reacting to computed columns prop changes

### LINKED ISSUE #3370: Make Arkime Table computed so i18n works
Currently if Arkime Table callers are computed it still doesn't update the table headers (p1) and the table data (p2).

31453:
Add a watcher on the columns prop so that when i18n locale changes, computedColumns are updated in place

fixes #3370

## License


awick:
lgtm

PR #3878, opened 2026-04-10
Add JWT decoding support for header auth mode

31453:
When a reverse proxy (AWS ALB, Cloudflare Access, Google IAP) handles OIDC authentication, it forwards the identity as a JWT in a request header. Previously, header auth mode could only read the header value as a plain-text userId, requiring an external proxy to decode the JWT and re-set a plain header.

New config options:
```
  userNameHeaderJwt=true
  userNameHeaderJwtField=short_id

```
When enabled, the header value is decoded as a JWT (base64, no signature verification because the upstream proxy already verified) and the specified claim is extracted as the userId. JWT claims are also passed as `'vals'` to `[user-auto-create]` and `[user-role-mappings]`, enabling dynamic user provisioning and role assignment from token claims.

## License


awick:
Very cool! Need to look at it some more, but my first reaction is
* replace userNameHeaderJwt=true with authMode=header-jwt (this would be the first auth mode that you can't set without using authMode, which I want to require for Arkime 7 anyway)
* replace userNameHeaderJwtField with authUserIdField since that is what oidc already uses OR if you want a new field start it with auth

awick (dismissed):
LGTM, make sure you
* update CHANGELOG
* update arkimeweb

PR #3922, opened 2026-04-28
enforce hidePcap/disablePcapDownload on body and bodyhash endpoints

31453:
- Add User.checkPermissions(['hidePcap', 'disablePcapDownload']) middleware to /api/session/:nodeName/:id/body, /bodypng, /bodyhash, and /api/sessions/bodyhash/:hash, matching the existing /packets and /pcap* authorization model.
- Add api-sessions.t regression coverage: restricted users now receive 403 from all four body/bodyhash routes.

## License

### DIFF
--- a/tests/api-sessions.t
+++ b/tests/api-sessions.t
@@ -1,4 +1,4 @@
-use Test::More tests => 159;
+use Test::More tests => 167;
 use Cwd;
 use URI::Escape;
 use ArkimeTest;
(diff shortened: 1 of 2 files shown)


awick:
I think the body and bodypng should only depend on hidePcap since if you can see the pcap but can't download it I still want them to work. Since the bodyhash one is only used by the api we can leave those if you want.

PR #4029, opened 2026-06-10
Cycling help notes: welcome, what's new in v7, and per-page tips

31453:
## The owl now has opinions

Arkime has greeted new users with exactly one floating card since forever. This PR teaches that card to do what every other app's onboarding does: hold a deck of messages you can flip through, dismiss one at a time, nuke all at once, or shoo away until tomorrow.

### One card, eleven messages
\`HelpNotes.vue\` replaces \`WelcomeMessage.vue\` and cycles through: the classic welcome (new users), a one-time "Welcome to Arkime 7!" what's-new (upgraders), and nine per-page tips that deep-link into the Help page. Announcements always lead; once they're gone, the card opens on the tip for the page you're standing on.

### The three verbs
| Control | What it means | Lifespan |
|---|---|---|
| **Dismiss** | never show *this* message again | forever, per user |
| **Dismiss all** | never show *any* of them | forever, per user |
| **✕** | not now | until next visit |

### How state works
- New top-level user field \`dismissedHelpNotes\` (keyword array; \`'all'\` is the wildcard) — lives next to \`welcomeMsgNum\`, where this kind of state has always lived
- \`welcomeMsgNum\` is frozen as a back-compat seed: ≥1 means the welcome was already acknowledged, ≥2 the v7 notice too — existing users land exactly where they should with zero migration
- \`PUT /api/user/:userId/acknowledge\` gains a \`noteId\` mode (validated, deduped, capped); the legacy \`msgNum\` path is byte-identical

### Heads up
⚠️ **db version 86 → 87** — the users index is \`dynamic: strict\`, so run \`db.pl upgrade\` before deploying the viewer or dismissals silently won't stick.

Also along for the ride: a latent \`WelcomeMessage.vue\` bug (wrote to a nonexistent \`user.msgNum\` field) is gone with the component, and all 10 locale files carry translated strings for the new UI.

**Tested:** \`api-users.t\` grew 12 assertions covering the new endpoint mode; the full flow (cycle, wrap-around, dismiss, persist, dismiss-all, route-aware start) was driven end-to-end in a real browser against a dev cluster.


31453:
@awick help me with the text!

awick:
Is there anyway we could do this without db changes AND is compatible with v6? if not i think I'll add methods to get the current db version so we can enable/disable items.

PR #4033, opened 2026-06-11
Omit dns qc/qt fields when query class/type is unknown

31453:
The qc/qt emission added in #4024/#4025 uses an unconditional `%s`, so a root query whose class/type lookup fails prints the literal `(null)` into the session JSON. Guard each field and skip it when unset, and drop the two `(null)` lines from `tests/pcap/v6-http.test` — mirroring arkime/ja4@f18adec1, which already removed them from the ja4 copy of the same expectation.

Found via PR #4029's CI, which was the first run to clone the ja4 repo after that expectation change; dev7's next push would fail the ja4plus job the same way without this.


31453:
Superseded by #4032, which fixes the same (null) emission with an RFC 3597 CLASS%u/TYPE%u fallback — it reaches dev7 via the next main merge.

PR #4058, opened 2026-06-22
Fold Summary/Sessions/SPIView/SPIGraph into one tabbed Explore page

31453:
Combines the Summary, Sessions, SPIView, and SPIGraph pages into one tabbed **Explore** page that shares a single search bar, time range, and visualization (same pattern as #4039's Connections fold). The four navbar items collapse into one "Explore"; all four existing URLs still work and select their tab. Defaults to the Summary tab and a 15-minute query time.

### DIFF
--- a/common/vueapp/locales/de.json
+++ b/common/vueapp/locales/de.json
@@ -111,6 +111,8 @@
   },
   "navigation": {
     "arkime": "Arkime",
+    "explore": "Erkunden",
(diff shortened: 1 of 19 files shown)


31453:
I don't actually like this. But if you do, I'll add polish @awick

awick:
i have thoughts

PR #4065, opened 2026-06-23
Preserve redirect URL query parameters across OIDC login

yanover:

When using `authMode=oidc`, the first unauthenticated request to a URL with query parameters (e.g. `/sessions?expression=...&startTime=...&stopTime=...`) loses all parameters after the OIDC round-trip. The user is redirected to `/sessions` instead of the original URL. 

The second click works because the session already exists.

**Fix**
1. Exclude `/auth/login/callback` from overwriting ogurl

2. Capture ogurl in a local variable before `passport.authenticate` runs a local variable survives session regeneration, unlike `req.session.ogurl`.

3. Try multiple sources to find ogurl, instead of only checking `req.body.ogurl`, check three sources in order and use the first one that has a value:
   - `req.body.ogurl` : form auth submits it as a hidden POST field (unchanged behavior)
   - `req.session.ogurl` : still set if Passport didn't regenerate the session
   - `savedOgurl` : the local variable captured before `passport.authenticate`, survives session regeneration (fixes OIDC)

yes

## License

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -39,6 +39,8 @@ NOTICE: Create a parliament config file before upgrading (see https://arkime.com
 ## Release
   - #4048 Node 22.23.0
   - #4056 Add Fedora 44 builds
+## All
+  - #4065 Fix OIDC login not preserving the redirect URL's query parameters (thanks @yanover)
 ## Capture
   - #4054 Fix GRE keepalive packets adding bogus 00:00:00:00:00:00 MACs
   - #4055 Fix DNS memory leak when parsing malformed answers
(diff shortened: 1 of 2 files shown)


awick:
thanks! can you test the change I made to see if it works for you? Reviews have been complaining for a while that we don't do any basic security checks, so it seemed like a good time to add.

yanover:
> thanks! can you test the change I made to see if it works for you? Reviews have been complaining for a while that we don't do any basic security checks, so it seemed like a good time to add.

Hey ! Ty for the addition.

Tested and works great !

The security check doesn't break the legitimate OIDC flow,  `ogurl` is always a relative path, so it passes the guard you've added without issue.

Issue #3971, opened 2026-05-12
Arkime Tab v2


awick:
## Overview
The current Arkime Dashboard tab provides a fixed set of visualizations. To increase utility and flexibility, this proposal shifts toward a **modular, widget-based dashboard**. Users should be able to build a customized workspace by adding, configuring, and arranging various visualization elements (graphs, maps, SPI data) that persist across sessions.

## Core Objectives
- **Flexibility:** Move from a "one-size-fits-all" view to a customizable layout.
- **Granularity:** Shift configurations (limits, fields, types) from the global header to individual widgets.
- **Efficiency:** Allow saved expressions per widget to enable multi-layered analysis in a single view and only ask for the fields selected to improve aggregation time

---

### 1. Modular Widget System
Users should be able to add multiple types of elements to their dashboard. Each element is an independent "Widget."
* **Widget Types:** Support for all existing Arkime visualizations:
    * Timeline/Current Graphs
    * SPI Graphs (Tree, Matrix, Donut, ...)
    * Map Views
* **Per-Widget Configuration:** Each widget should feature an **Edit Button** to configure:
    * **Field Selection:** (e.g., `srcIp`, `host.name`, `protocols`).
    * **Metric Basis:** (e.g., Sessions, Packets, Bytes, or Lookups).
    * **Data Limit:** Number of items to display (e.g., Top/Bottom 10, 20, 50).
    * **Local Filtering:** Ability to save a specific **Search Expression** or **View** per widget. This filter is applied *on top* of the global session search.

### 2. Layout and Grid UI
To keep the dashboard organized, the UI should follow a structured grid system:
* **Column Control:** Toggle between a **2-column** or **3-column** layout.
* **Sizing:** Widgets should have a fixed width based on the column count. Height should be configurable (e.g., Standard or Double Height). Maybe support double width?
* **Internal Scrolling:** To prevent the page from becoming excessively long, widgets should have **internal scroll bars** for high-density data (e.g., showing 50 bar chart items).

### 3. Persistence and Management
* **Saved Dashboards:** Users can save the entire configuration (layout + widget settings) as a "Saved Summary."
* **Default Dashboard:** Option to set a specific saved dashboard as the default landing page for the tab.
* **Searchable Selector:** The dashboard selector should include a search bar and the top X number of used dashboards.

---

## Example Scenario
A user creates a "Security Operations" dashboard with the following widgets:
[...]

PR #4077, opened 2026-06-26
Arkime Tab v2: modular widget dashboard (#3971)

31453:
Implements #3971 — the Arkime tab is now a fully modular, per-widget dashboard. **Every** visualization — including Map, Timeline, and the capture stats — is a configurable, resizable, reorderable widget, persisted as a saved dashboard. Built on the existing summary/shareables infrastructure; the legacy summary path stays backward compatible.

## Widget types
Each widget has a per-widget editor that **leads with visualization type** and grays out inputs that don't apply to it:
- **Field-bound:** Bar · Pie · Table · **Intersection** (SPIGraph unique/combination table) · Heatmap (a field's values over time) · Treemap
- **Session-wide:** Timeline · Map · Capture Statistics · Time Information — fed by the host's single global stats fetch (the old pinned viz band + stats band are gone; the default dashboard seeds these).

## Highlights
- **Per-widget config** — field(s), metric, Top/Bottom N, width/height, and a local filter (saved View and/or expression) ANDed with the global search.
- **Numeric-field metric** — bar/pie/table/heatmap/treemap visualize Sessions or any numeric field (summed per value; Top-N ordered by it).
- **Multi-field (up to 3)** — a chips field selector for pie/treemap/intersection/table. Pie → nested **sunburst**, treemap → **nested rectangles**, intersection → **nested combination table**; table shows each field's top-N **side by side**.
- **Layout** — fixed 4-column grid on a 160px row unit; widgets span 1–4 columns / 1–8 rows; **drag-to-reorder** and **drag-to-resize** (right/bottom edges + corner).
- **Configurable palette** applied across all charts.
- **Saved dashboards** — `summaryConfig` shareable, JSON import/export, searchable selector, default landing dashboard.
- The dashboard **always aggregates** (facets + map) so the timeline/map populate on any range, bypassing the large-dataset "fetch viz" toggle.
- Per-widget **info tooltip** + discrete Edit / Export / Remove buttons.

## Backend (additive; legacy intact)
- `/api/sessions/summary` — per-widget `widgets[]` (length/order/metric/local expression), `noStats`, multi-field batch.
- `/api/spigraph` — optional numeric `metric` param (heatmap intensity / treemap size).
- `/api/spigraphhierarchy` — backs the nested intersection / pie / treemap (1–3 fields).

## Notes / discussion
- Nested **pie/treemap/intersection are count-based** (the hierarchy endpoint carries no metric); the metric applies to the table and single-field charts.
- **Bar and Heatmap are intentionally single-field** (like Bar, Heatmap is single-dimension — one field's values over time). Combination questions are better served by a widget's local filter + the Intersection table than by cramming nested rows into a value×time matrix.
- Docs in `SHAREABLES.md`; i18n added; tests extended (`api-sessions.t`, `api-summary.t`, `api-spigraph.t`, `api-spigraphhierarchy.t`).

Kept as a **draft** — still iterating on the UX before review.

🤖 Generated with [Claude Code](https://claude.com/claude-code) and greatly reduced by @31453


31453:
One design point worth chatting through (also in the description under **Design decision to discuss**): I kept **Map** and **Timeline** as the pinned, always-on visualization rather than making them addable widgets. Every dashboard widget is *field-bound* (top-N of one field, with a per-field view mode + metric); map/timeline are session-wide and don't fit that model, so adding them as widgets would need a separate widget shape. Curious whether you'd rather fold them in or keep them pinned.

PR #4103, opened 2026-07-05
Time picker uses ISO year-first format with paste support and a date …

awick:
…picker button

- Start/End are text inputs displaying YYYY/MM/DD HH:mm:ss like Arkime 5
- Typed or pasted dates accepted in Y/M/D, Y-M-D, and ISO forms; invalid input resets to the current value
- Calendar button opens the native browser date picker
- New search.openDatePickerTip key added to all locales

## License


31453 (approved):
Reviewed and tested locally against the sessions page. Paste/parse of Y/M/D, Y-M-D and ISO dates works, the calendar button opens the native picker, and the YYYY/MM/DD HH:mm:ss display is back. Pushed a small tweak (3e994c83a) moving the calendar button to after the timezone info. LGTM.

PR #4108, opened 2026-07-07
Validate full s2s auth token on the sessions receive route

31453:
Hardens the `/api/sessions/receive` gate to fully validate the s2s auth token (path/user/date/method) instead of only checking that it decrypts. The validation is shared with the s2s passport strategy via a new `Auth.validateS2SObj`.


awick:
LGTM, please add a CHANGELOG entry

PR #4109, opened 2026-07-07
Fix tshark pcap fetch double-encoding the session id

31453:
## License

### DIFF
--- a/viewer/apiSessions.js
+++ b/viewer/apiSessions.js
@@ -2989,7 +2989,7 @@ class SessionAPIs {
         });
       } else {
         // makeStreamRequest already encodeURI()s the path; encodeURIComponent on the id would double-encode '@' (%40 -> %2540) and 500 the owning node
-        const remotePath = `/api/session/${encodeURIComponent(req.params.nodeName)}/${req.params.id}/pcap`;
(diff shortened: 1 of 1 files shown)


viewer/apiSessions.js
```diff
@@ -2988,7 +2988,8 @@ class SessionAPIs {
           if (writeStream && writeStream.writable) { writeStream.end(); }
         });
       } else {
-        const remotePath = `/api/session/${encodeURIComponent(req.params.nodeName)}/${encodeURIComponent(req.params.id)}/pcap`;
+        // makeStreamRequest already encodeURI()s the path; encodeURIComponent on the id would double-encode '@' (%40 -> %2540) and 500 the owning node
+        const remotePath = `/api/session/${encodeURIComponent(req.params.nodeName)}/${req.params.id}/pcap`;
```
awick: I think it should skip encoding nodeName also then?

31453: I don't know. I debated. Node names cant have chars like sessionIds to encode multiple times and I KNOW that the session id is already encoded at this place, but not sure about the node name.

awick: The comment says it does the whole path, lets trust the AI and remove from node :)

PR #4144, opened 2026-07-22
Mask packed SCTP stream id in HTTP/2 parser to fix heap OOB access

31453:
For SCTP traffic, `which` packs the stream id into its upper bits (`ARKIME_WHICH_SET_ID`). After an HTTP/1.1 `Upgrade: h2c` transition on a non-zero SCTP stream, this packed value reaches `http2_parse` and is used directly to index the `HTTP2Info_t` arrays, which are sized `[2]` (one per direction) — a heap out-of-bounds read/write for any stream id > 0 (e.g. stream 1 → `which` = 256).

Mask `which` back to the direction bit at the top of `http2_parse`, and apply the same mask when storing `http2->which` in `http2_classify` so the request/response direction comparison stays correct. Both are no-ops for plain TCP.

### DIFF
--- a/capture/parsers/http2.c
+++ b/capture/parsers/http2.c
@@ -416,6 +416,8 @@ LOCAL int http2_parse(ArkimeSession_t *session, void *uw, const uint8_t *data, i
 {
     HTTP2Info_t            *http2          = uw;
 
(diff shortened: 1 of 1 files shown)


awick:
Lol, vulnhunter was just complaining about this too. I think #4145 i think is a more complete fix for now.

PR #4178, opened 2026-08-04
Record whether a TCP handshake actually completed

jeffreygolden:
## Problem

`tcpflags.*` counters are per-session and direction-agnostic, so `syn > 0 && syn-ack > 0 && ack > 0` does not mean a handshake completed. All three can be satisfied by:

- syn-ack retransmits to a peer that never replied
- two connection attempts merged into one session (a bare SYN with a different ISN doesn't split the session unless a RST or FIN was already seen — `tcp_pre_process`)
- an ack that acknowledges nothing valid, including a crafted one

There's currently no field recording whether the sequence numbers line up, so a session that never synchronized looks identical to one that did. That makes broken TCP sessions hard to troubleshoot, and makes a fabricated session indistinguishable from a real one — a sender that never received the syn-ack cannot know the responder's ISN (RFC 6528), which is the only unforgeable signal available.

## What this adds

Two independent booleans, each recording one thing that was actually checked:

| field | true when |
| --- | --- |
| `tcp.synValidated` | a syn-ack acknowledged the ISN of a SYN observed in this session (`SYN-ACK.ack == SYN.seq + 1`) |
| `tcp.ackSynchronized` | an ack acknowledged the other side's syn-ack ISN (`ACK.ack == SYN-ACK.seq + 1`), and continued from the observed SYN (`ACK.seq == SYN.seq + 1`) when a SYN was observed |

Both true is a fully validated three-way handshake, and each is useful alone:

| synValidated | ackSynchronized | |
| --- | --- | --- |
| true | true | handshake validated |
| true | false | SYN seen and answered, never completed |
| false | true | synchronized, SYN not observed (capture started mid-flow) |
| false | false | nothing proven |

Two booleans rather than one state field, so adding a check later can't change the meaning of existing values or break saved queries.

Also adds `tcp.srcISNCnt` — distinct initial sequence numbers seen from the source. Normally 1, 0 when no SYN was observed, greater than 1 when multiple connection attempts were merged into one session. Useful on its own for spotting NAT port reuse and merged flows.

## Notes

- **Fails closed.** With no syn-ack observed from the other side, nothing is claimed. These fields never guess.
- **Session splitting is unchanged.** The ISN-divergence test in `tcp_pre_process` now gates the new counter, and the existing RST/FIN test still gates the split, so session boundaries are byte-identical.
- **Per-direction state**, so a crafted syn-ack from the peer can't supply both halves of a comparison.
- **Cost:** three added comparisons on values already in the same cache lines, in branches that already invoke `ARKIME_RULES_RUN_FIELD_SET`. Session struct grows 400 → 424 bytes.

## Results

Seven synthetic sessions (RFC 5737 documentation addresses), replayed through a patched build:

| case | syn/syn-ack/ack | synValidated | ackSynchronized |
| --- | --- | --- | --- |
| genuine handshake | 1/1/1 | true | **true** |
| ack with wrong ack number | 1/3/1 | true | false |
| two SYNs, different ISNs | 2/2/1 | true | false |
| SYN + bogus ack only | 1/1/1 | true | false |
| crafted syn-ack + matching ack, both from peer | 1/2/1 | true | false |
| correct ack number, seq not continuing from the SYN | 1/1/1 | true | false |
| no SYN observed, ack does acknowledge the syn-ack | 0/1/1 | **false** | true |

Every one satisfies `syn > 0 && syn-ack > 0 && ack > 0`. Only the first is a real handshake.

Verified identically on `v6.5.0` and on `main`. I left the test capture and harness wiring out of this PR since `tests/` has its own conventions — happy to add both if you'd like them here.


awick:
Thanks! I might not get a chance to look at this until next week. I'm curious if databytes > 0 will answer part of your question?

jeffreygolden:
No worries on the timeframe, we might run a fork until this becomes part of an official release. I started with databytes, seemed like a strong signal, but I was able to find cases in our data where databytes = 0 _and_ we had a full TCP handshake. It's mostly VNC full-connect scans.

awick:
If you have a test pcap please add to tests/pcap or DM to me, don't worry about the golden files I'll regenerate.

PR #4190, opened 2026-08-11
update urlhaus and malwarebazaar integrations

m-jingu:
abuse.ch now requires an Auth-Key for the URLHaus and MalwareBazaar APIs, the same key ThreatFox already asks for (free at https://auth.abuse.ch/). Neither integration sends one today, so every lookup comes back `401 Unauthorized`.

While fixing that I noticed both integrations also have an empty `card.fields`, so even a successful lookup renders an empty card. This PR addresses both.

## Changes

**Auth-Key support** (mirrors the existing ThreatFox integration)

- add a `key` setting (`password: true`, `required: true`)
- read it with `getUserConfig(user, 'key')` and send it as the `Auth-Key` header
- return early when no key is configured, so the integration is skipped rather than repeatedly failing with 401

**`card.fields`**

- `urlhaus`: `query_status`, `firstseen`, `url_count`, `urlhaus_reference`, plus a table over `urls[]` (url, status, threat, date added, reporter, tags, reference)
- `malwarebazaar`: `query_status` plus a table over `data[]` (file name, signature, file type/size, sha256, md5, first/last seen, delivery method, reporter, tags)

Field names follow the abuse.ch API responses, and the layout follows the ThreatFox card. `pivot: true` is set on the URL and hash columns so they can be searched directly from the results.

## Testing

Verified on a live Arkime 6.6.0 deployment (Cont3xt on EL8, Elasticsearch 7.17.29):

- before: every lookup logged `AxiosError: Request failed with status code 401`
- after, with a key configured: `/api/integration/stats` shows `directError: 0` for both, with `directGood: 5` (URLHaus) and `directGood: 1` (MalwareBazaar), i.e. real results are returned
- results now render in the UI; previously the card came back empty even when data was fetched
- with no key configured the integration is skipped, as intended
- `npm run lint` passes

I did not run `tests.pl` since this only touches Cont3xt integration files.

## License

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -62,6 +62,8 @@ NOTICE: Create a parliament config file before upgrading (see https://arkime.com
   - #4187 Fix crash when redefining the type of a builtin field via wise or custom-fields
   - #4188 Add wiseRequestTimeout setting, default 30 seconds, so a stuck wiseService doesn't delay capture shutdown for minutes
 ## Cont3xt/Viewer
+  - #4190 Cont3xt URLHaus and MalwareBazaar integrations now require an API key (free at https://auth.abuse.ch/)
+  - #4190 Fix Cont3xt URLHaus and MalwareBazaar rendering an empty card by adding their missing card fields
   - #4134 Add an MCP server at POST /mcp for viewer and cont3xt, off by default, enable with mcpEnabled.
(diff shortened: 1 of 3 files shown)


awick:
thanks!

PR #4213, opened 2026-08-21
Store column, info field and spiview layouts as shareables

awick:
Moves the three per-user layout types off the user record and into the shareables index, so they can be shared with users and roles like anything else — and so there's one way to store all of them.

| type | was | now |
|---|---|---|
| Column Layout | `user.columnConfigs` | `sessionsTableLayout` shareable |
| Info Field Layout | `user.infoFieldConfigs` | `sessionsInfoLayout` shareable |
| SPI View Layout | `user.spiviewFieldConfigs` | `spiviewLayout` shareable |

Sharing a column layout was previously impossible; it's the most obviously useful of the three ("use the team's standard session columns").

## Migration

`db.pl <host:port> shareables-import <type|all>`

Deliberately **not tied to a db version** — it's a plain command that can be run whenever and as often as you like:

- **Idempotent** on `type + creator + name`, so re-running copies nothing new
- **Leaves the originals in place**, so Arkime 6 keeps working against the same cluster during a trial
- **Re-runnable**, so work done in 6 after the first import can be picked up later
- `--dryrun` to preview

```
=== --dryrun ===        3 would be copied, 0 already existed
=== real run ===        3 copied, 0 already existed
=== re-run ===          0 copied, 3 already existed
=== new layout made in 6, re-run ===
Copying spiview 'Made In Six' for anonymous
1 copied, 3 already existed
```

Copy only, never the other direction — 7 never writes back to the 6 location. Adding a fourth type later is one entry in the `%SHAREABLE_IMPORTS` table.

## Settings tabs are now editable

The three layout tabs were read-only lists with a delete button. They now share a `LayoutEditor` dialog covering all three shapes:

- **Column layouts** — field list plus a sort order with per-field direction
- **Info field layouts** — plain field list
- **SPI View layouts** — the `field:count,...` string, parsed into chips and reassembled with the per-field value counts preserved

Field selection reuses the existing `FieldSelectDropdown`. Edit and delete are hidden on layouts shared with you read-only, which wasn't expressible before.

Sharing itself is still edited in the Shareables tab, which stays the one place responsible for permissions.

## Notable fix

`createShareableService.update` always sent all four share arrays with `|| []` defaults, so a partial update — like SPI View saving just its fields — would silently strip the sharing off a shared item. It now sends only the keys given, which the API already treats as "leave the rest alone".

## Verification

- Full suite **5753 passing**, lint clean, bundle builds
- `api-shareables.t` extended to 159 tests, covering the import for all three types: dry run copies nothing, real run produces correctly-shaped private shareables, originals survive on the user doc, re-runs don't duplicate
- Exercised in the browser: all three editors round-trip (rename, add/remove fields, flip sort direction), Sessions and SPI View read and write shareables and no longer call `/api/user/layouts/*`, no console errors
- All 10 locale files stay valid, key-identical and line-for-line aligned

## Notes for review

- `/api/user/layouts/:type` is intentionally left in place — it's what keeps Arkime 6 working against the same cluster. Removing it is cleanup for whenever 6 compatibility is dropped, along with the leftover fields on the user record.
- An upgrading user sees empty layout lists until someone runs the import, and nothing currently prompts them. The CHANGELOG note covers it; a startup hint when unmigrated layouts are detected might be worth adding.
- Name uniqueness used to be enforced per user per type. As shareables that no longer holds — two people can both have "My Layout", and one shared with you can collide with your own.


awick:
Pushed two follow-ups.

## Removed the layouts API

Since Arkime 6 runs its own viewer.js rather than this one, everything behind `/api/user/layouts/:type` was dead once the UI moved to shareables. Removed:

- all four `/api/user/layouts/:type` routes and their handlers
- the private helpers behind them — `#userColumns`, `#userInfoFields`, `#userSpiview`, the three validators, the three setters, the three updaters and `#deleteLayout`
- the layout parts of `/api/user/config` — `sessions` no longer returns `colConfigs`/`infoConfigs`, `spiview` no longer returns `fieldConfigs`
- the four client `UserService` layout functions

**The data stays on the user record** — that's what Arkime 6 reads, and the import copies rather than moves.

Net **−809 lines**. Verified the routes are gone by confirming they now return the same SPA catch-all as a known-bogus path.

The layout CRUD tests in `api-users.t` went with the API (46 assertions), replaced by five asserting the config endpoint no longer carries layouts. `api-shareables.t` now seeds its v6 fixture by writing the user record directly, which is more faithful to what an upgrade actually finds.

## Shared layouts sort last, with an icon

In all three dropdowns — Sessions column configs, Sessions info fields, SPI View field configs — your own layouts come first, then ones shared with you, each still in the API's name order. Shared entries get an `mdi-account-multiple` icon whose tooltip names the creator.

Sorting happens once in `createLayoutService.list()`, so every consumer gets it without repeating the logic, and `shareableToLayout` now carries `shared` and `creator` through.

```
Arkime Default
My Own            (no icon)
Team Standard     (shared icon, "Shared with you by someoneelse")
```

Full suite **5712 passing**, lint clean.

31453:
While reviewing, I pushed a few small fixes (permission gates on shared layout save/delete, tracking the loaded layout by id instead of name, server-side data validation, the 50-item list cap, unhandled rejections in `getSessionsConfig`).

One piece is more of a judgment call, flagging it separately: in `db.pl shareables-import`, the dedup key (`type+creator+name`) assumes a layout name is still unique per user, but this PR removes that uniqueness. Since the import is explicitly designed to be run "whenever" during an extended 6/7 trial (per the re-run example in the PR description), there's a real window where a user creates a new same-named shareable in 7 before their same-named legacy layout gets imported — the importer would then skip it as "already existed" and silently drop it for good.

This only bites in that specific ordering, so it's optional rather than blocking. I included a fix (compares data, not just name, and imports a colliding legacy layout under a disambiguated name with a warning instead of dropping it) — take it or leave it.

31453 (approved):

PR #4215, opened 2026-08-24
Fix role-based settings being ignored outside common/user.js

Juwon1405:

`expandFromRoles()` stores the expanded permission settings (`emailSearch`, `removeEnabled`, `packetSearch`, `hideStats`, `hideFiles`, `hidePcap`, `disablePcapDownload`) in the private `#allSettings` field. `User` exposes `getRoles()` for `#allRoles` and `getExpression()` for `#allExpression`, but there is no accessor for the settings, so code in other files reads the public property on the user object instead.

That property holds only what the user document sets. When a role is what grants the setting, it is `undefined`, and these three call sites treat that as `false`:

- `viewer/buildQuery.js` — `emailSearch: req.user.emailSearch === true`
- `viewer/apiCrons.js` — `emailSearch: user.emailSearch === true`
- `cont3xt/audit.js` — `if (!req.user.removeEnabled)`

The visible effect: a user whose document does not set `emailSearch` cannot search email fields even when one of their roles enables it, while `checkPermissions()` and `GET /api/user` report the permission as granted, because those read `#allSettings`. The two answers disagree for the same user, which makes it look like the Users page is showing something other than what the server enforces.

This shows up on any deployment that keeps permissions on roles and leaves the user documents free of per-user setting fields — for instance when `userAutoCreateTmpl` does not list them, so auto-created users only ever get the values their role provides.

The fix adds `getSetting()`, mirroring how `getExpression()` exposes `#allExpression`, and switches those three call sites to it. `#allSettings` is initialised to `{}`, so calling it before expansion returns `undefined`, exactly as reading the property did.

I deliberately did not make `expandFromRoles()` write the expanded values back onto the instance properties, which would have been a shorter fix.

`save()` hands the instance itself to `User.setUser()`, so inherited values would be persisted onto the user document by ordinary saves — a `lastUsed` update is enough — and role-based settings would slowly turn into per-user ones.

Please note this is a behaviour change for affected deployments: settings that a role grants now take effect at these call sites where they previously did not.

None — happy to open one first if you would prefer to discuss it there.

Yes. `npm run lint` from the top level passes with no errors. The first draft used `name` as the parameter and lint caught it as `no-shadow`; the parameter is now `setting`.

No — I do not have an OpenSearch/Elasticsearch instance with the test PCAPs loaded in the environment where I prepared this, so I could not run the viewer test suite honestly. The change is one new accessor plus three call-site swaps, with no change to how `#allSettings` is computed.

I did verify the behaviour against a live user database, by loading `common/user.js` directly and comparing three things per user: `getSetting('emailSearch')`, the expression `buildQuery.js` evaluates, and the `checkPermissions()` verdict. Before the change the first two disagreed with the third for every user whose document did not set the field; after it, all three agree.

If you would like the test suite run before merging, please point me at it and I will set the environment up.

## License

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -36,6 +36,8 @@ NOTICE: Cross-cluster Shortcuts require you to not restart all your viewers at o
 NOTICE: Create a parliament config file before upgrading (see https://arkime.com/settings#parliament and https://arkime.com/faq#how_do_i_upgrade_to_arkime_5)
 
 6.7.1 2026/08/xx
+## All
+  - #4215 Fix email searches, periodic queries and cont3xt history deletes ignoring permissions that come from a role instead of the user (thanks @Juwon1405)
 
 6.7.0 2026/08/19
 ## Release
--- a/common/user.js
+++ b/common/user.js
@@ -1774,6 +1774,18 @@ class User {
     return this.#allRoles;
   }
 
+  /**
+   * Return a setting expanded from ourselves and the roles we have.
+   *
+   * #allSettings is private, so code outside this file cannot read it and has
+   * to use this accessor. Reading the public property directly returns the
+   * value stored on the user document, which silently ignores roles when the
+   * user document does not set that field.
+   */
+  getSetting (setting) {
+    return this.#allSettings[setting];
+  }
+
   getExpression () {
     return this.#allExpression;
   }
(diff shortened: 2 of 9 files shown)


awick:
thanks!

PR #4216, opened 2026-08-24
Keep multi viewer answering when one cluster is down

Juwon1405:

Running a multi viewer over a number of clusters, we kept hitting cases where one unhealthy backend took down queries the remaining clusters could still have answered. Three separate things combine to cause that.

**1. `makeRequest` puts no timeout on the outgoing request.**

A backend that accepts the socket but never answers fires neither `end` nor `error`, so `done()` is never called. The `async.map` in `simpleGather` then never completes, `res.send` is never reached, and the viewer request hangs forever. Once enough of them pile up they exhaust the agent's `maxSockets`, and at that point the whole multi viewer stops answering — `/api/eshealth` included, which is what makes this look like a total outage rather than one bad cluster.

Added a timeout on the request. On expiry the request is destroyed and the backend is reported as an empty result, which the code below already knows how to handle. It is configurable with `multiESBackendTimeout` (seconds, default 120) — I am happy to change the name or drop the setting and hard-code it if you would rather not add config surface.

**2. `simpleGather` hands failed backends to the combiners.**

A backend that failed arrives as a stub carrying only `.cluster`, with none of the keys the combiners dereference (`hits`, `nodes`, `tasks`, `indices`, ...). Every combiner fed from `simpleGather` — `_search`, `_msearch`, `_nodes`, `_tasks`, `_stats`, `_aliases` — throws on it.

Rather than guarding each combiner, the stubs are dropped in the single place all of them are fed from. Two properties keep this conservative:

- if every backend answered, `good.length === results.length` and it is a no-op, so healthy deployments behave exactly as before;
- if every backend failed, the results are passed through untouched, leaving the existing per-endpoint handling in place.

**3. `hits.total` is not a number any more.**

On Elasticsearch 7 and OpenSearch it is `{value, relation}`, so `obj.hits.total += result.hits.total` concatenates objects into a string and the merged total comes out as `"0[object Object][object Object]..."`. `addTotal()` adds the numbers while keeping the accumulator's shape, so when both sides are plain numbers it is the same addition as before.

This one is independent of the other two and reproduces on any multi viewer sitting in front of ES7/OpenSearch.

None — happy to open one first if you would prefer to discuss the approach, particularly the config setting in (1).

Yes, `npm run lint` from the top level passes with no errors.

No — I do not have an OpenSearch/Elasticsearch instance with the test PCAPs loaded in the environment where I prepared this, so I could not run the viewer suite honestly. `tests/api-multies.t` is the one I would expect to matter here.

What I can say is that all three changes have been running on a multi-cluster multi viewer, where they turned a repeated whole-UI hang into "one cluster is missing from the results" and fixed the corrupted totals. Please do run the suite before merging.

## License


awick:
Thanks! I think these need some work.

To run the test suite you just need to do `./tests.pl --viewer` and it will load the pcap and run the suite

1. This only is reachable for some calls, I think we should do a more complete fix
2. should this handle error responses too?
3. did you actually hit this issue, or is this just bad AI? I thought we request rest_total_hits_as_int everywhere, so please share what API is causing this issue, otherwise this can be removed.

awick:
Hi, I refactored the code some and removed #3, if you can respond with more info we can add back. Thanks for the PR.

PR #4222, opened 2026-08-28
Fix success toast showing [object Object] after PCAP/CSV export

31453:
## Summary

After exporting sessions to PCAP or CSV, the success toast displayed the literal text `[object Object]` instead of a real message (or nothing).

`SessionsService.exportPcap`/`exportCsv` trigger the download via a hidden link click rather than a real fetch response, so they resolve with a bare `{ success: true }` object — no `text`, `message`, or `i18n` field. The shared `resolveMessage()` helper fell back to `String(obj)` whenever none of those fields were present, and `String({ success: true })` evaluates to the literal string `"[object Object]"`, which then got rendered verbatim in the toast.

## Fix

`resolveMessage()` now passes strings through unchanged and returns `''` instead of stringifying the object when there's nothing to show. This matches how the vast majority of call sites already use the helper (`resolveMessage(x, t) || 'default message'`), so callers that want a message on success (or a fallback on error) already handle the empty case correctly — no other call sites needed to change.

## Test plan
- [x] `npx eslint common/vueapp/resolveI18nMessage.js` passes
- [ ] Export a PCAP and a CSV from the sessions view and confirm the success toast no longer reads `[object Object]`

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -42,6 +42,7 @@ NOTICE: Create a parliament config file before upgrading (see https://arkime.com
   - #4217 Fix HTTP/2 sessions leaving http.uri empty, only http.uri.path was filled in
 ## Viewer
   - #4216 Fix multi viewer hanging or failing queries when one cluster is down or unhealthy (thanks @Juwon1405)
+  - #4222 Fix session PCAP/CSV export success toast showing "[object Object]"
 
 6.7.0 2026/08/19
 ## Release
--- a/common/vueapp/UserDropdown.vue
+++ b/common/vueapp/UserDropdown.vue
@@ -158,7 +158,7 @@ export default {
         }
       }).catch((error) => {
         this.loading = false;
-        this.error = resolveMessage(error, this.$t);
+        this.error = resolveMessage(error, this.$t) || this.$t('users.unknownErr');
       });
     },
     updateUsers (userId) { // emits both the new array and changed user-value
--- a/common/vueapp/Users.vue
+++ b/common/vueapp/Users.vue
@@ -816,7 +816,7 @@ export default {
         this.dbUserList = JSON.parse(JSON.stringify(response.data));
       }).catch((error) => {
         this.loading = false;
-        this.error = resolveMessage(error, this.$t);
+        this.error = resolveMessage(error, this.$t) || this.$t('users.unknownErr');
       });
     },
     reloadUsers () {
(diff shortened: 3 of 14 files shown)


31453:
## Code review: confirmed findings

Reviewed this diff at extra-high effort (10 finder angles + independent verification passes). The fix itself is correct for its stated goal, but the generalized `resolveMessage()` fallback (`String(obj)` → `''`) introduces silent-failure regressions at a few existing call sites that relied on the old fallback always being truthy.

### 1. [`common/vueapp/Users.vue:819`](https://github.com/arkime/arkime/blob/main/common/vueapp/Users.vue#L819) (and `:840`) — user-list load failures now show no error banner
`loadUsers()`/`reloadUsers()` do `this.error = resolveMessage(error, this.$t)` with no `||` fallback. `apiGetUsers` (`common/user.js`) returns `{ success: false, recordsTotal: 0, recordsFiltered: 0, data: [] }` with no `.text`/`.message`/`.i18n` on its invalid-query and ES-search-throw paths. Before this diff, `this.error` became the ugly-but-visible `"[object Object]"`; after, it's `''`, and the template's `v-if="error"` (line 73) hides the banner entirely — the failure is now completely silent to the admin.

### 2. [`common/vueapp/UserDropdown.vue:161`](https://github.com/arkime/arkime/blob/main/common/vueapp/UserDropdown.vue#L161) — role-picker dropdown swallows load errors silently
Same mechanism via `apiGetUsersMin`'s catch block, which also returns a response with no text field on search failure. The template's `v-else-if="error"` (~line 63) skips the warning block for a falsy `error`, so the dropdown (used app-wide for role assignment) silently renders 0 users with no explanation.

### 3. Altitude concern: fix scoped to the shared helper instead of the two actual offending callers
`resolveMessage()` is used by ~120 call sites across `common/vueapp` and `viewer/vueapp`. The root cause is that `SessionsService.exportPcap`/`exportCsv` synthesize a message-less `{ success: true }` stub response — but the fix makes the *shared* helper permissive for **any** message-less object, app-wide, rather than fixing the two export callers directly. This removes a useful "loud but ugly" signal that previously caught backend responses missing `.text` (as findings #1 and #2 demonstrate), and no regression test was added to pin down the new `''` contract.

---
Also flagged, lower-confidence / worth a look but not blocking: a Hunt.vue floating-toast case where a failed `removeUser`/`addUsers` call could display a stale "success" toast instead of an error (shared `floatingError`/`floatingSuccess` state), the export success toast now showing nothing at all (seems intentional per the PR description), a few pre-existing hand-rolled stringification spots elsewhere that are now inconsistent with this fix, and the lack of unit tests for `resolveMessage()`'s changed fallback behavior.

PR #4225, opened 2026-08-31
Add sankey and connections dashboard widgets

awick:
Add sankey and connections dashboard widgets
- Sankey widget: flows between 1-3 nested fields via spigraphhierarchy
- Connections widget: display-only src -> dst force graph with hover details
- Edit modal caps connections at exactly a source + destination field pair
- api-summary.t now covers widgets[], noStats, per-widget expressions and the multi-metric metrics[]/sortMetric/metricValues path (87 -> 122 tests)
- multies: honor the requested terms aggregation order when merging clusters and sum all metric sub-aggs, not just db/pa
- Document metrics/sortMetric/metricValues in the /api/sessions/summary API doc
- Add sankey/connections keys to all locale files

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>
## License


31453:
## Review notes

Reviewed the full diff for correctness plus reuse/simplification/efficiency. 15 findings, 13 confirmed against the code, 2 marked plausible (reachability or judgement calls). **The UI/visual items below are being handled in a separate follow-up commit** — flagging them here so they're tracked, not as blockers on this one.

---

### `viewer/multies.js` — required, and fixes more than the CHANGELOG claims

Both multies changes are load-bearing for the dashboard, not incidental:

- **`aggAdd`**: the widget aggregation at `apiSessions.js:3604` names its sub-aggs `bytes`, `packets`, and `m0`/`m1`/… (the `aggKey` template at `apiSessions.js:3412`) — never `db`/`pa`. Without the change, `doc_count` merges correctly while bytes/packets/metric columns keep only whichever cluster reached the bucket first.
- **`_order`**: `apiSessions.js:3606` orders Top-N by `{ [sortAggKey]: order }`, and multies never truncates merged buckets — `convert()` does the `slice(0, length)` at `apiSessions.js:3429`. So re-sorting by `doc_count` didn't only reorder a widget, it changed *which rows survived the cut*.

Worth noting: `db` and `pa` don't exist anywhere in the tree any more — grep finds them only in the comment this PR adds. `git log -S` dates that branch to `4a8fbf58d` ("Switched from ES facets to aggregations"). The live consumer of that code path today is **`dbHisto`**, the Sessions page timeline graph, whose sub-aggs are `source.bytes` / `destination.bytes` / `source.packets` / `client.bytes`… (`buildQuery.js:432-446`), read back at `viewerUtils.js:166`. So on a multiviewer the timeline's **sessions** series has been correct while its **bytes/packets** series showed a single arbitrary cluster's contribution per bucket. This PR fixes that too — it deserves its own CHANGELOG line, and the current wording ("instead of just bytes/packets") is slightly off since bytes/packets weren't being summed either.

Two issues with the implementation:

**1. `multies.js:1043` — min/max sub-aggs are now summed** *(confirmed)*

The generic loop sums every sub-agg with a numeric `.value`. `apiStats.js:127` puts `first: { min: { field: 'first' } }` under a `terms` agg on `node`, which routes through the same generic `/:index/_search` handler. When two clusters both have a bucket for the same node name, `o1.first.value += o2.first.value` sums two epoch timestamps and `/api/stats` reports a retention date decades out. Previously `o1.db` was undefined so the value was left alone — wrong, but plausible-looking rather than absurd.

Recursing into `aggAdd` does **not** fix this: its min/max detection is name-based (`includes('min')`, `=== 'firstPacket'`) and this sub-agg is named `first`, so recursion sums it identically. The fix needs to come from the query shape — have `newResult` stash the operators the same way it now stashes `_order`:

```js
_subAggs: { bytes: 'sum', packets: 'sum', first: 'min' }
```

on the bucket container (which is already the object `aggAdd` holds as `obj1[aggname]`), have the loop consult it defaulting to `sum`, and delete it in `aggConvert2Arr` next to `_order`. One gotcha: `apiSessions` uses the `aggs` key while `apiStats` and `buildQuery` use `aggregations`, so it needs `aggConfig.aggs || aggConfig.aggregations`.

**2. `multies.js:935` — tiebreak still returns `NaN` for string keys** *(confirmed empirically)*

The rewritten comparator handles strings in the primary comparison via `localeCompare`, but the tiebreak stayed `a.key - b.key`. Running the exact comparator over 8 keyword buckets tied on `doc_count`: `'tcp' - 'udp'` is `NaN`, the comparator becomes inconsistent, and the sort degenerates to insertion order — feeding the same buckets in two different arrival orders produced two different results. Insertion order in multies is cluster response order, so for `protocols` / `ip.src` / `host.http` (where tail counts tie constantly) the merged ranking varies between identical requests, and combined with the `slice(0, length)` above it changes Top-N membership. Mirroring the `typeof === 'string'` branch in the tiebreak closes it.

---

### Behaviour

**3. `widgetData.js:184` — connections widget samples 20 sessions** *(confirmed)*

`length: widget.length || 100` never falls back to 100: `blankForm()` sets `length: 20` and `save()` always writes it. `/api/connections` uses `length` as the session sample size, and `lengthItems` only offers `[10, 20, 50, 100]` — where 100 is the *minimum* the SPI Graph connections page offers (`ConnectionsGraph.vue:31` goes to 100000). Default widgets build a graph from 20 sessions that reads as the whole picture. This probably wants its own sample-size scale rather than reusing the Top-N `length` control.

**4. `viewModes.js:41` — Order is offered for sankey/connections but ignored** *(confirmed)*

Adding both to `AGG_VIEW_MODES` enables the modal's Order select (`:disabled="!aggEnabled"`) and makes `widgetInfo()` render "Top N"/"Bottom N", but neither `fetchHierarchy` nor `fetchConnections` sends `order`, and `buildWidgetParams` only forwards view/bounding/interval/cluster. Picking "Bottom 20" returns the identical top-20 and the info tooltip then confirms "Bottom 20". The comment on this line already says order doesn't apply — splitting the flag into length-only vs length+order would make that true in the UI.

**5. `ConnectionsWidget.vue:130` — link hover reports the source node's total** *(confirmed)*

`.on('mouseover', (e, d) => onHover(e, d.source))` discards `d.value`, the pair's session count. A source with 500 sessions over 20 destinations shows "500 sessions" when hovering an edge worth 3 — the edge is the only place the pair-level number is exposed.

**6. `HierarchyWidget.vue:123` — single-field sankey shows an untranslated "Top Talkers" root** *(confirmed)*

The kept root is `apiSessions.js:3327`'s hard-coded `{ name: 'Top Talkers' }`, and it gets `fieldIdx: Math.max(0, -1) = 0`, so hovering it attributes the grand total to the first field. Visible in every non-English locale.

**7. `widgetData.js:183` — silent degradation below 2 fields** *(plausible)*

`dstField: toDb(fields[1] || fields[0])` sends src == dst for a one-field widget, and with no fields both resolve to `undefined`, which `JSON.stringify` drops so the server silently defaults to source.ip/destination.ip. Not reachable through the modal (`save()` enforces the pair) — only via an imported or hand-edited dashboard config, or a field exp that stops resolving after a rename. Cheap to close by rejecting a short field list in `fetchConnections` so `useSpigraphWidget` surfaces it on the card.

---

### Rendering — *deferred to the follow-up UI commit*

**8. `ConnectionsWidget.vue:115` — anisotropic rescale voids `forceCollide`** *(confirmed)*

x and y are normalized independently to fill the box, so the layout's aspect ratio is destroyed while radii and collide distances stay in pre-scale units. The grid is `repeat(4, minmax(0, 1fr))` with `grid-auto-rows: 160px`, so a 4x1 widget on a 1600px page is roughly 1580x120 of plot area — a ~13:1 box fed a roughly square layout. Nodes that `forceCollide` separated by `radius + 3` end up a couple of pixels apart. A single uniform scale = `min(sx, sy)` plus centering fixes it.

**9. `HierarchyWidget.vue:188` — sankey reserves 80px of right margin that labels never use** *(confirmed)*

`margin.right = 80` is reserved "for labels", but right-half nodes get `x = d.x0 - 4` with `text-anchor: end`, i.e. drawn *inside* the diagram over the incoming ribbons. The reserved strip is always blank, so the diagram is squeezed to ~80% width for no gain and long values still overlap the links. Either place last-column labels at `d.x1 + 4` with anchor `start`, or drop the margin to match the other sides.

---

### Efficiency / cleanup

**10. `ConnectionsWidget.vue:163`** *(confirmed)* — the ResizeObserver re-runs the full `sim.tick(200)` force simulation, but the result doesn't depend on widget size (coordinates are normalized afterwards). Window resize fires this continuously. Caching the settled positions and re-running only the coordinate mapping on resize would avoid it. (The drag handles snap to whole grid cells via `Math.round` in `startResize`, so those fire only a few times per drag.)

**11. `HierarchyWidget.vue:105`** *(confirmed)* — `hierarchyToSankey` is a near-verbatim copy of `formatDataFromSpigraphSankey` in `spigraph/Hierarchy.vue:649`: same cumulative helper, same `${name}_${depth}` ids, same multi-level/root-keep branch. Two copies that now have to stay in sync or the two sankeys disagree on the same data.

**12. `HierarchyWidget.vue:120`** *(confirmed)* — `cumulative(n)` walks the whole subtree and is called twice per node (node entry + link), and `traverse` re-descends into subtrees of nodes already in `nodeMap`. At the default size=20 a 3-field sankey is up to 20 + 400 + 8000 buckets, so this dominates the render. Memoizing in a `Map` keyed by node and reusing the value for the link is a two-line change.

**13. `ConnectionsWidget.vue:170`** *(confirmed)* — `watch(() => props.widget.viewMode, render)` can never fire; the component only mounts under `v-else-if="w.viewMode === 'connections'"`, so a mode change unmounts it. Carried over from `HierarchyWidget`, where it's load-bearing (pie/treemap/sankey share one component).

**14. `viewModes.js:56`** *(plausible, judgement call)* — `FIELD_PAIR_VIEW_MODES` + `requiresFieldPair` + `maxFields` + a bespoke `twoFieldsRequired` message is a parallel list layered on `MULTI_FIELD_VIEW_MODES`, whose limit of 3 stays hardcoded in two places. A single `FIELD_COUNT_LIMITS = { connections: [2, 2], pie: [1, 3], … }` with one parameterized message would replace the second list and both hardcoded 3s, and stop each future mode needing two registrations.

---

### Test coverage

**15. `tests/api-summary.t`** *(confirmed)* — the new `postSummary` single-vs-multi comparisons are good coverage for the summary path, including the metric-ordered merge. But the multies changes touch the merge path for *every* viewer bucket aggregation, and nothing exercises a bucket carrying a non-sum sub-agg. `api-stats.t:37` only counts rows (`is (@{$mstats->{data}}, 3)`) and never reads the retention value, so finding 1 ships green — as does the `dbHisto` timeline fix this PR makes. A single-vs-multi compare on `/api/stats` and on a `facets=1` sessions query would cover both.

31453 (approved):

PR #4226, opened 2026-08-31
Add Featherprint passive device fingerprinting

awick:
Viewer-resident subsystem that builds per-IP device records from session metadata (ARP, DHCP, DNS/mDNS, NBNS, SSDP) with no capture push or wise dependency.

- apiFeatherprint.js: polling engine (background per-subnet monitor with chunked backfill + persisted cursor, and on-demand lookup), longest-prefix subnet PolicyTree, per-IP signal extraction and rule-based classifier
- featherprint.db.js: ES persistence for device/history/alert/mac records
- featherprint.rules.yaml: MAC source ranking + device classification rules
- Featherprint.vue: tracked/lookup/alerts/admin page with session pivot menus
- db.pl: add featherprint, _history, _alerts and _macs indices (v89), plus dhcp.yiaddrIp/paramReqList and dns.answers ptr/srv/port/weight mappings
- viewer.js routes + init, apiMisc featherprintLookup value action, router and navbar wiring, [featherprint-subnet]/-defaults config, en + 9 locales
- Lookup applies the caller's forced expression and time limit so a restricted user cannot read session data outside their scope
- tests: api-featherprint.t, feather-gen.py, feather_synthetic pcap/baseline

## License


31453 (commented):
Review of the featherprint engine, persistence, packaging and wiring.

I validated each item before writing it up rather than reporting statically: most of the engine claims below were reproduced by loading the real `apiFeatherprint.js` / `featherprint.db.js` against a stub `Db` that records every call, so the numbers and event traces in the comments are actual output rather than estimates. The packaging one was reproduced by copying only what `viewer/Makefile.in` installs into a scratch dir and running from there.

**14 confirmed, 1 refuted.** The refuted one, for transparency: I expected the JSDoc block for `/featherprint/ip/:ip` to attach to `#isIpOrCidr`, since two private helpers sit between the comment and `apiGetIp`. Running `jsdoc2md` with the repo's own partials shows the `@name` tag overrides symbol inference and all nine endpoints render correctly — no change needed.

The three I'd weigh most heavily:

- **`apiFeatherprint.js:318`** — a DHCP OFFER/ACK binds the client's MAC and hostname onto the DHCP *server's* IP. The gateway ends up carrying a client's MAC, accumulates every client's hostname, and each lease emits a spurious `changeIp`.
- **`viewer/featherprint.rules.yaml`** — never installed by `make install`, so on a packaged deployment the classifier is empty: everything classifies `unknown` and `changeIp` never fires at all.
- **`viewer.js:1597`** — `/lookup` applies the caller's forced expression, as the PR description says, but `/search`, `/ip/:ip`, `/history/:ip` and `/alerts` don't, so the persisted store hands back the same session-derived data the expression was meant to fence off.

The rest are correctness, efficiency and robustness items inline. One UI item (`Featherprint.vue:646` — timestamps ignoring the user's timezone) is flagged for the followup commit rather than this one.

Genuinely nice feature, and several parts are carefully done — the longest-prefix policy tree with auto-unignore, the content-hashed history ids, and the cross-window MAC-rank guard all clearly anticipate real failure modes. Comments are non-blocking.

viewer/apiFeatherprint.js
```diff
@@ -0,0 +1,1599 @@
...
+
+    // DHCP: broadcast DISCOVER/REQUESTs come from 0.0.0.0 and OFFER/ACKs from
+    // the server, so bind by the server-assigned yiaddr or the requested IP;
+    // source.ip covers unicast renewals from the leased address. The `dhcp`
+    // policy flag (default true) can turn this off per subnet.
+    if (src?.dhcp && flags?.dhcp !== false) {
+      const dhcpBound =
+        toArray(src.dhcp.yiaddrIp).includes(ip) ||
+        toArray(src.dhcp.requestIp).includes(ip) ||
+        src?.source?.ip === ip;
```
31453: **DHCP OFFER/ACK binds the client's identity onto the server's IP.**

`dhcpBound` ORs in `src.source.ip === ip`, so an ACK *sent by* the DHCP server matches the server's own IP and absorbs the client's `dhcp.mac`, `dhcp.host`, `classId` and `paramReqList`. The comment above says this disjunct covers unicast renewals from the leased address, but nothing restricts it to client-originated messages.

Drove the real engine with one textbook ACK (`source.ip=10.177.0.1`, `yiaddrIp=10.177.0.14`, `dhcp.mac=aa:bb:cc:00:00:04`, `dhcp.host=win11-rig`):

```
ip=10.177.0.1   mac=aa:bb:cc:00:00:04/dhcp  names=["win11-rig/dhcp"]  class=windows_pc
ip=10.177.0.14  mac=aa:bb:cc:00:00:04/dhcp  names=["win11-rig/dhcp"]  class=windows_pc
```

The gateway becomes a clone of the client. Since `dhcp` outranks `srcMac` it also feeds `#trackMacForIp`. Two leases across two windows:

```
w1: 10.177.0.1  newIp, newMac
    10.177.0.14 newIp, newMac, changeIp 10.177.0.1->10.177.0.14
w2: 10.177.0.1  changeMac aa:bb:cc:00:00:04->de:ad:be:ef:00:15, newName, changeName
    10.177.0.15 newIp, newMac, changeIp 10.177.0.1->10.177.0.15
gateway record: mac=de:ad:be:ef:00:15  names=["bobs-laptop","win11-rig"]
```

6 alerts for two ordinary leases, and every client gets a bogus `changeIp` off the gateway. Dropping the `source.ip` disjunct when `yiaddrIp`/`requestIp` are present and don't contain `ip` would keep renewals working.

viewer/featherprint.rules.yaml
```diff
@@ -0,0 +1,179 @@
+# featherprint.rules.yaml -- featherprint classifier rules.
```
31453, line 1: **This file never ships.** `viewer/Makefile.in:25` installs only `*.js package.json package-lock.json` into `$(VIEWERDIR)` (plus `views`/`public`/`vueapp/dist` via `$(CP) -pr`), so the yaml is left behind. On a packaged install none of `findRulesFile()`'s candidates exist and `loadRules()` returns `{macSourceRank:{}, classify:[]}`.

Reproduced by copying just the install globs into a scratch dir and running the engine from there:

```
FEATHERPRINT - no rules file found; classifier disabled
  chromecast mDNS      -> classification = unknown
  same MAC on a new IP -> history kinds = ["newIp","newMac","newIp","newMac"]
```

The same input from the source tree emits `changeIp`. With an empty `macSourceRank`, `#trackMacForIp`'s guard `(rank[source] ?? 0) <= (rank.srcMac ?? 0)` is `0 <= 0` for every MAC, so the arpwatch-style `changeIp` detection is entirely dead in production and every device classifies `unknown`. Tests don't catch it because they run from `viewer/`.

Adding the yaml to the `$(INSTALL)` line, or to the `$(CP) -pr` alongside `views public`, fixes it.

viewer/viewer.js
```diff
@@ -1574,6 +1586,61 @@ app.post( // update cron endpoint
...
 
+// featherprint apis ----------------------------------------------------------
+app.get( // featherprint device record endpoint
+  ['/api/featherprint/ip/:ip'],
+  [ArkimeUtil.noCacheJson],
+  FeatherprintAPIs.apiGetIp
+);
+
+app.get( // featherprint device search endpoint
+  ['/api/featherprint/search'],
```
31453, line 1609: **The persisted read APIs don't apply the caller's forced expression.**

`/lookup` correctly ANDs the caller's admin-set expression into the session query via `#applyUserExpression`, and the PR description calls that out as deliberate. But `/search`, `/ip/:ip`, `/history/:ip` and `/alerts` never touch `req.user` — across those five handlers, `user:` appears only in `apiLookup`. They're gated by viewer's blanket `arkimeUser` check and nothing else.

So a user restricted to `ip == 10.1.0.0/16` is correctly refused `/lookup?ip=10.9.0.5`, but `GET /api/featherprint/search?limit=10000` returns every monitored device — IPs, MACs, MAC history, DHCP hostnames, mDNS/NBNS/SSDP names and services — all derived from the session data that expression was meant to fence off.

Worth deciding explicitly whether the persisted store is in scope for `user.expression`. Right now the two halves of the feature disagree.

viewer/apiFeatherprint.js
```diff
@@ -0,0 +1,1599 @@
...
+    for (const f of [...FeatherprintAPIs.ALERT_FLAGS, 'ignore']) {
+      if (defaults[f] === undefined) defaults[f] = false;
+    }
+    // DHCP signal collection is on unless explicitly disabled.
+    if (defaults.dhcp === undefined) defaults.dhcp = true;
+
+    const subnetsSection = ArkimeConfig.getSection('featherprint-subnet') ?? {};
+    const subnets = [];
+    for (const cidrText of Object.keys(subnetsSection)) {
+      const cidr = FeatherprintAPIs.parseCidr(cidrText);
```
31453, line 718: **A typo'd CIDR stops the viewer from starting.**

`parseCidr` validates the prefix length but never the address, so `10.0.0/8` returns `{ip:'10.0.0', bits:8}` instead of `null` and slips past the `WARNING - invalid CIDR` skip right below. `iptrie.add()` then throws:

```
parseCidr("10.0.0/8") = {"text":"10.0.0/8","ip":"10.0.0","bits":8,"family":4}

loadConfig "10.0.0/8"        -> THREW TypeError: IPTrie
loadConfig "10.0.0.256/24"   -> THREW TypeError: IPTrie
loadConfig "192.168.1.0/241" -> OK (warned + skipped)
```

`initialize()` calls `loadConfig()` synchronously and `premain()` has no `.catch()`. viewer.js's `process.on('unhandledRejection')` only `console.trace`s, so the process doesn't exit — it just never reaches `main()` and never binds the port. A silent hang rather than a clear config error.

A `net.isIP(ip) === 0` check inside `parseCidr` would route these to the warn-and-skip path that's already there.

viewer/apiFeatherprint.js
```diff
@@ -0,0 +1,1599 @@
...
+    const includeCidrs = this.subnets
+      .filter(s => !s.resolvedFlags.ignore)
+      .map(s => s.cidr.text);
+
+    if (includeCidrs.length === 0) return null;
+
+    const should = [];
+    for (const cidr of includeCidrs) {
+      for (const field of ipFields) {
+        should.push({ term: { [field]: cidr } });
```
31453: **The subnet prefilter grows at 6 x subnets.**

One `term` per (IP field, subnet) in a single `bool.should`. Measured against the real `#buildEsQuery`:

```
   1 subnets ->    6 should clauses
  10 subnets ->   60
 200 subnets -> 1200
 700 subnets -> 4200
```

Around 170 subnets this passes ES 7's default `indices.query.bool.max_clause_count` of 1024, and ~700 passes the 4096 default on ES 8 / OpenSearch 2. When it trips, `tick()` logs `tick chunk error`, breaks the chunk loop, and the cursor stops advancing — the monitor stalls silently.

Collapsing to one clause per field rather than one per (field, subnet) keeps this at 6 regardless of subnet count.

viewer/featherprint.db.js
```diff
@@ -0,0 +1,146 @@
...
+
+  // --------------------------------------------------------------------------
+  // Get a single device record by IP. Returns the doc (with _id) or null.
+  static async getIp (ip) {
+    return FeatherprintDb.#findOne(FeatherprintDb.INDEX, { ip }, true);
+  }
+
+  // Upsert a device record. Used by monitor mode.
+  static async upsertDevice (doc) {
+    await Db.indexNow(FeatherprintDb.INDEX, doc.ip, doc);
```
31453: **Every device/mac write forces an index refresh, and every read is a search.**

All four doc types here are addressed by a known `_id`, but reads go through `#findOne` -> `Db.search`, which is what forces the writes to use `Db.indexNow` (`refresh: true`) for read-your-writes. Measured on one chunk touching 100 IPs:

```
first pass:   {searchSessions:1, search:200, indexNow:400, index:200}  -> 8.0 ops/IP, 4 forced refreshes/IP
steady state: {searchSessions:1, search:200, indexNow:200}             -> 4.0 ops/IP, 2 forced refreshes/IP
```

Two forced refreshes per IP per chunk even when nothing changed. A chunk over a /16 is thousands of refreshes on `featherprint`/`featherprint_macs`.

`Db.get(index, id)` (db.js:337) is a realtime GET that needs no refresh — moving the `#findOne` callers to it would let the writes use plain `Db.index` and roughly halve the round trips. It throws on 404, so callers would need a catch.

viewer/apiFeatherprint.js
```diff
@@ -0,0 +1,1599 @@
...
+      mac: bestMac
+        ? {
+          value: bestMac.value,
+          source: bestMac.source,
+          firstSeen: macHistory[0]?.ts ?? bestMac.ts,
+          lastSeen: bestMac.ts,
+          history: macHistory
+        }
+        : (previous?.mac ?? null),
+      names: FeatherprintClassify.#dedupNames([
```
31453, line 538: **`names` and `services` grow without bound.**

`ssdp.*` is capped via `slice(-20)`, but the merged `names`/`services` (and `mac.history`) have no cap and no ageing. Every SSDP `USN` and `SERVER` header is pushed as a *name*, and one UPnP device advertises a distinct USN per NT.

Simulated a single device sending 60 distinct NT/USN announcements:

```
names=61  services=60  ssdp.usn=20 (capped)  alerts=119
```

Each new entry also writes a history doc and, under `alertAll`, an alert. Same shape for a monitored resolver or CDN-facing host, where every DNS A answer naming that IP appends a name forever. The whole doc is re-serialized and re-indexed on every tick.

viewer/apiFeatherprint.js
```diff
@@ -0,0 +1,1599 @@
...
+  static async #scrollSessions (query, limit) {
+    // Db.searchScroll (under searchSessions) does a plain search up to the ES
+    // 10k window and scrolls beyond it, so pass the real limit through.
+    query.size = limit;
+    // @timestamp is session start, but sessions rotate on lastPacket, so a
+    // session falling in our window may live in a later/earlier index. Match
+    // cron: search across all session indices each time.
+    const r = await Db.searchSessions(Db.getSessionIndices(true), query, {});
+    const hits = r?.hits?.hits ?? [];
+    if (hits.length >= limit) {
```
31453, line 1212: **Hitting `featherprintLookupLimit` warns but still marks the chunk done.**

`#scrollSessions` logs and returns the truncated set; `aggregate` resolves normally with no truncation signal, and `tick()` advances `state.lpValue = chunkEnd` regardless. Ran a tick with `featherprintLookupLimit=5` where every chunk hit the limit:

```
WARNING - featherprint: window scan hit featherprintLookupLimit (5); some sessions were not examined
persisted cursor lpValue=2026-09-01T19:23:19Z   (advanced to ~now)
```

The unexamined sessions are never revisited — no retry with a narrower window, nothing persisted to mark the gap. On a busy segment, device records silently miss MACs, hostnames and services with only a log line as evidence. Halving the chunk and re-running on truncation instead of advancing would make it recoverable.

viewer/apiFeatherprint.js
```diff
@@ -0,0 +1,1599 @@
...
+        ip,
+        kind: ev.kind,
+        ts: Date.now(),
+        before: ev.before,
+        after: ev.after,
+        acked: false,
+        message: `featherprint: ${ip} ${ev.kind}`
+      };
+      try {
+        await FeatherprintDb.insertAlert(alertDoc);
```
31453: **Alerts aren't deduplicated on replay, but history is.**

`appendHistory` builds a deterministic id from `{ip, kind, ts, sha1(before/after)}` so re-processing collapses onto one doc. `insertAlert` passes `undefined` as the id, so every replay inserts a fresh alert. Replaying one window:

```
alert docs 2 -> 4   (distinct history ids 2 -> 2)
```

`tick()` advances `state.lpValue` in memory and *then* calls `saveState`, whose failure is logged and swallowed — so a failed state write, or a restart before it lands, replays the window and re-fires every alert. Same on a cron-leader change that re-reads a stale cursor. Operators end up acking the same event twice.

viewer/apiFeatherprint.js
```diff
@@ -0,0 +1,1599 @@
...
+    const prevFirst = previous?.firstSeen;
+    const record = {
+      ip,
+      firstSeen: prevFirst ? Math.min(prevFirst, firstSeen) : firstSeen,
+      lastSeen,
+      mac: bestMac
+        ? {
+          value: bestMac.value,
+          source: bestMac.source,
+          firstSeen: macHistory[0]?.ts ?? bestMac.ts,
```
31453: **`mac.firstSeen` is the first MAC ever seen, not the current one's first sighting.**

`macHistory[0]` is the oldest entry in the MAC *history*, which is a different MAC once the binding has changed:

```
current mac=bb:bb:bb:00:00:02  firstSeen=1000  lastSeen=9000
history=[{mac:aa:aa:aa:00:00:01, ts:1000}, {mac:bb:bb:bb:00:00:02, ts:9000}]
```

It reports MAC B as first seen at t=1000, before it existed on the network — wrong for every device that has ever changed MAC. The matching `macHistory` entry, or `bestMac.ts` when the MAC is new, is the right source.

viewer/apiFeatherprint.js
```diff
@@ -0,0 +1,1599 @@
...
+        state.lastTickAt = Date.now();
+        state.lastChunkSessions = result.sessionsScanned;
+        state.lastChunkAlerts = result.alerts;
+        try {
+          await FeatherprintDb.saveState(state);
+        } catch (e) {
+          console.log('featherprint: saveState failed:', e?.message || e);
+        }
+
+        if (result.sessionsScanned === 0) {
```
31453, line 924: **`featherprintMaxChunksPerTick` doesn't bound empty chunks, so one tick can issue ~10k queries.**

`remaining` only decrements on non-empty chunks, so the fast-forward is unbounded — the chunk *size* caps at 24h but the *count* doesn't. Ran a real cold-start tick with `initialLookbackDays=-1` and a `min(@timestamp)` of 2000-01-01, every chunk empty:

```
session queries issued : 9749
```

The comment on the `initialLookbackDays` default says "-1 is cheap even across years of mostly-empty range" — ~10k cross-index searches over `sessions3-*` in a single tick isn't.

This is also the regression-suite default: `tests/config.test.ini:62` sets `featherprintInitialLookbackDays=-1` and `:459` scopes `[featherprint-subnet]` to `0.0.0.0/0` + `::/0`. The oldest session in the test corpus is 1999-03-11 (27.5 years, ~10,035 chunks at the 24h cap), so every `/regressionTests/processFeatherprints` drain pays that, twice per run — plus the background monitor doing the same on the cron leader for every viewer test run, writing device and alert docs for every IP in every unrelated test pcap.

`api-featherprint.t` notes all its devices live in 10.177.0.0/16; scoping the test subnet there would keep the coverage without the corpus-wide scan.

viewer/apiFeatherprint.js
```diff
@@ -0,0 +1,1599 @@
...
+    return n;
+  }
+
+  static loadConfig () {
+    const intervalSec = FeatherprintAPIs.#posInt('featherprintInterval', 300, 30);
+    const settings = {
+      monitorEnabled: Config.get('featherprintMonitorEnabled', true),
+      intervalSec,
+      notifier: Config.get('featherprintNotifier', undefined),
+      historyLimit: FeatherprintAPIs.#posInt('featherprintHistoryLimit', 50),
```
31453, line 684: **`featherprintHistoryLimit` and `featherprintDedupWindow` are parsed but never read.**

Zero uses of either anywhere in the module — and both are surfaced to operators through `/api/featherprint/config` and the admin "Settings:" pane. Someone setting `featherprintDedupWindow=3600` expecting alert de-duplication gets nothing (`#fireNotifiers` has no dedup at all), and `featherprintHistoryLimit` bounds neither `mac.history`, nor `ipHistory` (hardcoded 50), nor `listHistory` (which takes its own `limit`).

Either wire them up or drop both lines so `/config` stops advertising knobs that do nothing.

viewer/apiFeatherprint.js
```diff
@@ -0,0 +1,1599 @@
...
+  RULES = loadRules();
+  for (const r of RULES.classify) {
+    compileClause(r);
+    if (Array.isArray(r.anyOf)) r.anyOf.forEach(compileClause);
+  }
+}
+
+function clauseMatches (clause, ctx) {
+  if (clause.hasService !== undefined && !ctx.serviceTypes.has(clause.hasService)) return false;
+  if (clause.nameContains !== undefined && !ctx.nameStrs.some(n => n.includes(clause.nameContains))) return false;
```
31453: **`nameContains` is documented case-insensitive but isn't.**

`ctx.nameStrs` is lowercased; `clause.nameContains` isn't. The rules file documents it as `nameContains: <substr> -- any discovered name contains substr (case-insensitive)`.

With a rules file listing the uppercase form first:

```yaml
- class: apple_tv_upper
  nameContains: AppleTV
- class: apple_tv_lower
  nameContains: appletv
```

a device named `AppleTV.local` classifies as `apple_tv_lower` — the first rule never matches. Invisible with the shipped all-lowercase rules, but the file is explicitly user-overridable. `n.includes(clause.nameContains.toLowerCase())` fixes it.

viewer/vueapp/src/components/featherprint/Featherprint.vue
```diff
@@ -0,0 +1,866 @@
...
+    stopTimer () {
+      if (this.timer) { clearInterval(this.timer); this.timer = null; }
+    },
+    onRowClick (_event, row) {
+      const ip = row?.item?.ip ?? row?.item?.raw?.ip;
+      if (ip) this.selectIp(ip);
+    },
+    fmtTs (ts) {
+      if (!ts) return '—';
+      try { return new Date(ts).toISOString().replace('T', ' ').slice(0, 19); }
```
31453: *(UI — flagging for the followup commit, not this one.)*

`fmtTs` hand-rolls `new Date(ts).toISOString()`, which always renders UTC and ignores the user's timezone setting. Every other viewer page uses `timezoneDateString` from `common/vueapp/vueFilters.js` (Files, History, HuntData, HuntRow, SessionDetail, SessionField, StickySessions, ConnectionsGraph).

So a user on `local`/`localtz` sees Sessions in their own timezone while every Featherprint timestamp — first/last seen, alert time, history entries, monitor cursor — is UTC with no marker, and comparing a `lastSeen` against a session start silently mis-reads by the offset.

viewer/apiFeatherprint.js
awick: Fixed in 384a29a2f — `source.ip` is now only a fallback when the session carries no `dhcp.yiaddrIp`/`dhcp.requestIp`, so an ACK sent by the server no longer binds the client's mac/host/classId onto the server's own IP.

viewer/featherprint.rules.yaml
awick: Fixed in 384a29a2f — `featherprint.rules.yaml` added to the `$(INSTALL)` line in `viewer/Makefile.in`, so it ships with a packaged install.

viewer/apiFeatherprint.js
awick: Fixed in 384a29a2f — `parseCidr` now validates the address with `net.isIP`, so a typo'd subnet takes the existing warn-and-skip path instead of throwing inside `iptrie.add()`.

viewer/apiFeatherprint.js
awick: Fixed in 384a29a2f — the prefilter now emits one `terms` per field instead of one `term` per (field, subnet), so it stays at 6 clauses regardless of subnet count. This needed CIDR values to work in `terms`, which I added in arkimedb v0.1.26.

viewer/featherprint.db.js
awick: Fixed in 384a29a2f — the id-keyed docs (state/device/mac) are read with a realtime `Db.get` and their writes dropped the forced refresh, removing the two forced refreshes per IP per chunk.

viewer/apiFeatherprint.js
awick: Fixed in 384a29a2f — `names` and `services` are capped, and `mac.history`/`ipHistory` are now bounded by `featherprintHistoryLimit`.

viewer/apiFeatherprint.js
awick: Fixed in 384a29a2f — `aggregate` now returns a `truncated` flag and the tick halves the window and retries instead of advancing the cursor past unexamined sessions.

viewer/apiFeatherprint.js
awick: Fixed in 384a29a2f — alerts now use a deterministic id (`ip|kind|ts|content-hash`) with a create-only insert, so a replayed window neither duplicates an alert nor resurrects an acked one. Alert `ts` is the event time rather than `Date.now()`.

viewer/apiFeatherprint.js
awick: Fixed in 384a29a2f — `mac.firstSeen` now uses the `macHistory` entry matching the current MAC (falling back to `bestMac.ts`) instead of `macHistory[0]`.

viewer/apiFeatherprint.js
awick: Fixed in 384a29a2f — empty chunks now fast-forward logarithmically (the 24h cap was what forced the linear walk) with a hard per-tick iteration cap, and a truncated chunk halves and retries. The test subnet is also scoped to 10.177.0.0/16 so the monitor stops building docs for every IP in the corpus.

viewer/apiFeatherprint.js
awick: Fixed in 384a29a2f — `featherprintHistoryLimit` now bounds `mac.history` and `ipHistory`, and the unused `featherprintDedupWindow` was removed so `/config` stops advertising it.

PR #4227, opened 2026-08-31
Add opt in update available check to the version menu

awick:
- New shared UpdateCheckService fetches <updateCheckUrl>/releases-v<major>.json, compares versions locally and caches the result for a day
- Version menu gains a consent prompt, a check button, the result and a dismissable dot, so viewer, cont3xt, parliament and wise all get it at once
- New checkForUpdates (off, manual, auto) and updateCheckUrl settings, resolved by ArkimeUtil.updateCheckConfig which fails closed on a bad value
- CSP connect-src is only widened when an update origin is configured, so off is enforced by the browser rather than just by the client code
- Uses plain fetch with credentials omitted, fetchWrapper would have attached Arkime session cookies to an off origin request
- Add updateCheck translations to all 10 locales

## License


31453:
Review pass over `5e306fa`. This one is scoped to correctness, config/CSP wiring and structure — **a UI/UX-focused pass will follow in a separate commit**, so styling, layout and the interaction design of the menu itself are deliberately not covered here.

Everything below was verified by running the code rather than by reading alone: `updateCheckConfig` + the helmet directive were exercised through the real middleware, and `UpdateCheckService.js` was driven against a stubbed `localStorage`/`fetch`.

---

## Blocking

### 1. The feature is inert in parliament and wise

`parliament.js` / `wiseService.js` send the locals, both `index.html.ejs` declare the globals, and both widen CSP — but neither `main.js` copies them into `constants`:

```js
// parliament/vueapp/src/main.js:82  (wiseService/vueapp/src/main.js:81 is the same)
const constants = { PATH, VERSION, LOGOUT_URL, LOGOUT_URL_METHOD, BUILD_DATE, BUILD_VERSION, FOOTER_CONFIG };
initUpdateCheck(constants);   // CHECK_FOR_UPDATES / UPDATE_CHECK_URL are missing
```

`initUpdateCheck` sees `undefined`, falls to `state.mode = 'off'` (`UpdateCheckService.js:189`) and `Version.vue` renders no update UI at all. viewer and cont3xt do add the two keys — it shows up in the diffstat too (`viewer 7 +-`, `cont3xt 10 +-` vs `parliament 3 +`, `wise 3 +`). Net effect: those two apps pay the CSP widening but can never use it.

### 2. "Not now" is a one-way door into a dead panel

After `denyConsent()`, `needsConsent` is `false` forever, so the consent prompt never returns. In the `v-else` branch every condition is false (`status === 'idle'`, no `latest`, not `'error'`, not `'done'`), so the panel renders nothing but the **Check for updates** button — and that button can never do anything, because `checkForUpdates()` returns on the `!state.consent` guard at `UpdateCheckService.js:128`.

```
after deny + click "Check for updates":
  status=idle latest=undefined consent=false  ->  fetch calls = 0
```

Short of clearing `localStorage` there's no way back. Either `check()` should re-grant consent, or the button should be replaced by the consent prompt when `consent === false`.

---

## Behaviour gaps

### 3. `manual` (the default) never reads the 24h cache, so the dot dies on every page load

`initUpdateCheck` only calls `checkForUpdates()` when `mode === 'auto'` (`:207`), and `Version.vue`'s `check()` always passes `force: true` — so in `manual` the cache written at `:153` is never read:

```
manual: after consent  -> latest=7.1.0  dot=true
manual: after reload   -> latest=undefined  dot=false     <-- cache ignored
auto:   after reload   -> latest=7.1.0  dot=true
```

So by default the notice only exists in the session where the user clicked, which undercuts the point of the dot.

### 4. 404s and network failures are never cached, so `auto` re-fetches on every load

Both the 404 early-return (`:149`) and the `catch` (`:155`) skip `writeStorage`:

```
404 host:  5 page loads -> 5 network requests, cache written? false
down host: 5 page loads -> 5 network requests, cache written? false   (each waits the 10s AbortSignal.timeout)
200 host:  5 page loads -> 1 network request,  cache written? true
```

A negative result wants a short-TTL entry too.

### 5. The cache key ignores `baseUrl`

`readCache` (`:113-120`) checks only `major` and `ts`, so changing `updateCheckUrl` keeps serving the previous host's payload for up to 24h:

```
consent against versions.arkime.com, then switch updateCheckUrl to https://mirror.internal
 -> requests to the new host = 0, latest still 7.1.0 (from the old host)
```

Worth recording `state.baseUrl` alongside `major`.

### 6. `security` can never be true on the `payload.latest` fallback

`security` is only assigned inside the `releases[]` loop, so a feed that publishes just the precomputed `latest` — the exact case the comment at `:100` says the fallback exists for — silently loses the flag:

```
payload = { latest: '7.0.2', url: '...', security: true }
 -> state.latest=7.0.2  state.security=false   (blue dot, no "Includes a security fix")
```

The fallback needs a `payload.security` read.

---

## Hardening

### 7. `release.url` reaches `:href` unvalidated

`applyPayload` copies `release.url` straight into `state.latestUrl` (`:96`) and `Version.vue:57` binds it to `:href`; Vue 3 does not sanitize href bindings.

```
payload { version: '7.1.0', url: "javascript:alert(document.cookie)" }
 -> state.latestUrl === "javascript:alert(document.cookie)"
```

Real-world exploitability is low — `script-src` has no `'unsafe-inline'`, so CSP blocks `javascript:` navigation in production, and Chrome blocks it under `target="_blank"` regardless. But CSP is a no-op under `NODE_ENV=development` (`viewer.js:110`), and `updateCheckConfig` deliberately permits `http:` origins (`arkimeUtil.js:687`), so a MITM on a plain-http mirror controls this value. A scheme check in `applyPayload` is cheap.

### 8. `mode: 'cors'` makes internal mirrors need CORS headers, undocumented

The CHANGELOG advertises pointing `updateCheckUrl` at an internal mirror, but `:141` uses `mode: 'cors'`, so the mirror must return `Access-Control-Allow-Origin` for the viewer's origin. Dropping `releases-v7.json` onto a stock nginx fails with nothing but a generic "Update check failed". Worth documenting with the setting, and ideally distinguishing a transport failure from a bad payload.

---

## Structure

### 9. `getUpdateCheck()` is pasted into four files; `ArkimeConfig.loaded()` already exists

The same `let updateCheck; function getUpdateCheck () { updateCheck ??= ... }` block plus its comment appears at `viewer.js:93`, `cont3xt.js:68`, `parliament.js:151`, `wiseService.js:185`. The repo already has the config-ready hook for this — `viewer.js:129` and `:411` use `ArkimeConfig.loaded()`. Resolving there would also move the `WARNING - unknown checkForUpdates value` (`arkimeUtil.js:681`) to startup, where an admin will actually see it, instead of firing lazily on whichever request happens to arrive first.

### 10. `connect-src 'self' 'self'` when the feature is off

`content-security-policy-builder` does a plain `value.join(' ')` with no dedupe, so `connectSrc: ["'self'", () => ...origin ?? "'self'"]` emits the token twice on every response:

```
{}                          -> connect-src 'self' https://versions.arkime.com
{checkForUpdates:'off'}     -> connect-src 'self' 'self'
{checkForUpdates:'bogus'}   -> connect-src 'self' 'self'    (+ warning, fails closed — good)
```

Since it can only change on restart, building the array once (`origin ? ["'self'", origin] : ["'self'"]`) inside that same `loaded()` hook drops both the duplicate and the per-request closure in all four files.

---

## Coverage

### 11. No tests, and the client path is disabled in every dev build

No test files in the commit. `updateCheckConfig` is a pure `(get) => {mode,url,origin}` and the CSP header is directly assertable from `tests/api-*.t`, so the server half is cheap to cover. The client half can't be covered as-is: `configure.ac` pins the dev version to `7.0.0-GIT` and `isDevBuild` forces `off`, so nothing in `make check` or normal development ever exercises it.

```
VERSION=7.0.0-GIT -> state.mode=off
VERSION=7.0.0     -> state.mode=auto
```

Suggested minimum: unit tests for `updateCheckConfig` (off / manual / auto / typo / bad URL / http / trailing slash) and a viewer test asserting `connect-src` is and isn't widened.

---

## Design question

### 12. `manual` default widens CSP for every install on upgrade

`get('checkForUpdates', 'manual')` means an install that changes nothing gets `https://versions.arkime.com` in `connect-src` after upgrading — including for users who clicked "Not now", and in parliament/wise where (per #1) the client can't use it. Verified with no config at all:

```
{} -> connect-src 'self' https://versions.arkime.com
```

The PR's framing is that "off is enforced by the browser rather than just by the client code"; with this default, that guarantee only reaches sites that discover the setting and opt out. Defaulting to `off` would ship the browser-level enforcement instead. Entirely your call — flagging it because Arkime lands in a lot of restricted networks.

---

## Nits

- `Version.vue:22` — the dot is a bare `<span>` carrying only `aria-label`. A `span` maps to role `generic`, which prohibits an author-supplied name, so screen readers get nothing. `role="img"` (or `role="status"` on a wrapper, so its appearance is announced) would make the label reachable.
- `cont3xt.js:19`, `parliament.js:40`, `wiseService.js:19` — a stray blank line splits each file's `require` block, adding three hunks of pure noise and moving `git blame` on those imports onto this commit.

---

## Checked and clean

`ArkimeConfig` **is** imported in `wiseService.js`; `Config.get` / `ArkimeConfig.get` are safe to pass unbound (neither touches `this`); EJS `<%= %>` closes the JS string-breakout path since `'` becomes `&#39;`; `updateCheckConfig` fails closed on every malformed input I threw at it, including `checkForUpdates=true`; all 10 locale files have identical key sets (the one-line `x-pl` offset predates this PR); each app renders `index.html.ejs` exactly once, so no path is missing the new locals; `@common` resolves in all four vite configs; and `npx eslint` is clean on every changed file.

PR #4230, opened 2026-09-01
Add mcpMaxTagSessions to bound MCP tag changes

31453:
`arkime_add_tags` with no `ids` fell through to the api's own 1000000 default, so one tool call could tag a million sessions — shipped config, no operator error, arguments chosen by a model. Adding tags needs no permission while removing them needs `removeEnabled`, so the caller cannot always undo it.

- Add `mcpMaxTagSessions`, default 10000, `-1` for no limit.
- Test `ids` the way the api does. `addTags`/`removeTags` pick their by-query branch with `if (req.body.ids)`, so a falsy but present `ids` such as `ids:""` skipped the cap and restored the 1000000 default. Nothing validates tool arguments against `inputSchema`, so it was reachable from a tool call.
- `addtags`/`removetags` now return `count`, and the tools report a truncated change instead of plain success — previously a removal could leave 40000 sessions tagged while the assistant reported them cleared.

Config hardening originally in this PR moved to #4233.

Test plan 111, six new assertions. `npm run lint` clean. Not run against a live cluster.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -50,6 +50,8 @@ NOTICE: Create a parliament config file before upgrading (see https://arkime.com
 ## Viewer
   - #4216 Fix multi viewer hanging or failing queries when one cluster is down or unhealthy (thanks @Juwon1405)
   - #4222 Fix session PCAP/CSV export success toast showing "[object Object]"
+  - #4230 Add mcpMaxTagSessions capping how many sessions one MCP tag change may touch, default 10000, -1 for no limit
+  - #4230 The addtags and removetags apis now return count, how many sessions were changed, and the MCP tools report when a change was truncated by mcpMaxTagSessions
 
 6.7.0 2026/08/19
 ## Release
--- a/release/config.ini.sample
+++ b/release/config.ini.sample
@@ -195,11 +195,15 @@ dropGroup=daemon
 # mcpMaxQueryDays caps how much time a query may span, in days, floating point,
 # so 0.5 is 12 hours. Defaults to 7. Use -1 for no limit, which lets an
 # assistant ask for all data and can be expensive on a large cluster.
+# mcpMaxTagSessions caps how many sessions one tag change may touch. Defaults
+# to 10000. Use -1 for no limit. When a change hits the cap the tool says so
+# rather than reporting that the whole expression was applied.
 # See https://arkime.com/settings#mcp for the auth modes, the authJwt* settings
 # and full Okta/Athenz/apache examples.
 #mcpEnabled=true
 #mcpAuthMode=header
 #mcpMaxQueryDays=7
+#mcpMaxTagSessions=10000
 
 # Should we parse HTTP QS Values
 parseQSValue=false
--- a/tests/api-mcp.t
+++ b/tests/api-mcp.t
@@ -1,6 +1,7 @@
-use Test::More tests => 90;
+use Test::More tests => 111;
 use ArkimeTest;
 use JSON;
+use URI::Escape;
 use Test::Differences;
 use Data::Dumper;
 use strict;
@@ -248,6 +249,52 @@ $json = callTool("arkime_create_hunt", '{"name":"mcphunt","search":"zzz","search
 is($json->{result}->{isError}, JSON::true, "arkime_create_hunt requires a time window");
 
 viewerGet("/regressionTests/deleteAllViews");
+
+################################################################################
+# mcpMaxTagSessions - the test viewer sets it to 2, so one tag change can only
+# ever touch 2 sessions however many the expression matches
+################################################################################
+my $pwd = "*/pcap";
+
+# more ids than the cap is refused up front, before anything is tagged
+$json = callTool("arkime_add_tags", '{"date":-1,"tags":"MCPCAPTEST","ids":"a,b,c"}');
+is($json->{result}->{isError}, JSON::true, "arkime_add_tags with more ids than mcpMaxTagSessions is a tool error");
+like($json->{result}->{content}->[0]->{text}, qr/mcpMaxTagSessions is 2/, "and the error names the setting and the limit");
+
+# 3 sessions match the expression, only 2 of them may be tagged
+countTest(0, "date=-1&expression=" . uri_escape("tags==MCPCAPTEST"));
+$json = callTool("arkime_add_tags", "{\"date\":-1,\"expression\":\"file=$pwd/socks-http-example.pcap\",\"tags\":\"MCPCAPTEST\"}");
+is($json->{result}->{isError}, JSON::false, "arkime_add_tags by expression succeeds");
+esGet("/_flush");
+esGet("/_refresh");
+countTest(3, "date=-1&expression=" . uri_escape("file=$pwd/socks-http-example.pcap"));
+countTest(2, "date=-1&expression=" . uri_escape("tags==MCPCAPTEST"));
+
+# a capped change must say it truncated, not report plain success
+is($json->{result}->{structuredContent}->{truncated}, JSON::true, "a capped tag change is reported as truncated");
+is($json->{result}->{structuredContent}->{count}, 2, "and reports how many sessions it touched");
+like($json->{result}->{structuredContent}->{text}, qr/mcpMaxTagSessions is 2/, "and the text names the setting");
+
+# clean up, the tagged sessions are exactly at the cap so one call clears them
+$json = callTool("arkime_remove_tags", '{"date":-1,"expression":"tags==MCPCAPTEST","tags":"MCPCAPTEST"}');
+is($json->{result}->{isError}, JSON::false, "arkime_remove_tags succeeds");
+esGet("/_flush");
+esGet("/_refresh");
+countTest(0, "date=-1&expression=" . uri_escape("tags==MCPCAPTEST"));
+
+# a falsy but present ids must not skip the cap: the api picks its by-query
+# branch with `if (req.body.ids)`, so "" has to be treated as no ids here too
+$json = callTool("arkime_add_tags", "{\"date\":-1,\"expression\":\"file=$pwd/socks-http-example.pcap\",\"tags\":\"MCPCAPEMPTY\",\"ids\":\"\"}");
+is($json->{result}->{isError}, JSON::false, "arkime_add_tags with an empty ids succeeds");
+esGet("/_flush");
+esGet("/_refresh");
+countTest(2, "date=-1&expression=" . uri_escape("tags==MCPCAPEMPTY"));
+
+$json = callTool("arkime_remove_tags", '{"date":-1,"expression":"tags==MCPCAPEMPTY","tags":"MCPCAPEMPTY"}');
+is($json->{result}->{isError}, JSON::false, "arkime_remove_tags cleans up the empty ids tag");
+esGet("/_flush");
+esGet("/_refresh");
+countTest(0, "date=-1&expression=" . uri_escape("tags==MCPCAPEMPTY"));
 
 ################################################################################
 # mcpMaxQueryDays - the test2 viewer on 8124 sets nothing, so this exercises the
--- a/tests/config.test.ini
+++ b/tests/config.test.ini
@@ -76,6 +76,8 @@ packetPortalListen=true
 mcpEnabled=true
 mcpAuthMode=regressionTests
 mcpMaxQueryDays=-1
+# deliberately tiny so api-mcp.t can prove the tag cap is applied
+mcpMaxTagSessions=2
 plugins=test.so;tagger.so;wise.so;scrubspi.so;zeekintel.so
 cronQueries=true
 dontSaveBPFs=port 12345
--- a/viewer/apiMcp.js
+++ b/viewer/apiMcp.js
@@ -103,6 +103,61 @@ class MCPViewerAPIs {
   }
 
   // --------------------------------------------------------------------------
+  /**
+   * How many sessions one tag change may touch, mcpMaxTagSessions, default
+   * 10000, -1 for no limit. parseInt is not enough on its own: it stops at the
+   * first non digit, so '10k' would silently become a cap of 10.
+   */
+  static #maxTagSessions () {
+    const raw = String(ArkimeConfig.get('mcpMaxTagSessions', 10000)).trim();
+    if (!/^-?[0-9]+$/.test(raw)) {
+      console.log('ERROR - mcpMaxTagSessions is not a number, using 10000');
+      return 10000;
+    }
+    return parseInt(raw, 10);
+  }
+
+  // --------------------------------------------------------------------------
+  /**
+   * Build the query for a tag change and cap how many sessions it may touch.
+   * Tagging by expression otherwise inherits the api's own 1000000 default.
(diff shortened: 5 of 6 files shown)


31453:
## Review notes

Reviewed at extra-high effort and then validated each candidate by driving the real code — `MCPServer.router` behind express with a real `ArkimeConfig`, and the real `MCPViewerAPIs` tool handlers with stubbed api handlers (no ES). Findings below are the ones that survived; one candidate was refuted and is listed at the end.

**The premise checks out, and so does the fix.** Against `main`, `mcpAllowedIps=10.0.0.0/abc;10.0.0.0/33` really does die on the first unauthenticated `POST /mcp`:

```
Assertion failed: (prefix_len <= 128), function add_route_ipv6, file btrie.cc, line 317.
exit code: 134
```

Against this branch the same value gives `403 Forbidden` with both entries logged and no abort, so the new `[test4]` assertion is correct. `10.0.0.0/` and `::1/` also flip from allow-all to deny, and `127.0.0.0/8` still passes through to auth.

---

### 1. `"ids":""` bypasses `mcpMaxSessions` completely

`#tagQuery` chooses the by-query branch with `query.ids === undefined`, but `addTags`/`removeTags` choose it with `if (req.body.ids)` (apiSessions.js:2712, 2778) — a truthiness test. Any falsy-but-defined `ids` lands in the gap: `#tagQuery` skips setting `query.length`, and the api still takes the by-query branch, where `#sessionsListFromQueryChunky` defaults `req.query.length` to `1000000`.

Driving the real `arkime_add_tags` handler with `mcpMaxSessions=2`:

| `ids` argument | `query.length` set | branch taken | max sessions touched |
|---|---|---|---|
| omitted | `"2"` | by-query | **2** |
| `""` | – | by-query | **1000000** |
| `0` | – | by-query | **1000000** |
| `null` | – | by-query | **1000000** |
| `false` | – | by-query | **1000000** |
| `"a"` | – | ids | len(ids) |
| `"a,b,c"` | – | – | tool error, correct |

There is no input-schema validation on the way in (`#callTool` passes `params.arguments` straight to the handler), so all four of those are reachable from a tool call. Adding `"ids":""` restores the exact million-session behaviour the PR is closing.

Aligning the two conditions fixes it:

```js
if (!query.ids) {
  query.length = Math.max(1, limit);
} else if (...) {
```

Worth a test case too — the new section only covers `ids` omitted and `ids` over the cap.

### 2. A non-string `mcpAllowedIps` still takes the service down pre-auth

`#addCidr` assumes `cidr` is a string. `mcpAllowedIps=false` in a plain `.ini` is enough: `arkimeConfig.js:246` coerces the string `'false'` to boolean `false`, and `getFullArray` returns `[false]` for a non-string, so `cidr.split` throws a `TypeError`. `#checkIps` is called at mcpServer.js:87 — before the `try`/`catch` at line 133 — inside an async express handler, so the rejection is unhandled:

```
config mcpAllowedIps = [false]
  >> unhandledRejection: TypeError - cidr.split is not a function
  >> NO RESPONSE: AbortError - This operation was aborted
```

The request never gets a response. Viewer survives on the `unhandledRejection` handler at viewer.js:2396; **cont3xt has no such handler, so the process dies** — the same pre-auth, remote-triggerable shape as the `/33` abort, just via a different malformed value. Not newly introduced (`main` does the same), but `#addCidr` exists to reject malformed entries and this one slips past. `mcpAllowedIps=true` too.

```js
if (typeof cidr !== 'string') { return false; }
```

### 3. The same allow-all bug is still live in `userAuthIps`

The validated parse is a private `MCPServer` helper, so the identical unvalidated parse at auth.js:206-213 — fed from `getArray('userAuthIps')` at auth.js:100 — is untouched. Running that exact code against `arkime-iptrie`:

| `userAuthIps` | matches 1.2.3.4 | matches 8.8.8.8 | matches 2001:db8::1 |
|---|---|---|---|
| `10.0.0.0/` | yes | yes | no |
| `10.0.0.0/abc` | yes | yes | yes |
| `::1/` | yes | yes | yes |

None of these abort, so unlike the `/33` case the admin gets no startup error and no restriction — on the site-wide header-auth gate that this PR's own comment cites as the model for the MCP one. `userAuthIps=false` is also an uncaught startup `TypeError`, because `const parts = cidr.split('/')` at auth.js:207 sits outside the `try`.

Both call sites want one shared validated helper (`ArkimeUtil`, or exported from `auth.js`) rather than an MCP-private copy.

### 4. Capped tag changes silently report full success

The by-query branch truncates, but apiSessions.js:2743 returns an unconditional `{ success: true, text: 'Tags added successfully' }` — `total` is in scope and never reported. With `mcpMaxSessions=10000` and an expression matching 50000 sessions, the tool answers with plain success and the model reports the whole expression as tagged.

It is worse for `arkime_remove_tags`: asked to clear a tag that the UI applied to 50000 sessions, the assistant gets "Tags removed successfully" and 40000 still carry it, with nothing in the response indicating a partial apply. The ids path errors up front for exactly this reason; the by-query path should return the touched count, or say it truncated.

### 5. A `mcpMaxSessions` typo silently becomes a tiny cap

`parseInt` stops at the first non-digit, so `isNaN` only catches values with no leading digit at all:

| `mcpMaxSessions` | effective cap | logged? |
|---|---|---|
| `10000` | 10000 | – |
| `1O000` (capital O) | **1** | no |
| `10k` | **10** | no |
| `10 000` | **10** | no |
| `abc` | 10000 (default) | yes |

Combined with #4, every tag change is then silently truncated to 1 or 10 sessions and still reports success. A digits-only check — the same one `#addCidr` applies to the prefix — would catch all three.

### 6. `mcpMaxQueryDays` fails open on the same class of typo

Its new sibling logs and falls back to the shipped default; `#checkQueryDays` at apiMcp.js:87 returns silently on `NaN`, disabling the cap:

```
mcpMaxQueryDays="7"      date=-1   -> blocked
mcpMaxQueryDays="sevem"  date=-1   -> ALLOWED (no day cap)
mcpMaxQueryDays="sevem"  date=8760 -> ALLOWED (no day cap)
```

Two adjacent MCP caps with opposite behaviour on the same admin mistake. Since the PR is adding the twin, worth giving this one the same log-and-use-the-default treatment.

### 7. `mcpMaxSessions` does not bound `arkime_sessions`

The name reads as a global session cap but only the two tag tools consult it. `arkime_sessions` passes `length` straight through (apiMcp.js:169), capped only by `maxSessionsQueried` (2000000):

```
mcpMaxSessions=10000, arkime_sessions {"length":1000000} -> query.length=1000000
```

One read call can still pull a million session records into the model's context. Either apply the cap there too, or name it `mcpMaxTagSessions` to match "how many sessions one tag change may touch".

### 8. `mcpAllowedIps` is still only validated on the first request

`userAuthIps` `process.exit(1)`s at startup for a bad entry (auth.js:216). `mcpAllowedIps` is parsed lazily inside `#checkIps` and memoized, so a typo produces a clean silent startup and then one ERROR line at the moment the gateway's first request 403s — never repeated. Now that an unusable list fails closed, that is the difference between "the admin fixes it before anyone notices" and "/mcp is dead and the log line has scrolled away". Parsing at config load would surface it while they are still watching.

### 9. No test covers a valid explicit `mcpAllowedIps`

`mcpAllowedIps` appears exactly once across every test config, and it is the deliberately-malformed `[test4]` value. `[test3]` covers the unset loopback default. So a regression in `#addCidr` that rejected well-formed CIDRs — a tightened regex, an off-by-one on the range check, the v4 `96 + prefix` offset — would leave the suite green while breaking every proxied deployment, which is the reason the setting exists. I confirmed the accept path works today (`127.0.0.0/8` → 401), it just is not pinned. One node with `mcpAllowedIps=127.0.0.0/8` asserting the request reaches auth would cover it.

### Minor

- **Abbreviated IPv4 is accepted silently.** `mcpAllowedIps=10/8` returns `true` with no ERROR logged, because `::ffff:10` is a valid IPv6 literal (`::ffff:0.0.0.16`) — the entry becomes `0.0.0.0/8` and matches nothing intended, so the gateway is denied by a list the log says was accepted. A `/^\d+\.\d+\.\d+\.\d+$/` check on `parts[0]` in the v4 branch would reject it.
- **The loopback default discards `#addCidr`'s return value** (mcpServer.js:376-377). Both literals work today, so this is future-proofing only: if either ever failed, the trie would silently end up empty and deny everything with no log, since the `added === 0` diagnostic only covers the configured-list branch.
- **config.ini.sample doesn't mention the new failure behaviour** — that malformed entries are skipped with an ERROR and that a list with no usable entry denies everything. That is the part an admin needs when /mcp starts 403ing.

### Refuted

I flagged `list[0] !== ''` (mcpServer.js:360) as dead code on the grounds that `getFullArray` filters empty entries — **that was wrong**. arkimeConfig.js:269 returns arrays as-is without trimming or filtering, so a JSON/YAML `mcpAllowedIps: ["", "10.0.0.0/8"]` reaches `#checkIps` with `list[0] === ''`. The guard is live. Leave it.

31453:
## Review addressed — `30d6940`

Validated every point by driving the real code (express + real `ArkimeConfig` for the route, real `MCPViewerAPIs` handlers with stubbed apis). **All 9 findings and all 3 minors confirmed, all fixed.** The refuted `list[0] !== ''` point was correct to refute — guard left alone.

### 1. `"ids":""` bypassed the cap — confirmed, fixed

Reproduced exactly as described:

| `ids` | `length` set (before → after) |
|---|---|
| omitted | `2` → `2` |
| `""` | – → `2` |
| `0` | – → `2` |
| `null` | – → `2` |
| `false` | – → `2` |
| `"a"` | – → – (ids branch) |
| `"a,b,c"` | tool error → tool error |

Aligned the condition to `if (!query.ids)`. Regression test added.

### 2. Non-string `mcpAllowedIps` — confirmed, fixed

`mcpAllowedIps=false` → `[false]` → `cidr.split is not a function` pre-auth, request got **no response at all**. Confirmed cont3xt dies: viewer survives only via `viewer.js:2396`, and running the same harness without that handler is fatal. `true` was the same. Now `403` with the entry logged, in both service shapes.

### 3. `userAuthIps` — confirmed, fixed by sharing the parse

Verified the identical allow-all against `arkime-iptrie` (`10.0.0.0/` and `/abc` matched `8.8.8.8`; `/abc` and `::1/` matched IPv6 too), and that `cidr.split` sits outside its `try`. Extracted `ArkimeUtil.addCidrToTrie` and used it in both places.

⚠️ **Behaviour change worth a second opinion:** a `userAuthIps` entry that was previously accepted-and-ignored, e.g. `10.0.0.0/`, is now a startup `process.exit(1)`. That matches what the loop already did for a bad address, and failing loudly beats silently trusting every source IP — but it will stop a currently-"working" deployment that has such a typo. Say the word if you'd rather it log and skip.

### 4. Silent full success on a capped change — confirmed, fixed

`total` was in scope and discarded. `addtags`/`removetags` now return `count` (documented in the JSDoc, additive), and the tools replace the text when the count hits the cap:

> Only the first 2 matching sessions were tagged, because mcpMaxTagSessions is 2. Narrow the expression, or pass explicit ids, to tag the rest.

plus `truncated: true`. `arkime_remove_tags` gets the same with untag wording.

### 5 & 6. Integer validation — confirmed, fixed

| `mcpMaxTagSessions` | before | after |
|---|---|---|
| `1O000` | cap **1**, silent | 10000 + ERROR |
| `10k` | cap **10**, silent | 10000 + ERROR |
| `10 000` | cap **10**, silent | 10000 + ERROR |

`mcpMaxQueryDays="sevem"` went from **allowing `date=-1` and `date=8760`** to blocking both with an ERROR logged.

### 7. Naming — fixed by renaming

Took the rename option: **`mcpMaxSessions` → `mcpMaxTagSessions`**. It only ever bounded tag changes, and the setting is new in this PR so nothing is broken by renaming. I deliberately did *not* extend it to `arkime_sessions` — bounding read `length` is a real improvement but changes read-tool semantics nobody asked for; happy to do it as a follow-up if you'd prefer the broader cap.

### 8. Parse-at-first-request — acknowledged, not changed

Confirmed the difference is real (`userAuthIps` exits at startup; `mcpAllowedIps` logs once, lazily). Left as-is: the fix means threading a config-load hook through both service entrypoints, which felt like more churn than this PR should carry. Flagging rather than silently dropping — happy to add it.

### 9 & minors — fixed

- Test config reworked so the accept path is pinned: `[test5]` now uses `mcpAllowedIps=127.0.0.0/8;10.0.0.0/abc;10.0.0.0/33` — one good entry plus both malformed shapes. A tool call succeeding there proves the good entry is honoured, the bad ones are skipped, and `/33` did not abort. `[test4]` covers the non-string case. Also added `waitFor` for 8127/8128, which were started but never waited for.
- `10/8` now rejected (dotted-quad check on the v4 branch).
- Loopback default checks the return value and logs.
- `config.ini.sample` documents that malformed entries are skipped with an ERROR and an unusable list denies everything.

### Also found while fixing

`ArkimeUtil.sanitizeStr` threw on a truthy non-string — which is how my own new `mcpAllowedIps` error path crashed while trying to log the bad value (`false` survived only because it's falsy). Hardened it to coerce, since its whole job is making untrusted values safe to log.

Plan 105 → 117. `npm run lint` clean, `perl -c` clean. Still not run against a live cluster.

31453:
## Resliced — this PR is now just the tag-scope fix

Force-pushed to `52b73d8`. The config hardening moved to #4233.

The premise check that prompted this: of the three original fixes, **only one was not config-triggered**. The `mcpAllowedIps` allow-all and the `/33` abort both needed an operator typo to exist. This one didn't — with the shipped config and no mistake, `arkime_add_tags` with no `ids` tags up to 1,000,000 sessions, because the tool never set `length`. The arguments come from a model reading untrusted captured traffic, not from a config file. That is the only genuinely attacker/agent-reachable finding of the set, so it gets to stand on its own.

Two things I got wrong and am correcting:

- **I rated the `mcpAllowedIps` fail-open S1/HIGH on an "auth bypass" framing that presupposed a misconfiguration.** Prerequisite access is the first question in severity calibration, and "operator makes a typo" is not attacker-reachable. It should have been S2 at most, filed as robustness with a security consequence. Corrected in #4233, which frames it as a fail-open class rather than a vulnerability.
- **The PR shipped three different answers to "bad config"** — `userAuthIps` exited, `mcpAllowedIps` skipped-and-continued, the numeric caps logged-and-defaulted. I arrived at each locally and never compared them. #4233 replaces all three with one validated path that refuses to start.

Also dropped from here and folded into #4233: the `mcpMaxQueryDays` validation, and the `mcpAllowedIps` test fixtures. `mcpMaxTagSessions` still reads its value with a small local integer check so this PR stays correct standalone; it converts to `ArkimeConfig.getInt` once #4233 lands.

Net effect here: 10 files / +248−28 → **6 files / +129−11**, one commit. Test plan 117 → 111.

31453:
The UI always bounds tag scope — `ids`, `numVisible`, or `numMatching`, the count the user was shown. This tool omitted `length` entirely, so it was the only caller reaching the 1,000,000 fallback. This is parity. The config knob is the weakest part; count-first, like `arkime_create_hunt`, would also work.

PR #4233, opened 2026-09-02
Refuse to start on invalid access control or resource settings

31453:
Split out of #4230, which was patching individual settings. A setting whose job is to restrict access is worse than useless when silently ignored — the operator believes a control is in force and it is doing nothing.

- Validate access controls and resource bounds when the config loads, report every bad one at once, then exit. Owners register their own settings via `ArkimeConfig.registerValidated`. Matches the existing `ArkimeConfig.exit` convention.
- Add `ArkimeConfig.getInt`/`getNumber`. `parseInt` stops at the first non-digit, so `uploadFileSizeLimit=2g` meant **2 bytes** and `maxSessionsQueried=2m` meant **NaN** into an OpenSearch `size`.
- One `ArkimeUtil.parseCidr`/`addCidrToTrie`/`buildIpTrie` for every CIDR parse. A missing or non-numeric prefix matched everything, `/33` aborted the process via an uncatchable C++ assert, an ini `false` threw and killed cont3xt, and `/0` silently became `/32`. The wise and cont3xt sites are reachable from **query results and downloaded files**, not just operator config.
- `authJwtClockSkew=60s` made `clockTolerance` `NaN`, and jose compares `exp` against `NaN`, which is always false — **expired bearer tokens verified indefinitely**. Validated, plus a guard where jose is called.
- `ArkimeUtil.sanitizeStr` no longer throws on a truthy non-string.

Mostly a fail-open class rather than an attacker-reachable one — the config is the operator's — which is why it belongs at the config layer rather than each call site. The JWT expiry bug is the exception.

**Tradeoff:** refuse-to-start turns a latent typo into a failed boot on upgrade. Deliberate, but say so if you'd rather resource bounds fell back to the default.

`tests/config-validate.t`, 33 assertions, no cluster needed — includes allow-list matching, so a regression in the rewritten trie construction fails there rather than leaving the suite green.


31453:
## Code review

Reviewed at extra-high effort. Two findings undermine the core claim of the PR, and I verified both by running the patched `common/arkimeConfig.js` directly rather than by inspection.

---

### 1. `[nodeClass]` section settings escape validation entirely

`common/arkimeConfig.js:121`

`#validateSettings()` runs at line 121, **before** the `loadedCbs` loop at line 127. But `viewer/config.js:313` splices the `[nodeClass]` section into `#defaultSections` from *inside* an `ArkimeConfig.loaded(..., true)` callback. So validation reads a shorter section list than every runtime consumer does.

Verified with the same section ordering viewer uses:

```
$ cat c.ini
[default]
nodeClass=myclass
[myclass]
mcpAllowedIps=10.0.0.0/33
[testnode]
nodeClass=myclass

$ node run.js
CONFIG OK - validation passed. sections now: [ 'testnode', 'myclass', 'default' ]
runtime read of mcpAllowedIps: [ '10.0.0.0/33' ]
bad entries seen at runtime (silently dropped by mcpServer): [ '10.0.0.0/33' ]
trie find 10.1.2.3 -> undefined
```

Validation passes, then `mcpServer` silently drops the entry, leaving an empty trie that refuses **every** `/mcp` request with no log line at all. Same gap for the numerics: `maxSessionsQueried=2m` in a `[nodeClass]` section still silently means 2000000, which is the exact bug this PR exists to fix.

Either validate across the full section list (move the nodeClass resolution ahead of validation), or validate after the `loaded` callbacks have run.

### 2. `authJwtClockSkew` is not in `#VALIDATED`, and a non-numeric value disables JWT expiry

`common/arkimeConfig.js:326`

`common/auth.js:605` does `clockTolerance: +Auth.#authConfig.jwtClockSkew`. An operator writing `authJwtClockSkew=60s` (or `1m` — the setting name carries no unit hint) gets `NaN`. jose 4.15.9 `lib/jwt_claims_set.js:63` takes `typeof NaN === 'number'` at face value, then line 89 evaluates:

```js
if (payload.exp <= now - tolerance) { throw new JWTExpired(...) }
```

`exp <= NaN` is always false, so `JWTExpired` never fires. The `nbf` check fails open the same way. **Expired bearer tokens are accepted indefinitely on the `/mcp` JWT path.**

This is the same silent-coercion class the PR is fixing, on the most access-critical numeric in a file the PR already touches, while `uploadFileSizeLimit` and `maxSessionsQueried` — perf knobs by comparison — did make the table.

---

### 3. `MCPServer.#checkIps` discards `buildIpTrie`'s `bad` list

`common/mcpServer.js:332`

`buildIpTrie`'s own doc comment says *"Callers must treat a non empty bad list as a hard failure: a partially built allow list is worse than none."* This call site writes `ArkimeUtil.buildIpTrie(list).trie` and throws `bad` away, on the strength of the comment *"a bad entry never gets this far"* — which is false for a nodeClass-section list (finding 1, verified) and after a runtime `ArkimeConfig.reload()`.

`common/auth.js:209` does check `bad` and exit. This one doesn't, so the result is a silently partial or silently empty allow list where the old code failed loudly.

### 4. Int settings from `.json`/`.yaml` are truncated, not rejected

`common/arkimeConfig.js:296`

`getInt` returns `Math.trunc(value)` for anything already a `number`, so the validated path is only actually validated for one of the three supported config formats:

```
$ echo '{ "default": { "maxSessionsQueried": 2.9 } }' > c4.json
$ node -e "...initialize({defaultConfigFile:'./c4.json',...})"
CONFIG OK, getInt -> 2
```

A session cap of **2**, started clean. The `.ini` equivalent refuses to start with *"is not an integer"*. Multi-format config is first-class here, so this is the same silently-tiny-cap failure in a different file extension.

Related: when truncation does trip the minimum, the message reports the truncated value rather than what the operator wrote — `uploadFileSizeLimit: 0.5` is reported as `uploadFileSizeLimit is 0, which is below the minimum of 1`.

### 5. `min: 1` turns a working `uploadFileSizeLimit=0` into a boot failure

`common/arkimeConfig.js:329`

`viewer/internals.js` passes this to multer's `limits.fileSize`, where `0` makes busboy reject every file — a functioning way to disable uploads. `tests/config-validate.t:81` now asserts that config refuses to start, so a working deployment fails to boot on upgrade. That contradicts the PR's framing that only *broken* settings fail.

The bound also doesn't accomplish much: `uploadFileSizeLimit=1` (one byte, equally "no uploads") passes. `maxSessionsQueried=0` has the same shape. Either allow `0` for both, or give them the `unlimited` sentinel treatment `mcpMaxQueryDays` gets.

### 6. Empty `userAuthIps` blocks everything, and validation deliberately waves it through

`common/arkimeConfig.js:338`

The comment reads *"An all empty list is how 'not set' looks, leave that to the caller"* — but the caller doesn't handle it. `common/auth.js:206` tests `if (options.userAuthIps)`, and an empty array is truthy. Verified:

```
$ printf '[default]\nuserAuthIps=\n' > c5.ini
CONFIG OK; getArray -> [] truthy? true
bad [] find ::ffff:10.1.1.1 -> undefined find ::1 -> undefined
```

Neither the header-mode `127.0.0.0/8` + `::1` default nor the non-header `::/0` default is installed, so every request including loopback is refused, silently. Validation is the one place positioned to catch this and it explicitly opts out.

### 7. `getInt`/`getNumber` substitute the default with no warning

`common/arkimeConfig.js:298`

The doc calls hand coercion "a trap" because `'10k'` silently becomes 10. The replacement is equally silent in the other direction: `Config.getInt('maxSessionsQueried', 2000000)` on `'2m'` returns 2000000, no log line. For the ~54 reads left for the later sweep, converting them to `getInt` buys no diagnostic at all.

Worse, called without a default — as `#validateSetting` itself does at line 351 — it returns `undefined`, so arithmetic on `getInt('foo')` yields `NaN`. That's the `maxSessionsQueried=2m` → NaN bug, relocated. A `console.log('WARNING - <key> is not an integer, using <default>')` in the fallback closes it.

### 8. The startup error names the config file even when the value came from elsewhere

`common/arkimeConfig.js:379`

`reload()` merges every `ARKIME_*`/`ARKIME__*` env var into `#config`, and `processArgs` feeds `-o key=value` into `#override`; `getFull` reads both. So an operator running with `ARKIME__maxSessionsQueried=2m` sees `ERROR - /opt/arkime/etc/config.ini has 1 bad setting` and greps a file where the key isn't present.

Same path makes `ARKIME__uploadFileSizeLimit=""` — a common container idiom for "leave it unset" — resolve to `''` and refuse to start. Empty should probably be treated as unset for numerics the way it already is for cidrs.

---

### Scope: 8 of 10 CIDR→iptrie parses left on the old path

`common/arkimeUtil.js:201`

`addCidrToTrie` is the right shared fix, but only 2 of 10 call sites were converted, and the ones left behind are the ones fed by **remote data**:

| site | shape | consequence |
|---|---|---|
| `wiseService/source.splunk.js:108` | `+parts[1] \|\| 32`, no try/catch | a `/33` in a query result hits `assert(prefix_len <= 32)` at `btrie.cc:300` → `abort()` |
| `wiseService/source.databricks.js:108` | same | same |
| `cont3xt/integrations/csvjson/index.js:232` | `parts[1] !== undefined ? +parts[1] : 32` | `/abc` → `NaN` → `Uint32Value` coerces to 0 → **prefix 0, matches every IP**, from a downloaded file |
| `wiseService/source.reversedns.js:54` | no try/catch at all | throw escapes |
| `wiseService/wiseSource.js:88,104`, `wiseService/wiseService.js:856` | `+parts[1] \|\| 32` | a legitimate `/0` becomes `/32` |

The config-layer abort is genuinely fixed; the identical abort reachable from upstream query data is not.

### Altitude: `#VALIDATED` hard-codes app-specific setting names in `common/`

`common/arkimeConfig.js:326`

`uploadFileSizeLimit` and `maxSessionsQueried` are viewer settings; `mcpAllowedIps` and `mcpMaxQueryDays` are MCP settings. Because the table is private to the shared config class, parliament/wise/cont3xt all run checks against sections where those settings are meaningless, and an app can't register its own. The PR body names the recurring cost itself: `mcpMaxTagSessions` from #4230 "converts to `getInt` once that lands" — another edit to this central table from a different component's PR.

An `ArkimeConfig.registerValidated(spec)` called from each app's `initialize` (before the internal pass) keeps the mechanism shared and the setting list local to the owner.

---

### Tests

- **No test that a valid allow list still matches the right peers.** `userAuthIps` and `mcpAllowedIps` appear in `tests/` only in the new file, which asserts exit codes and stderr text exclusively. Both controls had their trie construction rewritten — `parseCidr`'s v4 → `::ffff:` + `(96 + prefix)` mapping, and the loopback default now via `buildIpTrie(['127.0.0.0/8', '::1'])` instead of two literal `add()` calls. I checked by hand that both are equivalent to the old code, but nothing keeps them that way, and a regression would widen or empty the allow list with the whole suite green.
- **The cluster-free test only runs in the cluster suite.** `@tests = glob("*.t")` lives only inside `doViewer` (`tests/tests.pl:501`), after `waitFor` on 9200/7379/4566 and the viewer startup. The capture path globs `pcap/*.pcap` and never runs `.t` files. So the stated benefit — "needs no cluster and no generated `common/version.js`" — isn't realised: the capture job skips it, the viewer job pays full bring-up first, and it's invisible to anyone running `./tests.pl` without a local cluster.
- **Temp file leak and env bleed.** `unlink $runner` runs only at the end, so a `die` in `tryConfig` leaves `/tmp/arkime-config-validate-<pid>.js` behind (an `END` block or `File::Temp` covers both files). And since `reload()` copies every `ARKIME_*` var into the parsed config, a developer or runner with `ARKIME__maxSessionsQueried` exported gets spurious failures in the "empty config starts" and "valid config starts" cases with no hint why — the child `node` should run with those stripped.

### Minor

- `release/config.ini.sample:194` enumerates the validated settings in prose, duplicating `#VALIDATED` with nothing tying them together; the PR body already schedules the next divergence. The block also lands mid-way through the header-auth settings rather than under a heading of its own.
- The diff leans heavily on multi-line explanatory comments (`arkimeConfig.js:117-120`, `288-294`, `315-324`; `arkimeUtil.js:81-83`, `168-177`) that argue for the design decision rather than describe the code. That argument is already in the PR description, which is a better home for it.

---

### On the stated tradeoff

The refuse-to-start policy reads as the right call to me, and the "all bad settings in one pass" output is the right shape. The problem isn't the policy, it's that findings 1, 3, 4, 6 and 8 each leave a route by which a bad setting still reaches runtime silently — so the fleet pays the boot-failure cost without getting the guarantee that justifies it. Findings 1 and 3 in particular should land before this merges.

I did not run the test suite; the verification above was done by driving `common/arkimeConfig.js` directly with scratch config files.

31453:
## Review addressed — `94c0ea0`

Validated all 11 findings by driving the code. **None refuted.** Everything is fixed here rather than deferred.

### The two that undermined the PR

**1 — `[nodeClass]` sections escaped validation.** Confirmed exactly as described. `#validateSettings()` now runs *after* the `loaded` callbacks, since `viewer/config.js:314` splices the section in from one:

```
before: CONFIG OK - sections now: ["testnode","myclass","default"]
        runtime mcpAllowedIps: ["10.0.0.0/33"] -> dropped silently, /mcp denied everything
after:  ERROR - 1 bad setting ... refusing to start
```

**3 — `#checkIps` discarded `bad`.** It now logs each unusable entry and installs an empty trie, denying everything loudly, which is what `buildIpTrie`'s doc comment asks of callers.

### The one that isn't about config at all

**2 — `authJwtClockSkew=60s` accepted expired JWTs.** Confirmed against the vendored jose 4.15.9:

| `clockTolerance` | expired token |
|---|---|
| `0` | rejected `ERR_JWT_EXPIRED` |
| `60` | rejected `ERR_JWT_EXPIRED` |
| `NaN` | **ACCEPTED** |

Added to the validated set, plus a guard where jose is called in case a caller supplies `authConfig` directly. This was the only attacker-relevant item in the set — a stolen expired token worked forever — and it's the strongest argument for the PR's premise, so it belongs here rather than in a separate change.

### Scope — the CIDR sites, with a trap

The remaining parses are converted, but **not the way suggested**: those tries are queried with raw values (`excludeIPs.find(value)`), while `parseCidr` maps v4 to `::ffff:`. Verified that mixing them breaks matching outright —

```
raw-added,    find('10.1.2.3') -> true
mapped-added, find('10.1.2.3') -> undefined   <- would have silently broken every wise ip source
```

So `parseCidr` takes a `mapV4` option and the wise/cont3xt sites keep their keying. Also split the error policy by data source: config-fed lists (`excludeIPs`, `onlyIPs`, wise excludes) still exit at startup; the remote-data-fed ones (splunk, databricks, csvjson, simpleSource) log and skip, because upstream data shouldn't kill the process. `/0` is now honoured instead of becoming `/32` — a behaviour change, and the right one.

### The rest

**4** json/yaml `2.9` now refuses instead of truncating to a cap of 2, and errors quote what the operator wrote. **5** `0` allowed for both bounds, and empty is unset — you were right that `min: 1` was arbitrary when `1` passed and is equally "no uploads"; the premise that someone deploys `0` I took on trust, but the conclusion stands either way. **6** empty `userAuthIps` is unset (`?.length`), so the loopback default applies again — pre-existing, not introduced here, but validation was the right place to catch it. **7** `getInt`/`getNumber` warn on fallback. **8** the message no longer names the config file, since the value can come from the environment or `-o`.

**Altitude** — `registerValidated()` added; `mcpServer`, `auth` and `viewer/config.js` register their own. Note the tradeoff this creates: a setting is only validated if its owning module is loaded. That's true in the apps, and the test now requires those modules the way the apps do so it can't drift.

### Tests: 21 → 33, all passing locally

Added the allow-list **matching** test you asked for, so a regression in the rewritten trie construction fails here instead of leaving the suite green. Plus the nodeClass case, clock skew, json/yaml, `0`/empty. Fixed the temp-file leak with an `END` block, and the child now runs with `ARKIME_*` stripped — I reproduced your env-bleed concern (`ARKIME__maxSessionsQueried=2m` breaks unrelated cases), and the suite is now 33/33 even with it exported.

### Not done

The `glob("*.t")` placement — you're right that it only runs inside `doViewer` after full bring-up, so "no cluster needed" isn't realised. Restructuring the runner felt out of scope for this PR; the description no longer claims the benefit.

PR #4244, opened 2026-09-08
bulk update changelog because elyse sucks at it :)

31453:
Backfills the v7.0.0 CHANGELOG with the UI work that landed on dev7 but was never written up.

🤖 Generated with [Claude Code](https://claude.com/claude-code) and reduced greatly by @31453


awick (approved):
LGTM

PR #4245, opened 2026-09-08
Add a roles page to parliament and tidy the shared users/roles pages

awick:
- Add a Roles page wrapping the shared common/vueapp/Roles.vue, with a /roles route
- Gate Users on usersAdmin and Roles on assignableRoles, matching what the apis enforce, instead of parliamentAdmin
- Guard the users and roles routes with the same checks so direct urls redirect
- Show the password change form in parliament, which registers /api/user/password but was excluded from the form
- Keep the role permission toggles to viewer and cont3xt, since they set arkime capabilities
- Drop the cont3xtDarkTheme and dark props, dead since the Vuetify migration removed the bindings they fed
- Give Roles.vue the same loading slot Users.vue has, so viewer's ArkimeLoading override works instead of being ignored
- Give Roles.vue arkime-container-fluid; its container-fluid class was a bootstrap leftover with no styles behind it
- Wrap viewer's Users page in arkime-container-fluid so it isn't flush against the window
- Drop the unused getTheme mapping from parliament's Users page

## License


31453:
dunno why claude had such a hard time with the commits 🦀  but the diff looks like it's supposed to 😄

31453 (approved):

PR #4256, opened 2026-09-11
Make every page's nav bands the same height with centered content

awick:
- .page-subnav owns its flex/height/padding: 44px, content centered
- .page-subnav--primary is a page's first chrome row: 48px, matching <arkime-search>
- strip Vuetify's negative block margin and col padding from v-row bands, which painted them 4px above the chrome and 4px past its bottom
- center the search row and drop the expression input's stray bottom margin
- dock the viz options button to the sub-navbar instead of nudging it with top:108%
- center v-pagination's buttons, which sat on a text baseline
- center the page tab strip and drop its translateY nudges; give it the chrome drop shadow it was hiding from .page-toolbar
- .sub-navbar (settings/upload/viewconfig) matches the 48px band, centered
- render the files pager unconditionally so it stops popping in after the fetch

## License

### DIFF
--- a/common/common.css
+++ b/common/common.css
@@ -331,7 +331,13 @@ pre {
   left: 0;
   right: 0;
   z-index: 100;
-  padding: var(--px-lg) var(--px-md) var(--px-sm) 13px;
+  /* same 48px band as the viewer's search bar / first .page-subnav, with the
+     title vertically centered -- symmetric padding, height from min-height,
+     not from the content plus lopsided padding */
+  display: flex;
+  align-items: center;
+  min-height: 48px;
+  padding: 0 var(--px-md) 0 13px;
   background-color: rgb(var(--v-theme-secondary-lightest));
   /* drop-shadow only below -- a full all-sides shadow leaks up under
      the primary navbar. */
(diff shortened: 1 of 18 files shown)


31453 (approved):
Reviewed the subnav-height normalization. Ran a multi-angle pass (correctness, removed-behavior, cross-file, reuse/simplification/efficiency/altitude, CLAUDE.md conventions) and found 5 confirmed regressions/bugs, all fixed and pushed in 0ff4e9dcb:

- `.page-subnav`'s new `padding-block: 6px` was pushing the band on Arkime.vue/Spiview.vue (44px `size="large"` buttons) from 44px to 56px — added a `page-subnav--tall` modifier to exempt those two rows
- Files.vue's removed `v-if="files"` was flashing a "Showing all 0" pager before data loads — restored the guard
- The tab-strip's `.v-btn__content { transform: translateY(-1px) }` was removed with nothing replacing it, leaving icon/label baseline misalignment — restored
- ViewConfigPage.vue's `marginTop` calc still hardcoded `50` after the band height changed to `48px` — updated to match

Verified all four visually against a local dev server (band heights measured via computed styles) plus a full `npm run lint` pass. Left two lower-confidence/plausible findings (a settings-page clearance edge case, minor Spigraph spacing) and two cleanup suggestions (magic-number duplication, a bit of redundant Pagination.vue CSS) unaddressed since they're not confirmed regressions.

PR #4262, opened 2026-09-14
Fix session queries not enforcing a role-inherited time limit

31453:
buildQuery.js read the raw per-user timeLimit field, which is undefined when the limit comes only from a role, so those users' queries were never time-restricted. Adds getTimeLimit() (mirrors getExpression()) and uses it for enforcement. Includes a regression test.


CHANGELOG
```diff
@@ -46,6 +46,7 @@ NOTICE: Create a parliament config file before upgrading (see https://arkime.com
   - #4260 Add esproxy to docker.sh (thanks @rht-jbittner)
 ## All
   - #4215 Fix email searches, periodic queries and cont3xt history deletes ignoring permissions that come from a role instead of the user (thanks @Juwon1405)
+  - Fix session queries not enforcing a timeLimit that comes only from a role instead of the user
```
awick: 4262

tests/api-unique.t
```diff
@@ -1,4 +1,4 @@
-use Test::More tests => 44;
+use Test::More tests => 45;
```
awick: 46

PR #4263, opened 2026-09-14
Fix cyberchef packet endpoint authorization bypass

31453:
## Summary
- `/cyberchef/:nodeName/session/:id` skipped the hidePcap/disablePcapDownload checks its sibling packet endpoints enforce, letting restricted users read raw packet payloads.
- Adds `hidePcap` and `disablePcapDownload` to the route's permission check.

## Test plan
- [ ] Confirm restricted user (hidePcap=true) gets 403 from `/cyberchef/:node/session/:id`
- [ ] Confirm unrestricted user still gets 200 with packet data


awick:
LGTM. We probably should go thru all the pcap end points for Arkime 7 and decide once and for all the difference because I think its confusing and much of the current code is "wrong".

* disablePcapDownload - really was supposed to be just about hiding the button and the endpoint for getting pcap easily. You could still copy paste the pcap if hidePcap was false and form the file.
* hidePcap - not supposed to be able to see it. (maybe this should imply disablePcapDownload?)

So I think really this should just be hidePcap, but we can refix in 7 depending what we decide.

PR #4265, opened 2026-09-16
Cont3xt MCP search answers with overviews and bounds result size

awick:
- cont3xt_search defaults to the overview per indicator, detail=full for raw data
- Add mcpMaxResultBytes, oversized entries are marked omitted rather than cut
- Add mcpDefaultView so an operator picks the integrations a bare search hits
- Support tags on cont3xt_search and cont3xt_ui_link

## License


31453:
Two issues in the new size-budgeting code:

**1. `mcpMaxResultBytes` doesn't hold once `uiUrl` is attached** — `#fit` (apiMcp.js:352) sizes/trims the result and returns it, but `mcpServer.js`'s `#callTool` appends `data.uiUrl` *after* the handler returns, uncounted. Any `cont3xt_search` that resolves a view can end up over the cap. The truncation test dodges this by using `doIntegrations` with no view — the one case `ui()` skips generating a link.

**2. `cont3xt_integration_search`'s truncated size is measured against the wrong object** — the fallback decision uses `#bytes(body)` (apiMcp.js:546) but reports `#bytes(body.data)` (apiMcp.js:561). Envelope overhead can push `body` over the limit while `body.data` alone reads under it, so `truncated: {bytes: 99930, limit: 100000}` can show up self-contradictorily. Should measure the same object it decided on.

